PR #1, opened 2016-01-11
Fixed package decl + JSON tag issues (via go vet).

elithrar:
Fixed the import path in `cmd/flarectl` to refer to the (correct) `github.com/jamesog/cloudflare-go` import path. This was preventing flarectl from being built automatically.


jamesog:
Thanks!

PR #3, opened 2016-03-10
Add WAF

oholiab:
This adds functions and types for listing both WAF packages and WAF rules


oholiab:
Sorry about the casing, I changed all my "Id" to "ID" but forgot about "Waf".

I assume you meant only removing the var api

jamesog:
No problem, thanks for contributing. This is great!

PR #7, opened 2016-03-17
Implement cache purge functionality

Tenzer:
Let me know if there's anything you would like to get changed.

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -39,16 +39,15 @@ func NewZone() *Zone {
 func (api *API) makeRequest(method, uri string, params interface{}) ([]byte, error) {
 	// Replace nil with a JSON object if needed
 	var reqBody io.Reader
-	switch method {
-	case "GET", "DELETE":
-		reqBody = nil
-	default:
+	if params != nil {
 		json, err := json.Marshal(params)
 		if err != nil {
 			fmt.Println("Error marshalling params to JSON:", err)
 			return nil, err
 		}
 		reqBody = bytes.NewReader(json)
+	} else {
+		reqBody = nil
 	}
 	req, err := http.NewRequest(method, apiURL+uri, reqBody)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


cloudflare.go
```diff
@@ -40,7 +40,7 @@ func (api *API) makeRequest(method, uri string, params interface{}) ([]byte, err
 	// Replace nil with a JSON object if needed
 	var reqBody io.Reader
 	switch method {
-	case "GET", "DELETE":
+	case "GET":
```
jamesog: I think this will break `DeleteDNSRecord`.

The `default` case may need an additional check if `params == nil` and set `reqBody`, else do the rest.

cloudflare.go
```diff
@@ -397,3 +397,18 @@ type WAFRulesResponse struct {
 		TotalCount uint `json:"total_count"`
 	} `json:"result_info"`
 }
+
+type PurgeZoneCacheRequest struct {
+	Everything bool `json:"purge_everything"`
```
jamesog: Please call this `PurgeEverything`

cloudflare.go
```diff
@@ -397,3 +397,18 @@ type WAFRulesResponse struct {
 		TotalCount uint `json:"total_count"`
 	} `json:"result_info"`
 }
+
+type PurgeZoneCacheRequest struct {
```
jamesog: I'm wondering if this should be merged with `PurgeCacheRequest` and set the `omitempty` tag. What do you think?

cloudflare.go
Tenzer: I have removed the `switch` method now and replaced it with a simple `if`. I did (silently) break `DeleteDNSRecord` :/

cloudflare.go
Tenzer: Done.

cloudflare.go
Tenzer: Sounds good to me, I wasn't aware of the `omitempty` tag. I first tried to put both the file and site purge into the same request, but got an error back from the API that only one of `files` and `purge_everything` was allowed, even though I only had set one of them. This has also been made in my latest change.

jamesog:
Looks good, thanks!

PR #15, opened 2016-04-19
Add support for /ips

jamesog:
A new API endpoint exposes CloudFlare's public IP ranges.

This endpoint does not require logging in to the API.


jamesog:
I started adding this to `flarectl` too but that expects the API environment variables to be set. Probably need to refactor that a bit to not require API keys for the `ips` sub-command.

PR #19, opened 2016-04-22
Add error handling to makeRequest()

Tenzer:
It's a best effort attempt at giving the most relevant error message possible.

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -11,6 +11,7 @@ package cloudflare
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


cloudflare.go
```diff
@@ -60,10 +61,17 @@ func (api *API) makeRequest(method, uri string, params interface{}) ([]byte, err
 	// req.Header.Add("Content-Type", "application/json")
 	client := &http.Client{}
 	resp, err := client.Do(req)
-	// TODO(jamesog): Check resp.StatusCode != http.StatusOK
-	// if resp.StatusCode != http.StatusOK
 	defer resp.Body.Close()
 	resBody, err := ioutil.ReadAll(resp.Body)
+	if resp.StatusCode != http.StatusOK {
```
jamesog: Should this come before we try to read the body?

Tenzer: According to my testing where I entered invalid credentials, the body was still set. I don't know if we have a good way of testing against some endpoint which has an empty body, but would assume we at least always can read an empty string from the response?

jamesog: Ah, you mean like if the API returned a 403 because the wrong email/API key pair were supplied? That makes sense.

jamesog:
LGTM.

PR #20, opened 2016-04-24
Add Page Rules

jamesog:
I've broken from the trend of adding the struct definitions in cloudflare.go and put them in pagerules.go with the code instead - I think this makes more sense.

If everyone thinks that works, I'll open another PR to move the existing structs.

### DIFF
--- a/cmd/flarectl/flarectl.go
+++ b/cmd/flarectl/flarectl.go
@@ -369,6 +369,55 @@ func zoneKeyless(*cli.Context) {
 func zoneRailgun(*cli.Context) {
 }
 
+func pageRules(c *cli.Context) {
+	if err := checkEnv(); err != nil {
+		fmt.Println(err)
+		return
+	}
+	if err := checkFlags(c, "zone"); err != nil {
+		return
+	}
+	zone := c.String("zone")
+
+	z, err := api.ListZones(zone)
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+
+	zid := z[0].ID
+
+	rules, err := api.ListPageRules(zid)
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+	fmt.Printf("%3s %-32s %-8s %s\n", "Pri", "ID", "Status", "URL")
+	for _, r := range rules {
+		var settings []string
+		fmt.Printf("%3d %s %-8s %s\n", r.Priority, r.ID, r.Status, r.Targets[0].Constraint.Value)
+		for _, a := range r.Actions {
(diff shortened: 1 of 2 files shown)


terinjokes:
LGTM 👍

elithrar:
LGTM, with the only caveat that `PageRulesActions` might be a little brittle if the keys for Page Rule options change (i.e. rare, but it *does* happen).

terinjokes:
@elithrar If they do, I'd file a bug against WWW. That shouldn't happen, especially not now that the API is public.

cmd/flarectl/flarectl.go
```diff
@@ -618,6 +667,26 @@ func main() {
...
 		{
+			Name:    "pagerules",
+			Aliases: []string{"p"},
+			Usage:   "Page Rules",
+			Subcommands: []cli.Command{
+				{
+					Name:    "list",
+					Aliases: []string{"l"},
+					Action:  pageRules,
+					Usage:   "List DNS records for a zone",
```
Tenzer: I am pretty sure this is incorrect :)

cmd/flarectl/flarectl.go
```diff
@@ -378,7 +427,7 @@ func main() {
 	app := cli.NewApp()
 	app.Name = "flarectl"
 	app.Usage = "CloudFlare CLI"
-	app.Version = "2015.12.0"
+	app.Version = "2016.4.0"
```
Tenzer: Should we consider using semantic versioning here? It's better at signifying the impact of the version bumps.

cmd/flarectl/flarectl.go
jamesog: Darn! Thanks, will fix that shortly.

cmd/flarectl/flarectl.go
jamesog: This is pseudo-semver :-) I've used our own standard for versioning, but happy to change it. This program is mostly an example anyway.

Tenzer: Right, I was thinking more in the scope of the library itself, I guess we don't really have any versioning for it right now?

jamesog: Yeah, that's true. I wasn't sure how to handle it for the library, other than using git tags. I'm open to ideas.

Tenzer: I think git tags should work. We can start with version 1.0.0 a bunch of the current structural changes are done and the library is more feature complete. In my eyes versioning is also as much about having a changelog so users of the library have a way of identifying any changes they need to be aware of, without having to look through commit histories and such.

Issue #18, opened 2016-04-22
Align function signatures


Tenzer:
Currently we have a couple of different ways of invoking the functions the library provides. For instance the DNS related functions all take the related zone as a string in the input, the WAF related functions take a zone ID as input and the cache purge and `ZoneDetails()` methods take a `Zone` struct as input.

We should probably come to some conclusion of how we want users to interact with the library, and make sure the functions follow that. It will also make it more clear how the function signatures of future functions will look as we get them implemented.

I am suggesting we take a `Zone` struct as input where possible, it will put the load of looking up the zone on the user, but also means the user is able to cache it however they want, keeping the library leaner and "less magical". Another approach would be to always work on the zone ID, since that is the bare minimum we would need in order to interact with the API. Thoughts?

jamesog:
Initially I wanted as little friction as possible for users of the library/API so thought it made more sense, in most cases, to pass in the zone name and have the function do the zone ID lookup for you. There were one or two cases where it seemed to make more sense to use the object instead.

I do agree that this needs cleaning up and made more consistent.

elithrar:
Q: What benefit does passing `Zone` as the function argument give (today, and in the future) if we require the user to do the name => ID lookup prior?

I 100% agree that rationalizing the function signatures is useful, and would be OK with a functions accepting a `zoneID string` instead of a `zoneName string`.

We can provide a helper function around the API—`ZoneByName(zoneName string) (string, error)` via `GET /zone?name=$name`—that returns the zoneID for the package user to cache locally and then pass to functions as needed.

Tenzer:
Accepting `Zone` instead of a zone ID would mean we could check if the plan of a zone allow an action to be made, without firing a request to the API. That's the only thing I could think of from the top of my head, and may not be enough reason to require users to use `Zone` everywhere.
I think accepting the zone ID as a string would be the best option then.

Tenzer:
This was done in #22.

PR #22, opened 2016-04-25
Refactoring of method signatures

Tenzer:
As discussed in #18, we should have a more consistent signature for the methods provided in the package, this is an attempt to implement that. It changes all methods interacting with the CF API to use a zone ID, if needed by the API.

It also adds the `ZoneIDByName()` convenience method to get a zone ID from a zone name, and makes use of that in `flarectl`.

/cc @jamesog @elithrar

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -35,6 +35,20 @@ func NewZone() *Zone {
 	return &Zone{}
 }
 
+// ZoneIDByName retrieves a zone's ID from the name.
+func (api *API) ZoneIDByName(zoneName string) (string, error) {
+	res, err := api.ListZones(zoneName)
+	if err != nil {
+		return "", err
+	}
+	for _, zone := range res {
+		if zone.Name == zoneName {
+			return zone.ID, nil
+		}
+	}
+	return "", errors.New("Zone could not be found")
+}
+
 // Params can be turned into a URL query string or a body
 // TODO: Give this func a better name
 func (api *API) makeRequest(method, uri string, params interface{}) ([]byte, error) {
--- a/cmd/flarectl/flarectl.go
+++ b/cmd/flarectl/flarectl.go
@@ -185,11 +185,18 @@ func zoneRecords(c *cli.Context) {
 		cli.ShowSubcommandHelp(c)
 		return
 	}
+
+	zid, err := api.ZoneIDByName(zone)
+	if err != nil {
+		fmt.Println(err)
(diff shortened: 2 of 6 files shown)


cloudflare.go
```diff
@@ -35,6 +35,20 @@ func NewZone() *Zone {
...
 }
 
+// ZoneIDByName retrieves a zone's ID from the name.
+func (api *API) ZoneIDByName(zoneName string) (string, error) {
+	res, err := api.ListZones(zoneName)
+	if err != nil {
+		return "", err
+	}
+	for _, zone := range res {
+		if zone.Name == zoneName {
```
jamesog: Might be worth checking `zone.Status` too. I'm not sure what would happen if the API user was part of a MU org and the zone existed in multiple accounts/orgs (e.g. if moving a zone between accounts).

jamesog:
I think `ZoneIDByName` should go in `zone.go` instead of `cloudflare.go` - what do you think?

cloudflare.go
Tenzer: Hmm, I'm not sure that's something I have an easy way of testing. Mostly I tried to improve upon the previous check of just taking the first zone returned by `ListZones`, but we can definitely add more checks as needed.

Tenzer:
I had that first but moved it to `cloudflare.go` since it doesn't call the actual API and is probably more of a general function in the library than just being related to zones (it will for instance also be used for DNS, WAF, etc. stuff).

elithrar:
Can I be a *huge* pain and suggest we use `zoneID` instead of `zid` and `recordID` instead of `rid`? Totally obvious to us but helps with clarity in some of the longer functions.

jamesog:
Yep, I think that's fair. Jeppe is just copying my lazy style, but as it's not in a loop where [short variables are generally OK](https://github.com/golang/go/wiki/CodeReviewComments#variable-names), we should make the names more descriptive.

Tenzer:
I am all for more descriptive names - as James say I based it on what already was in the code. I'll rebase it and push again.

Tenzer:
Actually, I ended up making a separate commit for it, it will make it easier to review this specific change.

jamesog:
LGTM 👍

jamesog:
@oholiab FYI - this changes the function signature of the WAF functions. You'll need to use `api. ZoneIDByName()` prior to calling the WAF functions.

PR #23, opened 2016-04-25
Use the correct field name for status of a zone in the Zone struct

Tenzer:
A silly little bug I found while testing the changes in #22.


elithrar:
Good catch. LGTM.

jamesog:
Yeah, that was a stupid bug. Should totally fire the guy that did that. 😒

PR #24, opened 2016-04-25
Move switch's default statement to the bottom for better readability

Tenzer:
The developer in question has stated this was due to stupidity, so let's try to correct that.


jamesog:
Mea culpa.

PR #10, opened 2016-03-26
[breaking] New() returns error; supports Options.

elithrar:
Ref: https://github.com/jamesog/cloudflare-go/issues/8

- New now has a variadic opts argument that accepts a slice of Option.
- Options for changing the underlying http.Client and http.Header map added.
- Updated flarectl to handle the updated New() signature.
- A new Error interface for returning different errors, and wrapping the
  cause, to provide easier inspection of errors without having to type-assert
  for all possible causes.

The `Error` interface will be expanded to have a `NetworkError` implementation as well for wrapping HTTP errors, and possibly a (breaking) change to handle parsing/API response errors.

We also now vendor deps for `flarectl`.

PR #25, opened 2016-04-26
Better and more consistent error handling by wrapping the errors

Tenzer:
This is based on Dave Cheney's [talk from Gocon 2016](http://dave.cheney.net/paste/gocon-spring-2016.pdf) and make use of the github.com/pkg/errors package to do error wrapping.

We no longer print error messages and then return them as that would lead to the same error being handling multiple times.

I have also added a missing error check inside makeRequest() after the HTTP request has been tried, plus added a couple of TODO lines in error cases where we return `err` but we already know it's `nil`.

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -2,8 +2,9 @@ package cloudflare
 
 import (
 	"encoding/json"
-	"fmt"
 	"net/url"
+
+	pkgErrors "github.com/pkg/errors"
 )
 
 /*
@@ -17,14 +18,12 @@ func (api *API) CreateDNSRecord(zoneID string, rr DNSRecord) error {
 	uri := "/zones/" + zoneID + "/dns_records"
 	res, err := api.makeRequest("POST", uri, rr)
 	if err != nil {
-		fmt.Println("Error with makeRequest")
-		return err
+		return pkgErrors.Wrap(err, "Error from makeRequest")
 	}
 	var r DNSRecordResponse
 	err = json.Unmarshal(res, &r)
 	if err != nil {
-		fmt.Println("Error with unmarshal")
-		return err
+		return pkgErrors.Wrap(err, "Error from unmarshal")
 	}
 	return nil
 }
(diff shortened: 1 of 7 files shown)


elithrar:
This is similar (in approach) to some of the work I was doing in this PR: https://github.com/cloudflare/cloudflare-go/pull/10/commits/fb9939d2839d2cac32742cec29b69b0ad4299b38#diff-4aae454123f50b906f9aaef44f10011eR2 - which is based around how `net.Error` is an interface with methods for asserting behaviour.

Tenzer:
I think it's two different things which can perfectly co-exist. Wrapping of errors seems like a good way to provide context of the error as it's passed up the stack (although the error messages in this PR probably could be improved over time), while your approach is related to developers using this library who needs to handle errors.

dns.go
```diff
@@ -17,14 +18,12 @@ func (api *API) CreateDNSRecord(zoneID string, rr DNSRecord) error {
...
-		fmt.Println("Error with makeRequest")
-		return err
+		return pkgErrors.Wrap(err, "Error from makeRequest")
 	}
 	var r DNSRecordResponse
 	err = json.Unmarshal(res, &r)
 	if err != nil {
-		fmt.Println("Error with unmarshal")
-		return err
+		return pkgErrors.Wrap(err, "Error from unmarshal")
```
elithrar: Make this `const errUnmarshalError`?

dns.go
```diff
@@ -76,12 +75,12 @@ func (api *API) DNSRecord(zoneID, recordID string) (DNSRecord, error) {
 	uri := "/zones/" + zoneID + "/dns_records/" + recordID
 	res, err := api.makeRequest("GET", uri, nil)
 	if err != nil {
-		return DNSRecord{}, err
+		return DNSRecord{}, pkgErrors.Wrap(err, "Error from makeRequest")
```
elithrar: Same here. Can re-use the constant elsewhere.

elithrar:
LGTM besides the couple of comments re: making those error messages private constants. This will definitely help make errors clearer to callers + tie in with my changes in #10!

Tenzer:
The repeated error messages have now been put into constants in 18ff40f.

elithrar:
Thanks @Tenzer 💯

PR #26, opened 2016-04-30
[refactor] Replace stdlib errors pkg w/ pkg/errors

elithrar:
@Tenzer I realised that `github.com/pkg/errors` provides an `errors.New` function synonymous with the std. lib's `errors.New` after merging.


Tenzer:
Neat! That's a lot cleaner :)

PR #28, opened 2016-05-03
[breaking] CreateDNSRecord returns the response.

elithrar:
Addresses https://github.com/cloudflare/cloudflare-go/issues/27

PR #29, opened 2016-05-03
return the created DNS record with creation

mitchellh:
This is a breaking API change.

The CreateDNSRecord API wasn't returning the DNSRecord which meant that there was no way to safely retrieve the ID of the created record. This is necessary to make further requests.

We're currently using my fork for Terraform with this change since we require this functionality. If you want me to change it in any way to get it in we'd be happy to.


elithrar:
See #28 - which also includes the requisite changes in `cmd/flarectl`.

PR #53, opened 2016-05-05
[bugfix] CreateDNSRecord wasn't being passed a pointer.

### LINKED ISSUE #52: JSON Unmarshal error for CreateDNSRecord
I am trying implementing a dns add service function in my go program so I referred to flarectl and dns.go

	record := cf.DNSRecord{
		Name:    r.Name,
		Type:    strings.ToUpper(r.Type),
		Content: r.Content,
		TTL:     int(r.TTL),
		Proxied: r.Proxy,
	}

`if res, err := d.api.CreateDNSRecord(zoneId, record); err != nil {
		return nil, err`

I get `Error unmarshalling JSON: json: Unmarshal(nil *cloudflare.DNSRecordResponse)` when running a unit test. I double-checked that I am passing everything correctly by looking at flarectl.go. 

DNS record is actually created. Verified by using curl api.

Any help or insights are greatly appreciated!

elithrar:
Fixes #52

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -22,7 +22,7 @@ func (api *API) CreateDNSRecord(zoneID string, rr DNSRecord) (*DNSRecordResponse
 	}
 
 	var recordResp *DNSRecordResponse
-	err = json.Unmarshal(res, recordResp)
+	err = json.Unmarshal(res, &recordResp)
 	if err != nil {
 		return nil, errors.Wrap(err, errUnmarshalError)
 	}


Tenzer:
We should probably try to get some unit tests set up :/

jamesog:
I agree. @terinjokes was working on a mock API server for the Node.js implementation. We should probably try and do the same here.

terinjokes:
I have a mock server in Go for my work project that I'm fairly happy with.
I can put together an example of that in a PR here.

PR #61, opened 2016-05-07
[feature] Origin CA Support

iToto:
Lots of repeated code in the client - could benefit from a refactor. But a good starting point I think :)


origin_ca.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+// OriginCertificateID is the resource sent back for a revoke action
+type OriginCertificateID struct {
+	ID string `json:"id"`
+}
+
+// OriginCAResponse is the APIv4 response envelop containing the OriginCA result
+type OriginCAResponse struct {
+	Success  bool     `json:"success"`
```
elithrar: Embed `Response` here to avoid repeating `Success` `Errors` and `Messages` each struct. Ref: https://github.com/cloudflare/cloudflare-go/blob/master/cloudflare.go#L152-L155

.travis.yml
```diff
@@ -21,3 +21,4 @@ notifications:
     recipients:
       - jamesog@cloudflare.com
       - msilverlock@cloudflare.com
+      - salvatore@cloudflare.com
```
elithrar: 👍

elithrar:
Couple of minor changes to catch up to the moving target that is this library! Thanks for submitting this though—esp. for the tests.

cloudflare.go
```diff
@@ -15,6 +15,7 @@ const apiURL = "https://api.cloudflare.com/client/v4"
 
 // API holds the configuration for the current API client. A client should not
 // be modified concurrently.
+const errRequestNotSuccessfull = "Error reported by API"
```
Tenzer: You should probably add this to `errors.go` when you are rebasing. You also have an "L" to much at the end of the constant name.

iToto:
is there anything still outstanding here?

elithrar:
@itoto I don't think so! I didn't get notifications for your added commits. Will review today/tomorrow and merge—thanks!

errors.go
```diff
@@ -2,9 +2,10 @@ package cloudflare
 
 // Error messages
 const (
-	errEmptyCredentials = "invalid credentials: key & email must not be empty"
-	errMakeRequestError = "error from makeRequest"
-	errUnmarshalError   = "error unmarshalling the JSON response"
+	errEmptyCredentials     = "invalid credentials: key & email must not be empty"
+	errMakeRequestError     = "error from makeRequest"
+	errUnmarshalError       = "error unmarshalling the JSON response"
+	errRequestNotSuccessful = "Error reported by API"
```
Tenzer: Nitpick: In order to match the established pattern, the error message should start with a lower case letter.

origin_ca.go
```diff
@@ -0,0 +1,147 @@
...
+
+import (
+	"encoding/json"
+
+	"github.com/pkg/errors"
+)
+
+// OriginCA resource
+// This is the origin_ca resource definition (as defined here:
+// https://api.cloudflare.com/#origin-ca || https://api.cloudflare.com/#cloudflare-ca
```
Tenzer: It seems like the `#origin-ca` anchor no longer works, so we might as well remove it.

origin_ca.go
```diff
@@ -0,0 +1,147 @@
+package cloudflare
+
+import (
+	"encoding/json"
+
+	"github.com/pkg/errors"
+)
+
+// OriginCA resource
+// This is the origin_ca resource definition (as defined here:
```
Tenzer: I'm not quite sure why the opening parenthesis is doing in there :)

origin_ca.go
Tenzer: Actually, it seems like all the references pointing at `#origin-ca...` has been changed to `#cloudflare-ca...` in the documentation.

origin_ca.go
```diff
@@ -0,0 +1,147 @@
...
+// https://api.cloudflare.com/#origin-ca || https://api.cloudflare.com/#cloudflare-ca
+type OriginCA struct {
+	ID              string   `json:"id"`
+	Certificate     string   `json:"certificate"`
+	Hostnames       []string `json:"hostnames"`
+	ExpiresOn       string   `json:"expires_on"`
```
Tenzer: I think it makes better sense to return dates as `time.Time` types, I have made #68 to change it for the rest of the code.

Tenzer:
There's a bit of a blocker here: The Origin CA doesn't support the existing authentication method the rest of the methods use. It instead uses the `X-Auth-User-Service-Key` header with a specific token for the Origin CA actions.

We can probably make `api.makeRequest` support this extra header if it's set by the user, but we still need a way to let the user set this token - and also support doing requests to the Origin CA endpoints without the usual email and key fields being set.

origin_ca.go
```diff
@@ -0,0 +1,147 @@
...
+		return nil, errors.Wrap(err, errUnmarshalError)
+	}
+
+	if !originResponse.Success {
+		return nil, errors.New(errRequestNotSuccessful)
+	}
+
+	createdCert := originResponse.Result
+
+	return &createdCert, nil
```
Tenzer: Couldn't this line and the one above just be rewritten to: `return &originResponse.Result, nil`? That's also what is being done in the `OriginCertificate` method.

stretchkennedy:
What ended up happening with this PR? What work would need to be done to get it merged?

jamesog:
I'd forgotten this PR existed so had started working on this feature (along with implementing User-Service auth). I've got a mostly-working thing for User-Service, however I can't test this right now; it looks like GET /certificates is doing the wrong thing (or works differently from the API docs).

origin_ca.go
```diff
@@ -0,0 +1,146 @@
...
+// This is the origin_ca resource definition as defined here:
+// https://api.cloudflare.com/#cloudflare-ca
+type OriginCA struct {
+	ID              string    `json:"id"`
+	Certificate     string    `json:"certificate"`
+	Hostnames       []string  `json:"hostnames"`
+	ExpiresOn       time.Time `json:"expires_on"`
+	RequestType     string    `json:"request_type"`
+	RequestValidity string    `json:"requested_validity"`
+	Csr             string    `json:"csr"`
```
jamesog: Should be `CSR`.

origin_ca.go
```diff
@@ -0,0 +1,146 @@
...
+	"encoding/json"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// OriginCA resource
+// This is the origin_ca resource definition as defined here:
+// https://api.cloudflare.com/#cloudflare-ca
+type OriginCA struct {
```
jamesog: Might want to call this `Certificate` since it describes a certificate and not the CA itself. :-)

origin_ca.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// OriginCAResponse is the APIv4 response envelop containing the OriginCA result
+type OriginCAResponse struct {
+	Response
+	Result OriginCA `json:"result"`
+}
+
+// OriginCAResponseList is the APIv4 response envelop containing a listof OriginCA result
+type OriginCAResponseList struct {
```
jamesog: Ditto here.

origin_ca.go
```diff
@@ -0,0 +1,146 @@
...
+	Response
+	Result     []OriginCA `json:"result"`
+	ResultInfo ResultInfo `json:"result_info"`
+}
+
+// OriginCAResponseRevoke is the APIv4 response envelop containing a revoked cert id
+type OriginCAResponseRevoke struct {
```
jamesog: And here.

origin_ca.go
```diff
@@ -0,0 +1,146 @@
...
+	Csr             string    `json:"csr"`
+}
+
+// OriginCertificateID is the resource sent back for a revoke action
+type OriginCertificateID struct {
+	ID string `json:"id"`
+}
+
+// OriginCAResponse is the APIv4 response envelop containing the OriginCA result
+type OriginCAResponse struct {
```
jamesog: I don't think this response struct needs to be public.

jamesog:
I just created #104 to add User Service authentication.

iToto:
@jamesog any update on this?

jamesog:
Ah, sorry. Just 1.5 more things:

1) Can you change all the method signatures to match how they were changed in #115 please? (you're just missing an extra blank line)
1.5) This is otherwise fine; however I'm still puzzled how the user-service auth works. My change in #104 means it's possible to flip between key+email or user-service, but I'm still not sure if that should be done by the user or the method. I'm leaning towards the latter, I think. What do you think?

iToto:
Yeah, I also think it should be the latter. We shouldn't leave it up to the user to decide what type of auth to use.

jamesog:
Alright. Do you want to update the methods to call `api.SetAuthType(AuthUserService)`? I'm also thinking we should fetch the current value before doing that and resetting it when we're done.

jamesog:
This is now done in #136.

Issue #9, opened 2016-03-26
Resolve golint Suggestions


elithrar:
I'll probably tackle this, but I'll post it here for posterity:

```go
~/.go/src/github.com/jamesog/cloudflare-go(msilverlock/breaking-options ✗) golint
cloudflare.go:49:1: comment on exported function NewZone should be of the form "NewZone ..."
cloudflare.go:199:6: exported type ZoneResponse should have comment or be unexported
cloudflare.go:206:6: exported type ZonePlanResponse should have comment or be unexported
cloudflare.go:227:6: exported type ZoneSetting should have comment or be unexported
cloudflare.go:235:6: exported type ZoneSettingResponse should have comment or be unexported
cloudflare.go:242:1: comment on exported type DNSRecord should be of the form "DNSRecord ..." (with optional leading article)
cloudflare.go:261:1: comment on exported type DNSRecordResponse should be of the form "DNSRecordResponse ..." (with optional leading article)
cloudflare.go:269:1: comment on exported type DNSListResponse should be of the form "DNSListResponse ..." (with optional leading article)
cloudflare.go:277:1: comment on exported type ZoneRailgun should be of the form "ZoneRailgun ..." (with optional leading article)
cloudflare.go:285:6: exported type ZoneRailgunResponse should have comment or be unexported
cloudflare.go:292:1: comment on exported type ZoneCustomSSL should be of the form "ZoneCustomSSL ..." (with optional leading article)
cloudflare.go:308:6: exported type ZoneCustomSSLResponse should have comment or be unexported
cloudflare.go:315:6: exported type KeylessSSL should have comment or be unexported
cloudflare.go:327:6: exported type KeylessSSLResponse should have comment or be unexported
cloudflare.go:334:6: exported type Railgun should have comment or be unexported
cloudflare.go:353:6: exported type RailgunResponse should have comment or be unexported
cloudflare.go:360:1: comment on exported type CustomPage should be of the form "CustomPage ..." (with optional leading article)
cloudflare.go:371:6: exported type CustomPageResponse should have comment or be unexported
cloudflare.go:378:1: comment on exported type WAFPackage should be of the form "WAFPackage ..." (with optional leading article)
cloudflare.go:389:6: exported type WAFPackagesResponse should have comment or be unexported
cloudflare.go:400:6: exported type WAFRule should have comment or be unexported
cloudflare.go:414:6: exported type WAFRulesResponse should have comment or be unexported
[...]

PR #63, opened 2016-05-07
Run golint in Travis

Tenzer:
When I try to run `golint` with the syntax used in the other tests it seems to fail:
```
$ golint -set_exit_status $(go list ./... | grep -v '/vendor/')
open github.com/cloudflare/cloudflare-go: no such file or directory
open github.com/cloudflare/cloudflare-go/cmd/flarectl: no such file or directory
```
Using a bit of shell scripting seems to make it happier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,14 @@ matrix:
   allow_failures:
     - go: tip
 
+install:
+  - if [[ $TRAVIS_GO_VERSION == 1.6* ]]; then go get github.com/golang/lint/golint; fi
+
 script:
   - go get -t -v $(go list ./... | grep -v '/vendor/')
-  - diff -u <(echo -n) <(gofmt -d .)
-  - go vet $(go list ./... | grep -v '/vendor/')
+  - if [[ $TRAVIS_GO_VERSION == 1.6* ]]; then diff -u <(echo -n) <(gofmt -d .); fi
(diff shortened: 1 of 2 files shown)


Tenzer:
When this is merged we should be able to close #9.

jamesog:
Looks like golint and Go 1.4 don't get along any more.

Tenzer:
Yeah, I wonder if we can skip golint on Go 1.4 specifically?

jamesog:
Travis sets the environment variable `TRAVIS_GO_VERSION` - might be able to put an `if` block around that to skip for 1.4. From skimming the docs I can't obviously see a better way - doesn't look like the build matrix lets you exclude an install step.

Tenzer:
Thinking about it, I think it makes sense to skip the `gofmt` `go vet` and `golint` steps when we run versions other than 1.6 (or whatever the latest stable version is), as those tests only will produce the same errors across versions. The idea with testing in different versions is only to test if the code can compile and `go test` doesn't complain.

Tenzer:
It seems to work correctly now. We could probably consider adding a shell script in the repository with the static code tests we are doing in order to skip some of all the if statements in `.travis.yml`.

elithrar:
LGTM. A shame we can't just have `$TRAVIS_GO_LATEST` to avoid having to update this, but alas.

PR #66, opened 2016-05-10
[ref] Split out makeRequest. Reuse http.Clients.

elithrar:
- The http.Client is re-used across requests, enabling the use of keepalives
  from the underlying http.Transport.
- request creates the http.Request, adds headers and makes the API call,
  returning the *http.Response for further consumption. request should be used
  for API calls with large, streaming response bodies where reading the body
  into memory is not suitable.
- makeRequest wraps request; marshalling parameters to JSON, reading the
  response body and handling standard HTTP error codes.

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -69,20 +69,54 @@ func (api *API) ZoneIDByName(zoneName string) (string, error) {
 	return "", errors.New("Zone could not be found")
 }
 
-// Params can be turned into a URL query string or a body
-// TODO: Give this func a better name
+// makeRequest makes a HTTP request and returns the body as a byte slice,
+// closing it before returnng. params will be serialized to JSON.
 func (api *API) makeRequest(method, uri string, params interface{}) ([]byte, error) {
 	// Replace nil with a JSON object if needed
 	var reqBody io.Reader
 	if params != nil {
 		json, err := json.Marshal(params)
 		if err != nil {
-			return nil, errors.Wrap(err, "Error marshalling params to JSON")
+			return nil, errors.Wrap(err, "error marshalling params to JSON")
 		}
 		reqBody = bytes.NewReader(json)
 	} else {
 		reqBody = nil
 	}
+
+	resp, err := api.request(method, uri, reqBody)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


terinjokes:
We should probably consider accepting `golang.org/x/net/context` and using `golang.org/x/net/context/ctxhttp` to be ready for Go 1.7.

elithrar:
@terinjokes Did you have a use-case in mind for (e.g.) `http.Request.Context()` coming in 1.7? Passing values in the request probably isn't that useful for this lib.

https://godoc.org/golang.org/x/net/context/ctxhttp#Do might be useful for allowing users to set timeouts, but since we support passing a custom `http.Client` with your own `client.Timeout`, the major benefit of using `ctxhttp` is probably nullified.

(I might be missing something though!)

elithrar:
Note: `http.Request.WithContext` could potentially be used to cancel outgoing requests. We would need to expose that for each API call, or accept a context when creating a new `*API` that is propagated to each request and allows cancellation of all outgoing requests. 

Are there other solutions that enable a friendly public API?

terinjokes:
To use the cancellation support (either directly, or as part of WithTimeout/Duration) or to tie this request to an incoming request (IE, if I'm proxying as part of access control)

The Go documentation recommends against storing context on a struct, we should accept them as part of each method.

elithrar:
> The Go documentation recommends against storing context on a struct, we should accept them as part of each method.

Somewhat: that was walked back a bit when `context.Context` support was added to `http.Request` itself. 

I'm also hesitant to have every API method accept a `context.Context` — it's not likely to be a common use-case, since our API calls are short-lived (the need to cancel them before a timeout is rare) and it makes the API less friendly - e.g.

```diff
- func (api *API) CreatePageRule(zoneID string, rule PageRule) error
+ func (api *API) CreatePageRule(ctx context.Context, zoneID string, rule PageRule) error
```

Most users would end up calling `err := api.CreatePageRule(nil, zid, rule)` instead. I don't see another way to allow per-API-call contexts without adding overhead to the public API.

@jamesog @Tenzer - thoughts?

Tenzer:
I don't think being able to provide a context is big enough of a win compared to the downside of cluttering the API. If a trend starts forming later on (after Go 1.7 release) then we can re-evaluate. In my opinion, we should aim towards supplying an API which is familiar for Go developers to use, rather than aim for feature completeness (talking about Go features, not CloudFlare features).

jamesog:
I think this looks fine as it is.

PR #68, opened 2016-05-14
Provide dates sent from API as time.Time types instead of strings

Tenzer:
This means users of the library don't have to worry about what date format the API provides dates in and make it easier for people to use the dates in their applications.

I have also snug in a slight clean up of some unused code.


jamesog:
LGTM 👍

elithrar:
LGTM. Merging!

PR #69, opened 2016-05-15
[docs] Update Getting Started code

nishanths:
- Include missing package imports.
- Use log.Fatal where appropriate.
- Use user details endpoint instead of zones. A freshly created
  CloudFlare account with no sites added will not recieve visual feedback
  from running the example as there are no zones to range over. However,
  user details should exist even on a fresh account.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,15 +70,19 @@ func main() {
 	// Print user details
 	fmt.Println(u)
 
-	// Fetch the list of zones on the account
-	zones, err := api.ListZones()
+	// Fetch the zone ID
+	id, err := api.ZoneIDByName("example.com") // Assuming example.com exists in your CloudFlare account already
 	if err != nil {
 		log.Fatal(err)
 	}
-	// Print the zone names
-	for _, z := range zones {
-		fmt.Println(z.Name)
+
+	// Fetch zone details
+	zone, err := api.ZoneDetails(cloudflare.Zone{ID: id})
+	if err != nil {
+		log.Fatal(err)
 	}
+	// Print zone details
+	fmt.Println(zone)
 }
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,6 +47,8 @@ package main
```
elithrar: While you're here: might as well append `go` to the three backticks on line 45 to get the correct syntax highlighting.

nishanths: Done.

elithrar:
Thanks for this. One thing I'd like to retain (or demonstrate in a different way) is a zone interaction, given that adding/modifying zones is a core part of most customers' automation.

If you're willing to, extend the example to show the lookup of a `Zone` by name only - e.g. `ZoneIDByName` and then passing that into the `Zone` type to fetch the details for that zone. Use `example.com` as the zone name.

nishanths:
Sure thing, apologies for not thinking of it from that perspective. I'll get to it in a bit. 👍

nishanths:
Calling `ZoneIDByName` for a non-existent zone name on the account (for instance `example.com`) returns a zone-not-found error. So I've just put the old zones example code back.

elithrar:
Which is fine: I don't think the example has to be purely for a "fresh"
account.

A small docstring that says "// Assuming example.com exists in your
CloudFlare account already" would be enough.

Most users set up at least one zone via the UI before then using the API.

nishanths:
@elithrar Awesome, updated! Let me know if this looks alright. 
Implemented ZoneDetails at #70

README.md
```diff
@@ -55,17 +57,32 @@ var api *cloudflare.API
```
elithrar: Sorry, I only just saw this in (full) preview (doesn't show in the diff): line 56 - `var api *cloudflare.API` - is shadowed by line 60, so you can just remove line 56 for now.

nishanths: Good catch 👍

elithrar:
Thanks again @nishanths!

PR #71, opened 2016-05-16
Upping our badges game

Tenzer:
I am just throwing this out here as a suggestion...

PR #72, opened 2016-05-16
Switch to a flat-square style for badges in the README

Tenzer:
@terinjokes As discussed in #71, although I just placed them on their own separate lines, and let Markdown join the lines together. I think this is a slightly neater solution than having the references one place and the URLs another.


elithrar:
👌🏼

terinjokes:
LGTM

Issue #73, opened 2016-05-21
flarectl can't run without credentials provided


Tenzer:
The error handling made in 586e5a55, means that `flarectl` no longer can run without the environment variables with credentials being set to at least something. This doesn't give the best user experience, since neither `flarectl --help` or `flarectl ips` work any more, even though none of them make API calls which use credentials.

We may have to change `flarectl` so `cloudflare.New()` only will be called by the functions that actually need the authenticated API to function. A helper method can be introduced in order to minimise the code duplication. Any other suggestions?

jamesog:
Fixed in 7bbb639.

PR #74, opened 2016-05-21
flarectl: Move api invocation to checkEnv()

jamesog:
Since 586e5a5 `cloudflare.New()` will error if the API email and key are
not defined. Since flarectl was instantiating the API object at the
start, this meant that flarectl would not operate at all without having
the environment variables defined.

This moves the API object instantiation to `checkEnv()` so that, e.g.
`ips` and `--help` will still work. All other functions already call
`checkEnv()`.

Refs #73


elithrar:
LGTM

PR #81, opened 2016-05-31
Railgun and Zone Railgun

nishanths:


railgun.go
```diff
@@ -1,47 +1,311 @@
...
+}
+
+// railgunResponse represents the response from the Create Railgun and the Railgun Details endpoints.
+type railgunResponse struct {
+	Response
+	Result Railgun `json:"result"`
+}
+
+// railgunsResponse represents the response from the List Railguns endpoint.
+type railgunsResponse struct {
```
elithrar: `railgunListResponse` perhaps?

railgun.go
```diff
@@ -1,47 +1,311 @@
...
+	ResponseStatus  string `json:"response_status"`
+	Protocol        string `json:"protocol"`
+	ElapsedTime     string `json:"elapsed_time"`
+	BodySize        string `json:"body_size"`
+	BodyHash        string `json:"body_hash"`
+	MissingHeaders  string `json:"missing_headers"`
+	ConnectionClose bool   `json:"connection_close"`
+	Cloudflare      string `json:"cloudflare"`
+	CFRay           string `json:"cf-ray"`
+	// XXX: CloudFlare's online API documentation does not yet have definitions
```
elithrar: s/XXX/NOTE

railgun.go
```diff
@@ -1,47 +1,311 @@
...
+	// for the following fields. See: https://api.cloudflare.com/#railgun-connections-for-a-zone-test-railgun-connection/
+	CFWANError    interface{} `json:"cf-wan-error"`
```
elithrar: `string`

railgun.go
```diff
@@ -1,47 +1,311 @@
...
+	CFCacheStatus interface{} `json:"cf-cache-status"`
```
elithrar: `string`

elithrar:
Couple of minor comments, but otherwise looks good to me. 💯 for including tests.

railgun.go
nishanths: Previously we've used the plural form for similar structs. Still change to `railgunListResponse`?

- https://github.com/cloudflare/cloudflare-go/blob/master/zone.go#L71
- https://github.com/cloudflare/cloudflare-go/blob/master/ssl.go#L33

nishanths:
@elithrar Addressed the comments!

railgun.go
elithrar: Good point. It's not a major issue either way since it's private—we can change it later if need be!

elithrar:
LGTM. @jamesog and/or @jamesog to review and we can merge!

Issue #83, opened 2016-06-13
Very verbose HTTP error messages


benjojo:
When a HTTP error happens on the HTTP error message, the error string that comes back from the message contains the whole HTML error, This means the error messages are long enough to leave a nasty surprise (as I just discovered) : 

![screen shot 2016-06-13 at 15 19 34](https://cloud.githubusercontent.com/assets/1504626/16010700/4f85be1a-317a-11e6-8b3f-204639bf59f0.png)

benjojo:
The line that produces it is: https://github.com/cloudflare/cloudflare-go/blob/2c441b766adb2201bf54f31a43d3c2b016140d81/cloudflare.go#L106

What should we do, Clip the content (still, I don't know what useful info there is in there if the status is not something we expect) or just not post the content at all?

Tenzer:
It's a bit of a difficult situation. We basically have two choices:

1. We can do what we do now, which is to conclude that something went wrong (HTTP status 504), and since we got some content back in the response, we might as well include that in the error in order to provide as much context as possible.
2. We can check if the status code either is unknown or on a pre-defined list, and the make up our own error message saying something along the lines of "An unknown error occurred, HTTP status 504 received", which isn't necessarily very helpful to the user.

A potential third option could be to go with open 1, but then limit the size of the body in the response to a certain number of bytes/characters, but I can't quite figure out of that's better or worse than the above suggestions.

benjojo:
Surely there should be no reason for a user to get a error message like that in general (provided api.cloudflare.com is online), As I see in that code we already look at the some error messages, We could catch things like 522,523,524,504,500 and return a error object "Provider fault" or something like that, For other error messages I would agree, it may be good to provide context, but there is no point bombing HTML page in someones error log

Tenzer:
Fixed via #84.

PR #84, opened 2016-06-14
Handle API backend failures with a better error message (#83)

benjojo:
Everyone has bad days, CloudFlare is not immune to that, however
currently when CloudFlare's own API has a bad day, currently a
very verbose error message is returned that can cause very
large log lines ( #83 ) .

To address this we catch HTTP 522,523,524, and 504 with a
"Service Failure" message and do not include the responce
body back with it, since there is no need to return it
as it will most likely not be useful to anyone.


elithrar:
LGTM.

PR #87, opened 2016-07-13
[WIP] Add and fix some API methods to manage SSL, zones, and page rules

keshavdv:
This includes some changes that I had to make while working on implementing additional CF resources for Terraform and still needs tests.

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"io"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"time"
 
@@ -72,6 +73,7 @@ func (api *API) makeRequest(method, uri string, params interface{}) ([]byte, err
 	var reqBody io.Reader
 	if params != nil {
 		json, err := json.Marshal(params)
+		log.Printf("[DEBUG] Request is %s", string(json))
 		if err != nil {
 			return nil, errors.Wrap(err, "error marshalling params to JSON")
 		}
@@ -108,6 +110,7 @@ func (api *API) makeRequest(method, uri string, params interface{}) ([]byte, err
 		}
 		return nil, errors.Errorf("HTTP status %d: content %q", resp.StatusCode, s)
 	}
+	log.Printf("[DEBUG] Response is: %s", string(body))
 
 	return body, nil
 }
--- a/pagerules.go
+++ b/pagerules.go
@@ -2,6 +2,7 @@ package cloudflare
 
 import (
 	"encoding/json"
+	"strconv"
 	"time"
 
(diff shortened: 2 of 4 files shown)


cloudflare.go
```diff
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"io"
 	"io/ioutil"
+	"log"
```
elithrar: We definitely shouldn't log in the library (it breaks expectations around how applications are handling their own logs). If you need debug logs, we should bubble that up in the errors, or you can provide a custom `http.Client` that logs the request/response separately.

keshavdv: I completely agree -- this was left over from my debugging in terraform. I'll get rid of it as I finish adding some more functions and tests. Thanks for the quick feedback!

ssl.go
```diff
@@ -16,9 +15,9 @@ type ZoneCustomSSL struct {
 	Status        string     `json:"status"`
 	BundleMethod  string     `json:"bundle_method"`
 	ZoneID        string     `json:"zone_id"`
-	UploadedOn    time.Time  `json:"uploaded_on"`
```
Tenzer: We would like to keep dates as `time.Time` types instead of strings, as they are easier to work with in Go. Is there any specific reason for this change?

keshavdv: When I tested this, the format of the time string returned from Cloudflare ("2016-07-06 21:27:40.124575+00") did not match what was mentioned in the API docs ("2014-01-01T05:20:00Z"), causing this parsing to fail.

pagerules.go
```diff
@@ -76,12 +77,14 @@ var PageRuleActions = map[string]string{
...
 
+type MaybeInt int
+
 // PageRule describes a Page Rule.
 type PageRule struct {
 	ID         string           `json:"id,omitempty"`
 	Targets    []PageRuleTarget `json:"targets"`
 	Actions    []PageRuleAction `json:"actions"`
-	Priority   int              `json:"priority"`
+	Priority   MaybeInt         `json:"priority"`
```
jamesog: Why the custom type?

zone.go
```diff
@@ -497,6 +502,34 @@ func (api *API) ZoneAnalyticsByColocation(zoneID string, options ZoneAnalyticsOp
 
 // Zone Settings
 // https://api.cloudflare.com/#zone-settings-for-a-zone-get-all-zone-settings
+func (api *API) GetZoneSettings(zoneID string) ([]ZoneSetting, error) {
```
jamesog: Can you run `golint` and add method comments to any public methods please? Take a look at other methods we've recently updated for an example.

PR #89, opened 2016-08-02
API.request: Clone base headers to be thread safe

### LINKED ISSUE #88: [bug] API client is not thread safe
### Current behavior

Crashes when performing multiple api calls using same `*cloudflare.API` client

### Expected behavior

Safely perform multiple api calls using the same client in parallel

### Explanation

`*cloudflare.API.request` is not thread safe and will crash when multiple api requests are made from different goroutines . (This obviously holds true for all of the higher level methods that call it).

The problem happens because we reuse the same `http.Header` struct across all requests, and `http.Header.Set` are essentially writes to the underlaying `map[string][]string` which are not thread safe.

AaronO:
This PR makes `cloudflare-go` thread safe and fixes #88

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -122,7 +122,7 @@ func (api *API) request(method, uri string, reqBody io.Reader) (*http.Response, 
 	}
 
 	// Apply any user-defined headers first.
-	req.Header = api.headers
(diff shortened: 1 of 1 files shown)


elithrar:
Thanks for this!

Issue #49, opened 2016-05-03
Move structs from cloudflare.go to relevant files


Tenzer:
As mentioned in #20 we want to have the structs moved to the file where they are relevant.

jamesog:
Done in #92 :-)

PR #92, opened 2016-09-17
Tidy Structs

jamesog:
This tidies all of the struct definitions into their respective files per #49.

I was debating whether to squash the commits but I think having the individual commits makes more sense. Happy to squash if others aren't that bothered.


Tenzer:
LGTM

elithrar:
LGTM as well.

Issue #30, opened 2016-05-03
Implement User update


Tenzer:
https://api.cloudflare.com/#user-update-user

gliptak:
@Tenzer Can this be closed? Thanks

jamesog:
It can!

PR #96, opened 2016-11-09
Implement UpdateUser call

gliptak:
https://github.com/cloudflare/cloudflare-go/issues/30

Is there a recommended way to test against the "real" API?

### DIFF
--- a/user.go
+++ b/user.go
@@ -9,20 +9,20 @@ import (
 
 // User describes a user account.
 type User struct {
-	ID            string         `json:"id"`
-	Email         string         `json:"email"`
-	FirstName     string         `json:"first_name"`
-	LastName      string         `json:"last_name"`
-	Username      string         `json:"username"`
-	Telephone     string         `json:"telephone"`
-	Country       string         `json:"country"`
-	Zipcode       string         `json:"zipcode"`
-	CreatedOn     time.Time      `json:"created_on"`
-	ModifiedOn    time.Time      `json:"modified_on"`
-	APIKey        string         `json:"api_key"`
-	TwoFA         bool           `json:"two_factor_authentication_enabled"`
-	Betas         []string       `json:"betas"`
-	Organizations []Organization `json:"organizations"`
+	ID            string         `json:"id,omitempty"`
+	Email         string         `json:"email,omitempty"`
+	FirstName     string         `json:"first_name,omitempty"`
+	LastName      string         `json:"last_name,omitempty"`
+	Username      string         `json:"username,omitempty"`
+	Telephone     string         `json:"telephone,omitempty"`
+	Country       string         `json:"country,omitempty"`
+	Zipcode       string         `json:"zipcode,omitempty"`
+	CreatedOn     *time.Time     `json:"created_on,omitempty"`
+	ModifiedOn    *time.Time     `json:"modified_on,omitempty"`
+	APIKey        string         `json:"api_key,omitempty"`
+	TwoFA         bool           `json:"two_factor_authentication_enabled,omitempty"`
+	Betas         []string       `json:"betas,omitempty"`
+	Organizations []Organization `json:"organizations,omitempty"`
 }
 
 // UserResponse wraps a response containing User accounts.
@@ -54,7 +54,17 @@ func (api *API) UserDetails() (User, error) {
 // API reference:
 // 	https://api.cloudflare.com/#user-update-user
 //	PATCH /user
-func (api *API) UpdateUser() (User, error) {
-	// api.makeRequest("PATCH", "/user", user)
-	return User{}, nil
+func (api *API) UpdateUser(user *User) (User, error) {
+	var r UserResponse
+	res, err := api.makeRequest("PATCH", "/user", user)
+	if err != nil {
+		return User{}, errors.Wrap(err, errMakeRequestError)
+	}
+
+	err = json.Unmarshal(res, &r)
+	if err != nil {
+		return User{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return r.Result, nil
 }
--- a/user_test.go
+++ b/user_test.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"io/ioutil"
 )
 
 func TestUser_UserDetails(t *testing.T) {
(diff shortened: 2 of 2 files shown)


gliptak:
@Tenzer Please review.

Tenzer:
Could we not make `UpdateUser()` accept a `User` struct? That would make it easier to make changes to a user as the flow then would be:

1. Fetch the current user details with `UserDetails()`.
2. Update the fields that needs updating in the returned `User` struct.
3. Pass that struct to `UpdateUser()` which then internally can map things around to a new struct which is used for the request to the API, if necessary.

I think that would cater more for the typical flow developers would use when they would use this function.

gliptak:
@Tenzer I updated the patch based on your comment

Tenzer:
Thanks @gliptak! I do fail to see the point of splitting the functionality into two different functions now, as `UpdateUser()` basically doesn't do anything different than `updateUser()` other wise does - they take the same parameters and returns the same types. I think it would make more sense to just move the functionality from `updateUser()` into `UpdateUser()`, unless you had some reason to do it like this?

gliptak:
@Tenzer I refactored the code. Thanks

gliptak:
@jamesog The current pull request might fullfill @Tenzer 's use case better. Only the fields to be updated have to be modified (as contrasted to setting all other fields to ```nil```).

jamesog:
@gliptak with my suggestion you don't need to set all other fields to nil, they will implicitly be done so by not setting them. Setting the JSON tag to `omitempty` means they will not be sent to the API when encoded. This is better than duplicating structs.

gliptak:
@jamesog In the use case by @Tenzer above, the ```User``` object will already have values set (as it was returned by an earlier system call). So the callee would have to set all other fields to ```nil``` before ```UpdateUser``` call.

jamesog:
@gliptak this is not true. Read about how `omitempty` works. https://golang.org/pkg/encoding/json/#Marshal

gliptak:
@jamesog Thank you for the pointer. I updated ```User``` with ```omitempty``` and now my unittest fails. Do you see ```omitempty``` being used correctly?

gliptak:
Would you still like to move forward with this? Thanks

jamesog:
Hi @gliptak - apologies, I haven't had any time to review your changes until now :-( I'll reopen this, I think your changes look good.

jamesog:
@gliptak The changes in `user.go` seem fine, but the tests are failing. Can you check these, please? Thanks!

gliptak:
@jamesog Please review history above. The tests started failing, when I went ahead with the ```omitempty``` recommendation. Let me know if you would like to roll them back. Thanks

jamesog:
Ah, sorry, let me take another look.

Kellel:
@gliptak Looks like your test on line 77 is expecting the wrong data.

gliptak:
@Kellel Yes it does to match the API manual at https://api.cloudflare.com/#user-update-user

Kellel:
@gliptak I see what you are saying. 

This is what I think the options are:

1. Create a new User object in UpdateUser and copy over the fields that are allowed
2. Zero the other fields of the user struct before sending to the api
3. (worst option IMHO) After a little bit of testing against my personal zone I've validated that the Cloudflare api will except the whole user struct without error. It seems to just ignore the fields that are not excepted.

jamesog:
@Kellel Because `omitempty` is used in the JSON tag, when the field isn't defined in the struct (and therefore assumes the empty value for that type), the field doesn't get sent to the API when you marshal to JSON. This is desired for read-only properties like `ID` and `CreatedOn`.

jamesog:
So, inserting some debug statements into the test function, we get:

```
      mux.HandleFunc("/user", func(w http.ResponseWriter, r *http.Request) {
          assert.Equal(t, "PATCH", r.Method, "Expected method 'PATCH', got %s", r.Method)
          b, err := ioutil.ReadAll(r.Body)
          defer r.Body.Close()
         fmt.Println("LOG:", string(b))
         fmt.Println("LOG:", `{"country":"US","first_name":"John", "last_name": "Appleseed","telephone": "+1 123-123-1234","zipco  de": "12345"}`)
          if assert.NoError(t, err) {
              assert.JSONEq(t, `{"country":"US","first_name":"John",
              "last_name": "Appleseed","telephone": "+1 123-123-1234","zipcode": "12345"}`, string(b))
          }
```

```
=== RUN   TestUser_UpdateUser
LOG: {"id":"7c5dae5552338874e5053f2534d2767a","email":"user@example.com","first_name":"John","last_name":"Appleseed","username":"cfuser12345","telephone":"+1 123-123-1234","country":"US","zipcode":"12345","created_on":"2014-01-01T05:20:00Z","modified_on":"2014-01-01T05:20:00Z"}
LOG: {"country":"US","first_name":"John", "last_name": "Appleseed","telephone": "+1 123-123-1234","zipcode": "12345"}
```

So the test data in `assert.JSONeq()` does need to include the extra fields.

Kellel:
But then @jamesog we are sending more than https://api.cloudflare.com/#user-user-details says is explicitly allowing. As I mentioned, I tested the code as is and it worked correctly, so I'm happy with just having updated test data. But having the test explicitly match the api docs sounds like a win to me.

jamesog:
I think what we need to do to make this work correctly is turn the `CreatedOn` and `ModifiedOn` fields into `*time.Time` (as done in `zone.go` for Analytics), then the zero-value will be `nil` instead of `0001-01-01T00:00:00Z`.

Then, as Kellen points out, the test data needs to be updated. The following diff made the tests pass for me:

```
diff --git a/user.go b/user.go
index efe700f..54fb19f 100644
--- a/user.go
+++ b/user.go
@@ -17,8 +17,8 @@ type User struct {
        Telephone     string         `json:"telephone,omitempty"`
        Country       string         `json:"country,omitempty"`
        Zipcode       string         `json:"zipcode,omitempty"`
-       CreatedOn     time.Time      `json:"created_on,omitempty"`
-       ModifiedOn    time.Time      `json:"modified_on,omitempty"`
+       CreatedOn     *time.Time     `json:"created_on,omitempty"`
+       ModifiedOn    *time.Time     `json:"modified_on,omitempty"`
        APIKey        string         `json:"api_key,omitempty"`
        TwoFA         bool           `json:"two_factor_authentication_enabled,omitempty"`
        Betas         []string       `json:"betas,omitempty"`
diff --git a/user_test.go b/user_test.go
index 5ef97ac..7a729d5 100644
--- a/user_test.go
+++ b/user_test.go
@@ -6,8 +6,9 @@ import (
        "testing"
        "time"

-       "github.com/stretchr/testify/assert"
        "io/ioutil"
+
+       "github.com/stretchr/testify/assert"
 )

 func TestUser_UserDetails(t *testing.T) {
@@ -53,8 +54,8 @@ func TestUser_UserDetails(t *testing.T) {
                Telephone:  "+1 (650) 319 8930",
                Country:    "US",
                Zipcode:    "94107",
-               CreatedOn:  createdOn,
-               ModifiedOn: modifiedOn,
+               CreatedOn:  &createdOn,
+               ModifiedOn: &modifiedOn,
                TwoFA:      true,
                Betas:      []string{"mirage_forever"},
        }
@@ -73,8 +74,8 @@ func TestUser_UpdateUser(t *testing.T) {
                b, err := ioutil.ReadAll(r.Body)
                defer r.Body.Close()
                if assert.NoError(t, err) {
-                       assert.JSONEq(t, `{"country":"US","first_name":"John",
-                       "last_name": "Appleseed","telephone": "+1 123-123-1234","zipcode": "12345"}`, string(b))
+                       assert.JSONEq(t, `{"country":"US","first_name":"John","username":"cfuser12345","email":"user@example.com",
+                       "last_name": "Appleseed","telephone": "+1 123-123-1234","zipcode": "12345"}`, string(b), "JSON not equal")
                }
                fmt.Fprintf(w, `{
@@ -101,17 +102,14 @@ func TestUser_UpdateUser(t *testing.T) {
        modifiedOn, _ := time.Parse(time.RFC3339, "2014-01-01T05:20:00Z")

        userIn := User{
-               ID:         "7c5dae5552338874e5053f2534d2767a",
-               Email:      "user@example.com",
-               FirstName:  "John",
-               LastName:   "Appleseed",
-               Username:   "cfuser12345",
-               Telephone:  "+1 123-123-1234",
-               Country:    "US",
-               Zipcode:    "12345",
-               CreatedOn:  createdOn,
-               ModifiedOn: modifiedOn,
-               TwoFA:      false,
+               Email:     "user@example.com",
+               FirstName: "John",
+               LastName:  "Appleseed",
+               Username:  "cfuser12345",
+               Telephone: "+1 123-123-1234",
+               Country:   "US",
+               Zipcode:   "12345",
+               TwoFA:     false,
        }

        userOut, err := client.UpdateUser(&userIn)
@@ -125,12 +123,12 @@ func TestUser_UpdateUser(t *testing.T) {
                Telephone:  "+1 123-123-1234",
                Country:    "US",
                Zipcode:    "12345",
-               CreatedOn:  createdOn,
-               ModifiedOn: modifiedOn,
+               CreatedOn:  &createdOn,
+               ModifiedOn: &modifiedOn,
                TwoFA:      false,
        }

        if assert.NoError(t, err) {
-               assert.Equal(t, userOut, want)
+               assert.Equal(t, userOut, want, "structs not equal")
        }
 }
```

gliptak:
I'm not sure how this is expected to be updated ... Feel free modify and commit it as you all see best.

jamesog:
@gliptak That's up to you, I'm just trying to offer some help and advice with your feature so we can integrate it. I'm not intending to take the work away from you. :-)

I can't modify it as the PR comes from a branch in your fork, but it's probably best if you test my proposal too to see if you agree with it.

gliptak:
Applied diff from above. Thanks

jamesog:
Thanks @gliptak! And thanks for your patience.

gliptak:
@jamesog No worries :) Now we have a reviewed example to use go forward.

Issue #97, opened 2016-12-20
Problems updating DNS record


mwarkentin:
Hey there, I'm looking into https://github.com/hashicorp/terraform/issues/7690, and trying to determine if the issue is in the terraform code or in the Cloudflare library.

I'm having issues updating a record using `cloudflare-go` directly (I can create it ok). Wondering if someone can point out if I'm doing something incorrectly, or if this is actually a bug in the library.

I'm pretty new to golang, so not surprised if I'm doing something incorrectly here.

### cloudflare.go

```go
package main

import (
	"fmt"
	"log"
	"os"

	"github.com/cloudflare/cloudflare-go"
)

func main() {
	// Construct a new API object
	api, err := cloudflare.New(os.Getenv("CF_API_KEY"), os.Getenv("CF_API_EMAIL"))
	if err != nil {
		log.Fatal(err)
	}

	domain := os.Getenv("CF_DOMAIN")

	// Fetch the zone ID
	zoneId, err := api.ZoneIDByName(domain)
	if err != nil {
		log.Fatal(err)
	}

	newRecord := cloudflare.DNSRecord{
		Type:     "A",
		Name:     "CFAPI",
		Content:  "127.0.0.1",
		Proxied:  false,
		ZoneName: domain,
		ZoneID:   zoneId,
		TTL:      3600,
	}

	log.Printf("[DEBUG] CloudFlare Record create configuration: %#v", newRecord)

	r, err := api.CreateDNSRecord(zoneId, newRecord)
	if err != nil {
		fmt.Errorf("Failed to create record: %s", err)
	}

	recordId := r.Result.ID

	// In the Event that the API returns an empty DNS Record, we verify that the
	// ID returned is not the default ""
	if recordId == "" {
		fmt.Errorf("Failed to find record in Create response; Record was empty")
	}

	log.Printf("[INFO] CloudFlare Record ID: %s", recordId)

	updatedRecord := cloudflare.DNSRecord{
		ID:       recordId,
		Type:     "A",
		Name:     "CFAPI-modified",
		Content:  "192.168.10.10",
		Proxied:  true,
		ZoneName: domain,
		ZoneID:   zoneId,
		TTL:      120,
	}

	log.Printf("[DEBUG] CloudFlare Record update configuration: %#v", updatedRecord)

	err = api.UpdateDNSRecord(zoneId, recordId, updatedRecord)
	if err != nil {
		fmt.Errorf("Failed to update record: %s", err)
	}
}

```

[...]

PR #98, opened 2016-12-20
Bugfix for UpdateDNSRecord

jamesog:
Should address #97.


mwarkentin:
Looks like this fix worked: https://github.com/cloudflare/cloudflare-go/issues/97#issuecomment-268569747

PR #100, opened 2016-12-30
Add Codecov.io

gliptak:


gliptak:
Please review. Thanks

jamesog:
There's conflicts because of your other change to the Travis config.

@elithrar do you know what our policy is on adding services like this? I assume it's OK but not sure...

elithrar:
I'd prefer not to wire up a third party service at this stage without additional review. Our TravisCI build could instead be amended to output coverage on test runs - e.g. 

```diff
-   - go test -v -race $(go list ./... | grep -v '/vendor/')
+   - go test -v -cover -race $(go list ./... | grep -v '/vendor/')
```

gliptak:
What codecov does is reading the coverage.txt file generated by Travis and displays it as history/charts ...

elithrar:
Sure, but it needs read/write access to the repo to do it. That's the
problem.

gliptak:
https://github.com/codecov/codecov-bash/issues/52

gliptak:
@elithrar Would running upload using ```curl``` be acceptable? Thanks

elithrar:
That looks better. Two questions:

- what is the token in the URL?
- who owns the account?

gliptak:
I do not believe ```token``` is required for public repositories.
As for the account ownership question, my understanding is that Github userids with write access to the repository will have configuration access to https://codecov.io/gh/cloudflare/cloudflare-go

Issue #35, opened 2016-05-03
Implement Organization endpoints


Tenzer:
https://api.cloudflare.com/#user-s-organizations
https://api.cloudflare.com/#organizations
https://api.cloudflare.com/#organization-members
https://api.cloudflare.com/#organization-invites
https://api.cloudflare.com/#organization-roles

jamesog:
@gliptak I think you've (very kindly!) done all of these now, is that right?

gliptak:
@jamesog Yes, I implemented all the ones listed above.

jamesog:
Great! Thanks very much.

Issue #78, opened 2016-05-27
[breaking] Proposal: Make Response types private; expose Results only


elithrar:
Exposing `XXXResponse` types clutters the public API and offers minimal (if any) benefit to consumers of the package.

Discussion with @jamesog and @saljam internally leads to this proposal:
- [ ] `XXXResponse` types should be made private
- [ ] We should only expose `Response.Result` - which will be a `*DNSRecord` or `*PageRule`
- [ ] We unwrap the errors from `Response.Errors` and return them.

In addition we should look to standardize on returning/using pointer types, but that could also be tackled in a separate PR/issue.

jamesog:
Having just been toying with trying to get pagination working for DNS, I think we do need to keep the Response types public, or else return ResultInfo (with the page and count info) separately.

elithrar:
@jamesog Returning `ResultInfo` as the second of three params might be the best way.

jamesog:
So, something like this?

```
func (api *API) DNSRecords() (*DNSRecord[], ResultInfo, error)
```

elithrar:
Yep.

PR #101, opened 2017-01-08
Implement ListOrganizations

gliptak:
https://github.com/cloudflare/cloudflare-go/issues/35

@jamesog @elithrar Please review and guide on pagination.

One thought would be to rename ```ResultInfo``` to ```Paginator``` and add a second method signature taking in ```Paginator```. Or variadic args as per https://github.com/cloudflare/cloudflare-go/pull/4 ?

Thanks

### DIFF
--- a/organizations.go
+++ b/organizations.go
@@ -1,5 +1,11 @@
 package cloudflare
 
+import (
+	"encoding/json"
+
+	"github.com/pkg/errors"
+)
+
 // Organization represents a multi-user organization.
 type Organization struct {
 	ID          string   `json:"id,omitempty"`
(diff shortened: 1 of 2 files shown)


gliptak:
Could you point out the preferred approach? Thanks

jamesog:
Hi @gliptak - thanks for this!

Pagination we've discussed a bit in #6 and #78. It's going to be a breaking change to implement it so it's probably worth doing everything at once. I've got a WIP I'm testing at the moment but needs a bit more work before I submit it.

For now I think this PR is OK. I recently updated the `DNSRecords()` method (in #102) to do the pagination itself and return all records, until we've implemented pagination properly.

organizations.go
```diff
@@ -9,9 +15,28 @@ type Organization struct {
 	Roles       []string `json:"roles,omitempty"`
 }
 
-// OrganizationResponse represents the response from the Organization endpoint.
-type OrganizationResponse struct {
+// organizationResponse represents the response from the Organization endpoint.
+type organizationResponse struct {
 	Response
 	Result     []Organization `json:"result"`
 	ResultInfo ResultInfo     `json:"result_info"`
```
jamesog: Go lets you omit the second `ResultInfo` here but I'm OK either way having it or not.

gliptak: Removed ResultInfo

PR #102, opened 2017-01-08
DNSRecords: Return all records

jamesog:
Until pagination is implemented, the current implementation of
DNSRecords is lacklustre as it only returns a limited number of records
(20 by default).

This change updates DNSRecords to always return all records to the
caller, allowing it to use the data as it seems fit. It fetches the
maximum number of records per-page as allowed by the API to minimise the
API calls needed.

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -3,6 +3,7 @@ package cloudflare
 import (
 	"encoding/json"
 	"net/url"
+	"strconv"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


dns.go
```diff
@@ -75,21 +80,33 @@ func (api *API) DNSRecords(zoneID string, rr DNSRecord) ([]DNSRecord, error) {
...
+		if err != nil {
+			return []DNSRecord{}, errors.Wrap(err, errMakeRequestError)
+		}
+		var r DNSListResponse
+		err = json.Unmarshal(res, &r)
+		if err != nil {
+			return []DNSRecord{}, errors.Wrap(err, errUnmarshalError)
+		}
+		records = append(records, r.Result...)
+		if len(records) >= r.ResultInfo.Total {
```
terinjokes: The API returns a `total_pages` response (that's not included in `ResultInfo`), so this comparison can probably be changed to

```go
if page >= r.ResultInfo.TotalPages
```

jamesog: Thanks, I didn't see that field had been added. I'll add this to `ResultInfo` and use that instead. Much nicer.

PR #104, opened 2017-01-16
feat(auth): implement User Service authentication

jamesog:
This commit adds constants for choosing an authentication type
(key+email or user-service key) and a method for setting the
authentication type. It also updates `makeRequest` to set the
appropriate headers for authentication.

Some new endpoints require X-Auth-User-Service-Key instead of X-Auth-Key + X-Auth-Email.

This change makes no assumption that only one will be used at a time;
it's perfectly possible to tell the API to send both authentication
types (by using a bitmask), however no guarantee is made this this will
do what you expect.

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -12,15 +12,23 @@ import (
 )
 
 const apiURL = "https://api.cloudflare.com/client/v4"
+const (
+	// AuthKeyEmail specifies that we should authenticate with API key and email address
+	AuthKeyEmail = 1 << iota
+	// AuthUserService specifies that we should authenticate with a User-Service key
+	AuthUserService
+)
 
 // API holds the configuration for the current API client. A client should not
 // be modified concurrently.
 type API struct {
-	APIKey     string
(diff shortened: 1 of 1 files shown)


cloudflare.go
```diff
@@ -122,8 +136,13 @@ func (api *API) request(method, uri string, reqBody io.Reader) (*http.Response,
 
 	// Apply any user-defined headers first.
 	req.Header = cloneHeader(api.headers)
-	req.Header.Set("X-Auth-Key", api.APIKey)
-	req.Header.Set("X-Auth-Email", api.APIEmail)
+	if api.authType&AuthKeyEmail != 0 {
```
terinjokes: I would prefer the infix operators to be separated by a space.

jamesog: This is how gofmt does it.

terinjokes: Yeah, `gofmt` uses the spacing in statements like the above to show the precedence order. Oh well.

PR #112, opened 2017-02-02
Remove unnecessary reflect uses

FiloSottile:
TEST THIS, I made it from the GitHub interface. It should compile. If it compiles it should run.

### DIFF
--- a/cmd/flarectl/flarectl.go
+++ b/cmd/flarectl/flarectl.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"log"
 	"os"
-	"reflect"
 	"strings"
 
 	"github.com/cloudflare/cloudflare-go"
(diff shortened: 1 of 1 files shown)


cmd/flarectl/flarectl.go
```diff
@@ -280,7 +279,7 @@ func zoneRecords(c *cli.Context) {
 		case "MX":
 			r.Content = fmt.Sprintf("%d %s", r.Priority, r.Content)
 		case "SRV":
-			dp := reflect.ValueOf(r.Data).Interface().(map[string]interface{})
+			dp := r.Data.(map[string]interface{})
```
elithrar: Why not `dp, ok` here?

FiloSottile: I just maintained the previous semantic, but yes, better error handling than just a panic would be good.

jamesog: I freely admit that this entire thing is a huge hack and poorly-written.

jamesog:
Finally got around to testing this. LGTM. I agree this needs more error handling, we can fix that in another PR.

PR #115, opened 2017-02-12
Standardise doc comments, add examples, update zone plans

jamesog:


jamesog:
I've kind of made some assumptions about what a "good" comment format should look like. We've had a mix of styles in the code so far (mostly my fault from early code). I thought this format looked reasonable enough. It seemed helpful to keep the links to the API reference docs, but I'm happy to drop that if we don't think it's necessary.

dns.go
```diff
@@ -43,9 +43,8 @@ type DNSListResponse struct {
 }
 
 // CreateDNSRecord creates a DNS record for the zone identifier.
-// API reference:
-//   https://api.cloudflare.com/#dns-records-for-a-zone-create-dns-record
-//   POST /zones/:zone_identifier/dns_records
+//
+// API reference: https://api.cloudflare.com/#dns-records-for-a-zone-create-dns-record
```
elithrar: LGTM:
screen shot 2017-02-13 at 8 12 44 am

zone.go
```diff
@@ -88,17 +95,17 @@ type ZoneIDResponse struct {
...
 	Response
-	Result []ZonePlan `json:"result"`
+	Result []ZoneRatePlan `json:"result"`
 	ResultInfo
 }
 
-// ZonePlanResponse represents the response from the Plan Details endpoint.
-type ZonePlanResponse struct {
+// ZoneRatePlanResponse represents the response from the Plan Details endpoint.
+type ZoneRatePlanResponse struct {
```
elithrar: Self-note: Should be documented as an API breaking change in the next release.

dns.go
jamesog: Yep, I ran a local `godoc` server to test the examples worked correctly :-)

zone.go
jamesog: Want me to add a note in the commit message so it's more obvious in history?

elithrar:
Yes please.

zone.go
jamesog: Done.

Issue #12, opened 2016-04-01
Proposal: flarectl Revamp


elithrar:
**Summary**: Re-factor flarectl to make adding new sub-commands more straightforward.
- [ ] Move to https://github.com/urfave/cli (which is the new, canonical repo for codegangsta/cli) or use https://github.com/bobziuchkovski/writ - which I'm using for my own logshare-cli tool today, with GNU getopt style options.
- [ ] Break sub-commands into their own files
- [ ] Tests so that upstream changes to the library don't break the CLI in subtle ways

Open to suggestions/feedback.

jamesog:
It already had sub-commands using `codegangsta/cli` - but admittedly not everything is implemented yet. I think I added everything the library supported at the time.

I'm happy to switch to another CLI library though. flarectl definitely needs some work. I threw it together pretty quickly to work as a test app.

I've had some thoughts on tests too. I have a WIP for testing the library itself. Let's get together while I'm in town to talk about this a bit more.

PR #124, opened 2017-05-15
break sub-commands into their own files

ElvinEfendi:
rel https://github.com/cloudflare/cloudflare-go/issues/12

This moves DNS and zone sub-commands into their own files and the rest of the commands into `misc.go`.

The purpose is to make it cleaner to add a new sub-command.


jamesog:
Ah, beat me to it :-) I started reworking flarectl recently - I was thinking of switching from `cli` to `cobra` as it's a little nicer to use.

PR #125, opened 2017-05-15
application/json should be a default request content type

ElvinEfendi:
Turns out the library does not set `Content-Type` in request header. I was getting the following error for my `POST` request(`GET`s work as expected):
```
elvinefendi@Elvins-Pro flarectl (master)$ ./flarectl ch c --zone saastest1.com --hostname shop.elvinefendi.com
error from makeRequest: HTTP status 400: content "{\n  \"result\": {\n    \"ssl\": null\n  },\n  \"success\": false,\n  \"errors\": [\n    {\n      \"code\": 1001,\n      \"message\": \"error: unable to decode json request body. error='mime: no media type'\"\n    }\n  ]\n}\n"
```

The PR makes sure the `Content-Type` is set to `application/json` if it's empty - this fixes above issue.

According to the API docs it talks JSON only, why don't we always set `Content-Type` to `application/json`?

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -144,6 +144,10 @@ func (api *API) request(method, uri string, reqBody io.Reader) (*http.Response, 
 		req.Header.Set("X-Auth-User-Service-Key", api.APIUserServiceKey)
 	}
 
+	if req.Header.Get("Content-Type") == "" {
+		req.Header.Set("Content-Type", "application/json")
+	}
+
 	resp, err := api.httpClient.Do(req)
 	if err != nil {
 		return nil, errors.Wrap(err, "HTTP request failed")
--- a/cloudflare_test.go
+++ b/cloudflare_test.go
@@ -62,7 +62,18 @@ func TestClient_Headers(t *testing.T) {
 	teardown()
 
 	// it should set X-Auth-User-Service-Key and omit X-Auth-Email and X-Auth-Key when client.authType is AuthUserService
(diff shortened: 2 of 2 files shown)


ElvinEfendi:
> According to the API docs it talks JSON only, why don't we **always** set Content-Type to application/json?

@elithrar any reason to not do ^^?

handrews:
While rare, some APIs do return other content-types.  It's under HTTP Content-Type a bit further down in this section: https://api.cloudflare.com/#getting-started-responses

ElvinEfendi:
> While rare, some APIs do return other content-types. It's under HTTP Content-Type a bit further down in this section: https://api.cloudflare.com/#getting-started-responses

@handrews thanks for the link but I was asking about _request_ header. Is there any endpoint that requires non JSON request?

jamesog:
Ouch, thanks for catching and fixing this @ElvinEfendi. This rings bells back to my early implementations of the library. I think I once had a TODO or commented line to fix that, but it may have been removed in a code cleanup.

handrews:
@ElvinEfendi ah, that makes more sense.  I shouldn't respond to issues before my first cup of tea in the morning :-)

I believe there is at least one where you upload a file in a different format (either a DNS zone file or a configuration for something that is managed as YAML).  But we do not have alternate resource representations like (shudder) XML or anything.  The file uploads are essentially opaque data to the API.

PR #129, opened 2017-05-24
Change max per_page to 1000

HackAttack:
The documentation claims that the maximum number of records that can be requested in one result page is 50, but it appears to actually be 1000:

```
$ curl -H ... https://api.cloudflare.com/client/v4/user/billing/history?per_page=1000 | jq .
{
  "result": [],
  "result_info": {
    "page": 1,
    "per_page": 1000,
    "total_pages": 0,
    "count": 0,
    "total_count": 0
  },
  "success": true,
  "errors": [],
  "messages": []
}
$ curl -H ... https://api.cloudflare.com/client/v4/user/billing/history?per_page=1001 | jq .
{
  "success": false,
  "errors": [
    {
      "code": 6008,
      "message": "Invalid request parameters",
      "error_chain": [
        {
          "code": 11000,
          "message": "per_page variable must be a number between 1 and 1000"
        }
      ]
    }
  ],
  "messages": [],
  "result": null
}
```

### DIFF
--- a/custom_hostname.go
+++ b/custom_hostname.go
@@ -94,7 +94,7 @@ func (api *API) CreateCustomHostname(zoneID string, ch CustomHostname) (*CustomH
 // API reference: https://api.cloudflare.com/#custom-hostname-for-a-zone-list-custom-hostnames
 func (api *API) CustomHostnames(zoneID string, page int, filter CustomHostname) ([]CustomHostname, ResultInfo, error) {
 	v := url.Values{}
-	v.Set("per_page", "50")
+	v.Set("per_page", "1000")
 	v.Set("page", strconv.Itoa(page))
 	if filter.Hostname != "" {
 		v.Set("hostname", filter.Hostname)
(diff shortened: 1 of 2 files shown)


Tenzer:
The max number of results per request is different from endpoint to endpoint. In the case of the two parts of the code you have suggested changes to, the documentation actually specify [50](https://api.cloudflare.com/#custom-hostname-for-a-zone-list-custom-hostnames) and [100](https://api.cloudflare.com/#dns-records-for-a-zone-list-dns-records) as the maximum number for `per_page`. None of those values however have any relation to the billing history which is what you show in the example.

Although the documentation may specify one limit and the actual limit may be different, we should ideally stick with the limit specified in the documentation as we otherwise may get to a place where changes happen on the backend, so it gets closer to the documented max limit, which then will break the library.

HackAttack:
I used billing history in the example only to avoid pasting a zone id. I have observed the same 1000 limit on other endpoints, including `dns_records`, so I assumed 1000 was the actual max across all endpoints.

If the discrepancy between the documentation and the actual behavior is intentional, then I agree with following the documentation.

PR #136, opened 2017-06-14
[feature] Complete Origin CA Support from #61

cbranch:
Salvatore's work in #61 got us almost all the way there. I've addressed the outstanding issues and fixed a type error with the current API.

> Do you want to update the methods to call api.SetAuthType(AuthUserService)? I'm also thinking we should fetch the current value before doing that and resetting it when we're done.

I think it's better if makeRequest takes an AuthType. Setting it temporarily when calling the Origin CA functions is racy, IMO I expect to be able to use the API client from multiple goroutines because it isn't inherently stateful.


iToto:
Thanks for seeing this through @cbranch

Issue #108, opened 2017-01-20
Cannot unmarshal string into int when creating PageRule


OJFord:
I'm seeing a similar issue to #52 but with PageRules, in which they are actually created and I can retrieve fine with curl et al. - but an error is returned:

> error unmarshalling the JSON response: json: cannot unmarshal string into Go value of type int

jamesog:
Thanks, I've reported an internal bug to have this addressed.

jamesog:
The fix was deployed yesterday. Thanks for reporting this :-)

OJFord:
Fantastic, thanks! Should be able to get page rules working in Terraform now :slightly_smiling_face:

OJFord:
This seems to have regressed?

> error unmarshalling the JSON response: json: cannot unmarshal array into Go struct field PageRuleDetailResponse.result of type cloudflare.PageRule

As before, it's particularly annoying because it occurs even on operations that succeed.  c.f. terraform-providers/terraform-provider-cloudflare#12.

elithrar:
@OJFord Looking into this.

elithrar:
OK, I can't reproduce on the current HEAD (e6739d6) - no issues unmarshalling the response from the API:

```go
package main

import (
	"fmt"
	"log"
	"os"

	"github.com/cloudflare/cloudflare-go"
)

func main() {
	client, err := cloudflare.New(
		os.Getenv("CF_API_KEY"),
		os.Getenv("CF_API_EMAIL"),
	)
	if err != nil {
		log.Fatal(err)
	}

	zoneID, err := client.ZoneIDByName(os.Getenv("CF_ZONE_NAME"))
	if err != nil {
		log.Fatal(err)
	}

	rules, err := client.ListPageRules(zoneID)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Fprintf(os.Stdout, "%#v\n", rules)

	rule := cloudflare.PageRule{
		Targets: []cloudflare.PageRuleTarget{cloudflare.PageRuleTarget{
			Target: "url",
			Constraint: struct {
				Operator string `json:"operator"`
				Value    string `json:"value"`
			}{
				Operator: "matches",
				Value:    fmt.Sprintf("%s/*?test=yes", os.Getenv("CF_ZONE_NAME")),
			},
		},
		},
		Actions: []cloudflare.PageRuleAction{
			cloudflare.PageRuleAction{
				ID:    "ip_geolocation",
				Value: "on",
			},
[...]

PR #138, opened 2017-07-08
Make CRUD API + pagination treatment consistent across resources [WIP]

OJFord:
This PR is a (breaking) change to the `CreatePageRule` signature,
making it consistent with other `Create~` functions:
```go
func (api *API) Create~(zoneID string, ~) (*~Response, error)
```

Aside from motivations of consistency, without this API the user is
unable to discover the server-created ID for the page rule without
enumerating `ListPageRule` before and after creation - and assuming they
are the only user at that instant.

C.f. discussion in #108; ping @elithrar

Example use: (Should I add a test? I can't seem to make existing `*_test.go` files fail, so I'm not clear if they're running when I do `go test`, or how to add one and be sure it's running.)
```go
package main

import (
    "fmt"
    "os"
    cloudflare "github.com/cloudflare/cloudflare-go"
)

func main() {
    domain := os.Getenv("CLOUDFLARE_DOMAIN")
    api, err := cloudflare.New(os.Getenv("CLOUDFLARE_TOKEN"), os.Getenv("CLOUDFLARE_EMAIL"))
    if err != nil {
        fmt.Print(err)
        return
    }

    rule := cloudflare.PageRule{
        Targets: []cloudflare.PageRuleTarget{
            cloudflare.PageRuleTarget{
                Target: "url",
                Constraint: struct {
                     Operator string `json:"operator"`
                     Value    string `json:"value"`
                }{
                     Operator: "matches",
                     Value:    fmt.Sprintf("test.%s", domain),
                },
            },
        },
        Actions: []cloudflare.PageRuleAction{
            cloudflare.PageRuleAction{
                ID:    "always_online",
                Value: "on",
            },
        },
        Priority: 1,
        Status:   "active",
    }

    zoneID, err := api.ZoneIDByName(domain)
    if err != nil {
        fmt.Printf("Error finding zone %q: %s", domain, err)
        return
    }

    r, err := api.CreatePageRule(zoneID, rule)
    if err != nil {
        fmt.Printf("Failed to create page rule: %s", err)
        return nil
    }

    fmt.Printf("Result: %q", r.Result)
}
```

### DIFF
--- a/pagerules.go
+++ b/pagerules.go
@@ -104,18 +104,20 @@ type PageRulesResponse struct {
 // CreatePageRule creates a new Page Rule for a zone.
 //
 // API reference: https://api.cloudflare.com/#page-rules-for-a-zone-create-a-page-rule
-func (api *API) CreatePageRule(zoneID string, rule PageRule) error {
+func (api *API) CreatePageRule(zoneID string, rule PageRule) (*PageRuleDetailResponse, error) {
 	uri := "/zones/" + zoneID + "/pagerules"
 	res, err := api.makeRequest("POST", uri, rule)
 	if err != nil {
-		return errors.Wrap(err, errMakeRequestError)
+		return nil, errors.Wrap(err, errMakeRequestError)
 	}
-	var r PageRuleDetailResponse
-	err = json.Unmarshal(res, &r)
+
+	var pageRuleResp *PageRuleDetailResponse
+	err = json.Unmarshal(res, &pageRuleResp)
 	if err != nil {
-		return errors.Wrap(err, errUnmarshalError)
+		return nil, errors.Wrap(err, errUnmarshalError)
 	}
-	return nil
+
+	return pageRuleResp, nil
 }
 
 // ListPageRules returns all Page Rules for a zone.
--- a/pagerules.go
+++ b/pagerules.go
@@ -2,6 +2,8 @@ package cloudflare
 
 import (
 	"encoding/json"
+	"net/url"
+	"strconv"
 	"time"
 
 	"github.com/pkg/errors"
@@ -93,8 +95,8 @@ type PageRuleDetailResponse struct {
 	Result   PageRule `json:"result"`
 }
 
-// PageRulesResponse is the API response, containing an array of PageRules.
-type PageRulesResponse struct {
+// PageRuleListResponse is the API response, containing an array of PageRules.
+type PageRuleListResponse struct {
 	Success  bool       `json:"success"`
 	Errors   []string   `json:"errors"`
 	Messages []string   `json:"messages"`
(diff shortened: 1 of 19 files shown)


OJFord:
I initially wrote that this makes it consistent with _all_ other `Create~` functions. That's not true, looking more closely I realised `CreateKeyless` returns nothing at all, and others return non-pointers, or the `~Response.Result` instead of the whole `~Response`.

I wouldn't like to presume to do that, but it might be worth considering making them consistent even if in some cases there's no other reason to want the response/result (as there is here).

For reference:
```sh
$ rg 'func \(api \*API\) Create'
custom_hostname.go
73:func (api *API) CreateCustomHostname(zoneID string, ch CustomHostname) (*CustomHostnameResponse, error) {
railgun.go
51:func (api *API) CreateRailgun(name string) (Railgun, error) {
dns.go
48:func (api *API) CreateDNSRecord(zoneID string, rr DNSRecord) (*DNSRecordResponse, error) {
virtualdns.go
36:func (api *API) CreateVirtualDNS(v *VirtualDNS) (*VirtualDNS, error) {
pagerules.go
107:func (api *API) CreatePageRule(zoneID string, rule PageRule) (*PageRuleDetailResponse, error) {
keyless.go
27:func (api *API) CreateKeyless() {
origin_ca.go
58:func (api *API) CreateOriginCertificate(certificate OriginCACertificate) (*OriginCACertificate, error) {
zone.go
244:func (api *API) CreateZone(name string, jumpstart bool, org Organization) (Zone, error) {
ssl.go
56:func (api *API) CreateSSL(zoneID string, options ZoneCustomSSLOptions) (ZoneCustomSSL, error)  {
```

elithrar:
This is definitely something we need to resolve. We need to:

* Make it easy to access the data directly
* Paginate as needed - the library shouldn't traverse pages automatically (tens - hundreds!), but should make it really easy to fetch the "next" page if you so wish
* Use the full response object to raise errors (e.g. success: false)

pagerules.go
```diff
@@ -104,18 +104,20 @@ type PageRulesResponse struct {
...
 	res, err := api.makeRequest("POST", uri, rule)
 	if err != nil {
-		return errors.Wrap(err, errMakeRequestError)
+		return nil, errors.Wrap(err, errMakeRequestError)
 	}
-	var r PageRuleDetailResponse
-	err = json.Unmarshal(res, &r)
+
+	var pageRuleResp *PageRuleDetailResponse
+	err = json.Unmarshal(res, &pageRuleResp)
```
nkcmr: @OJFord Is this taking the address of a pointer? Could probably just pass in `pageRuleResp` without `&` since `pageRuleResp` is already a pointer, no?

OJFord:
@nkcmr I thought that was weird, but I don't know Go so I just copied from
DNS records, iirc. Can certainly fix :)

pagerules.go
OJFord: I thought that was weird, but I don't know Go so I just copied from [dns.go](https://github.com/cloudflare/cloudflare-go/blob/master/dns.go#L55-L56) - I'll fix it though if it is alright without. :slightly_smiling_face:

OJFord:
@elithrar Good point about pagination - I don't have enough of anything to test, but as far as I can tell the library currently returns the first page of default size, and gives no way* to access further pages? Which in a way is worse even than traversing them all automatically as you describe!

(*though, for methods that do return the whole response, not just the result as described above, a user could make requests for further pages using the (web) API directly, outside of the library...)

Do you want to do that within this PR though, as a sort of one-hit (go) API change?

elithrar:
@OJFord 

* One-hit change, yes please.
* I've pushed some new methods for new API endpoints and have exposed a `page` argument as well.

e.g.

https://godoc.org/github.com/cloudflare/cloudflare-go#API.ListZoneLockdowns
```go
func (api *API) ListZoneLockdowns(zoneID string, page int) (*ZoneLockdownListResponse, error)
```

https://godoc.org/github.com/cloudflare/cloudflare-go#API.CreateZoneLockdown
```go
func (api *API) CreateZoneLockdown(zoneID string, ld ZoneLockdown) (*ZoneLockdownResponse, error)
```

I think this is a sensible way forward, and the rest of the API should move to this.

cc/ @jamesog @Tenzer for feedback as well.

elithrar:
@OJFord - when you wrap up this PR, send me an email - msilverlock at cloudflare 🙃

pagerules.go
OJFord: Fixed in 9a98a8113f3ad245376eaece7547e2a40b7d8b47

elithrar:
Hey @OJFord - let me know if you're still keen on running with this

OJFord:
@elithrar I am! I'm so sorry for the delays, it's been my last comparatively free Summer, and in all honesty I've just had other priorities.

I hope to get back to it next week, but if you want to take it over I won't mind at all - at the end of the day my motivation is just to use it! :slightly_smiling_face:

elithrar:
@OJFord - no problems! I'm happy to wait.

Also: If you are also interested in doing more of this [our Go lib + Terraform provider] work, drop me an email: msilverlock @ cloudflare — would be great to see if there's a more formal way to help us out here.

SublimeDamien:
Hi,

Can I give some help to merge this PR ?

elithrar:
@SublimeDamien yes, help is more than welcome. Please ask if you have any questions.

SublimeDamien:
@elithrar  Ok, so if I want to help on this PR I have to fork on this repo OJFord:pagerule-consistent-api ? and open a new PR ? Or there is a way to push to this one ?

OJFord:
@sublimedamien

I believe to add commits here you'd have to open a PR against my forked repo, which I'd happily accept of course, it just adds unnecessary delay.. I'll see if I can add you as a contributor so you can push directly, otherwise just fork it again as you say.

elithrar:
Fork again, new PR is simplest. You can then pull Ollie’s changes in to
your own branch by adding his fork as another remote.

SublimeDamien:
@elithrar https://github.com/cloudflare/cloudflare-go/pull/154

PR #139, opened 2017-07-14
Allow DNS updates to change record type

its-johnt:


elithrar:
Could you also add a test for this?

jamesog:
Can you also update `flarectl` to allow setting the type there please?

elithrar:
Hey @its-johnt - do you think you can update flarectl to handle this too?

PR #144, opened 2017-09-22
[feat] Firewall access rules

CtrlZvi:
* [feat] Firewall access rules
* [docs] Add examples for zone firewall access rules

### DIFF
--- a/cmd/flarectl/firewall.go
+++ b/cmd/flarectl/firewall.go
@@ -1,0 +1,323 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"net"
+	"strconv"
+
+	"github.com/cloudflare/cloudflare-go"
+	"github.com/codegangsta/cli"
+)
+
+func firewallAccessRules(c *cli.Context) {
+	if err := checkEnv(); err != nil {
+		fmt.Println(err)
+		return
+	}
+
+	organizationID, zoneID, err := getScope(c)
+	if err != nil {
+		return
+	}
+
+	// Create an empty access rule for searching for rules
+	rule := cloudflare.AccessRule{
+		Configuration: getConfiguration(c),
+	}
+	if c.String("scope-type") != "" {
+		rule.Scope.Type = c.String("scope-type")
+	}
+	if c.String("notes") != "" {
+		rule.Notes = c.String("notes")
+	}
+	if c.String("mode") != "" {
+		rule.Mode = c.String("mode")
+	}
+
+	var response *cloudflare.AccessRuleListResponse
+	switch {
+	case organizationID != "":
+		response, err = api.ListOrganizationAccessRules(organizationID, rule, 1)
+	case zoneID != "":
+		response, err = api.ListZoneAccessRules(zoneID, rule, 1)
+	default:
+		response, err = api.ListUserAccessRules(rule, 1)
+	}
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+	totalPages := response.ResultInfo.TotalPages
+	rules := make([]cloudflare.AccessRule, 0, response.ResultInfo.Total)
+	rules = append(rules, response.Result...)
+	if totalPages > 1 {
+		for page := 2; page < totalPages; page++ {
+			switch {
+			case organizationID != "":
+				response, err = api.ListOrganizationAccessRules(organizationID, rule, page)
+			case zoneID != "":
+				response, err = api.ListZoneAccessRules(zoneID, rule, page)
+			default:
+				response, err = api.ListUserAccessRules(rule, page)
+			}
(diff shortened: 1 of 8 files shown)


firewall.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+	if accessRule.Configuration.Target != "" {
+		v.Set("configuration_target", accessRule.Configuration.Target)
+	}
+
+	var query string
+	var rules []AccessRule
+	page := 1
+
+	// Loop over makeRequest until what we've fetched all rules
```
elithrar: It may make more sense to take a `page int` arg to `ListAccessRules` instead, and allow the library consumer to call `client.ListZoneAccessRules` on their own if they need to paginate. This is what I did for ListZoneLockdowns: https://godoc.org/github.com/cloudflare/cloudflare-go#API.ListZoneLockdowns

This is probably esp. useful when you have ~10,000 rules and are fetching 100 at a time; as a consumer you can also create a `sync.WaitGroup` or worker pool and fire off the fetches concurrently to speed things up.

CtrlZvi: Easy enough to do. I will make that adjustment. I also just noticed that I didn't update flarectl, so I'll do that, too.

elithrar: 🙌🏼

firewall.go
```diff
@@ -0,0 +1,246 @@
...
+	"github.com/pkg/errors"
+)
+
+// AccessRule represents a firewall access rule.
+type AccessRule struct {
+	ID            string                  `json:"id,omitempty"`
+	Notes         string                  `json:"notes,omitempty"`
+	AllowedModes  []string                `json:"allowed_modes,omitempty"`
+	Mode          string                  `json:"mode,omitempty"`
+	Configuration AccessRuleConfiguration `json:"configuration,omitempty"`
```
SamWhited: Please run gofmt on this file.

firewall_example_test.go
```diff
@@ -0,0 +1,106 @@
+package cloudflare_test
+
+import (
+	"fmt"
+	"log"
+
+	// cloudflare "github.com/cloudflare/cloudflare-go"
```
SamWhited: Commented out import can be removed.

cmd/flarectl/firewall.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+	if c.String("notes") != "" {
+		rule.Notes = c.String("notes")
+	}
+	if c.String("mode") != "" {
+		rule.Mode = c.String("mode")
+	}
+
+	var rules []cloudflare.AccessRule
+	if organizationID != "" {
```
SamWhited: Arguably a bare `switch` statement is more idiomatic here. It also appears that a large chunk of logic is duplicated between all three cases, this could (probably?) be deduped.

cmd/flarectl/firewall.go
```diff
@@ -0,0 +1,380 @@
...
+				response, err := api.ListUserAccessRules(rule, page)
+				if err != nil {
+					fmt.Println(err)
+					return
+				}
+				rules = append(rules, response.Result...)
+			}
+		}
+	}
+	var output []table
```
SamWhited: `output := make([]table, 0, len(rules))` may avoid several expensive heap allocations and copies (although I have no idea if that's likely to matter or how long the output generally is).

firewall.go
CtrlZvi: Is there something you're objecting to on this file? When I run gofmt I see no changes.

cmd/flarectl/firewall.go
CtrlZvi: I haven't found a good way to deduplicate the logic, because the calls to key functions vary both in name and number of arguments. I'm open to suggestions, though. I can use a bare switch no problem. I was trying to be consistent with the use of if/else if/else over bare switch from zone.go

cmd/flarectl/firewall.go
CtrlZvi: Will fix.

firewall.go
elithrar: @SamWhited What diff are you seeing here? This looks fine to me; gofmt uses spaces to align struct fields/tags. Travis isn't reporting an error.

cmd/flarectl/firewall.go
SamWhited: Sounds good; consistency is probably more important and I'm not familiar with this code base. Thanks!

firewall.go
SamWhited: My apologies, the fields of this struct don't appear to be vertically aligned but maybe GitHub just isn't using a fixed width font. When I open it locally everything looks fine.

elithrar:
Thanks again @CtrlZvi!

PR #154, opened 2017-11-14
Make CRUD API + pagination treatment consistent across resources (2)

SublimeDamien:
https://github.com/cloudflare/cloudflare-go/pull/138

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -66,16 +66,17 @@ func (api *API) SetAuthType(authType int) {
 
 // ZoneIDByName retrieves a zone's ID from the name.
 func (api *API) ZoneIDByName(zoneName string) (string, error) {
-	res, err := api.ListZones(zoneName)
+	//TODO: paginate
+	response, err := api.FilterZones(1, zoneName)
 	if err != nil {
 		return "", errors.Wrap(err, "ListZones command failed")
 	}
-	for _, zone := range res {
-		if zone.Name == zoneName {
-			return zone.ID, nil
-		}
+
+	if len(response.Result) == 0 {
+		return "", errors.New("Zone could not be found")
 	}
-	return "", errors.New("Zone could not be found")
+
+	return response.Result[0].ID, nil
 }
 
 // makeRequest makes a HTTP request and returns the body as a byte slice,
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -67,7 +67,6 @@ func (api *API) SetAuthType(authType int) {
 
 // ZoneIDByName retrieves a zone's ID from the name.
 func (api *API) ZoneIDByName(zoneName string) (string, error) {
-	//TODO: paginate
 	response, err := api.FilterZones(1, zoneName)
 	if err != nil {
 		return "", errors.Wrap(err, "ListZones command failed")
--- a/cmd/flarectl/dns.go
+++ b/cmd/flarectl/dns.go
@@ -90,18 +90,18 @@ func dnsCreateOrUpdate(c *cli.Context) {
 	rr := cloudflare.DNSRecord{
 		Name: name + "." + zone,
 	}
-	records, err := api.DNSRecords(zoneID, rr)
+	records, err := api.FilterDNSRecords(zoneID, 1, rr)
 	if err != nil {
 		fmt.Fprintln(os.Stderr, "Error fetching DNS records: ", err)
 		return
 	}
 
 	var resp *cloudflare.DNSRecordResponse
-	if len(records) > 0 {
+	if len(records.Result) > 0 {
 		// Record exists - find the ID and update it.
 		// This is imprecise without knowing the original content; if a label
 		// has multiple RRs we'll just update the first one.
-		for _, r := range records {
+		for _, r := range records.Result {
 			if r.Type == rtype {
 				rr.ID = r.ID
 				rr.Type = r.Type
--- a/cmd/flarectl/firewall.go
+++ b/cmd/flarectl/firewall.go
@@ -320,12 +320,12 @@ func getScope(c *cli.Context) (string, string, error) {
 	var organization, organizationID string
 	if c.String("organization") != "" {
 		organization = c.String("organization")
-		organizations, _, err := api.ListOrganizations()
+		organizations, err := api.ListOrganizations(1)
 		if err != nil {
 			fmt.Println(err)
 			return "", "", err
 		}
-		for _, org := range organizations {
+		for _, org := range organizations.Result {
 			if org.Name == organization {
 				organizationID = org.ID
 				break
--- a/cmd/flarectl/misc.go
+++ b/cmd/flarectl/misc.go
@@ -61,14 +61,14 @@ func pageRules(c *cli.Context) {
 		return
 	}
 
-	rules, err := api.ListPageRules(zoneID)
+	rules, err := api.ListPageRules(zoneID, 1)
 	if err != nil {
 		fmt.Println(err)
 		return
 	}
 
 	fmt.Printf("%3s %-32s %-8s %s\n", "Pri", "ID", "Status", "URL")
-	for _, r := range rules {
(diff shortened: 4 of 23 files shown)


SublimeDamien:
@elithrar Can you trigger the CI please ?

elithrar:
Travis failed as it seems to have detected 'abuse' (it doesn't say what triggered that determination).

However, there's a lot of changes in here, and the commit messages don't line up. Can you rebase these commits to provide clearer commit messages?

SublimeDamien:
Ok I'll do that next week

elithrar:
@SublimeDamien Let me know when you're ready for me to re-review.

SublimeDamien:
Hi @elithrar I was little busy.
I split the commit in multiple commits

SublimeDamien:
@elithrar I have an error on the following command `diff -u <(echo -n) <(gofmt -d .)` but I don't understand the goal of this command and why there is a failure

mwarkentin:
@SublimeDamien Not 100%, but I think that's a check to make sure that code has been run through `gofmt` before check in? Might just need to run that first?

elithrar:
Correct, it’s gofmt.

elithrar:
Thanks @SublimeDamien - going to review later today!

SublimeDamien:
Ok, you're welcome.
I'm waiting your reviews

elithrar:
For the other reviewers, this PR *should* - 

* Allow a package user to paginate appropriately, depending on whether they find their results vs. network calls (and thus, time). Users with thousands of records would not want auto-pagination.
* Make method signatures as consistent as possible (pointer vs. value return, pointer vs. value args, etc) - this may require some bikeshedding (proposal: value for arg to avoid mutation, pointer for response to minimize copying large responses around)
* Return responses for all methods to allow the caller to retrieve IDs as necessary. The PageRules methods were the largest culprit here.

It will break the public API, and thus we'll bump the major version after merging.

cloudflare.go
```diff
@@ -67,16 +67,17 @@ func (api *API) SetAuthType(authType int) {
 
 // ZoneIDByName retrieves a zone's ID from the name.
 func (api *API) ZoneIDByName(zoneName string) (string, error) {
-	res, err := api.ListZones(zoneName)
+	//TODO: paginate
```
elithrar: Remove this comment, as not necessary. The FilterZones call should search for the exact name via the `name=${name}` param.

cmd/flarectl/dns.go
```diff
@@ -90,18 +90,18 @@ func dnsCreateOrUpdate(c *cli.Context) {
 	rr := cloudflare.DNSRecord{
 		Name: name + "." + zone,
 	}
-	records, err := api.DNSRecords(zoneID, rr)
+	records, err := api.FilterDNSRecords(zoneID, 0, rr)
```
elithrar: Why do we set page=0 here? The underlying call seems to coerce 0 to 1, but for clarity we should call page=1, which is what the HTTP API accepts.

cmd/flarectl/firewall.go
```diff
@@ -320,12 +320,12 @@ func getScope(c *cli.Context) (string, string, error) {
 	var organization, organizationID string
 	if c.String("organization") != "" {
 		organization = c.String("organization")
-		organizations, _, err := api.ListOrganizations()
+		organizations, err := api.ListOrganizations(0)
```
elithrar: Same here. Why not `ListOrganizations(1)`

cmd/flarectl/misc.go
```diff
@@ -61,14 +61,14 @@ func pageRules(c *cli.Context) {
 		return
 	}
 
-	rules, err := api.ListPageRules(zoneID)
+	rules, err := api.ListPageRules(zoneID, 0)
```
elithrar: `page=1`

cmd/flarectl/zone.go
```diff
@@ -64,13 +64,13 @@ func zoneList(c *cli.Context) {
 		fmt.Println(err)
 		return
 	}
-	zones, err := api.ListZones()
+	zones, err := api.ListZones(0)
```
elithrar: `page=1`

SublimeDamien:
@elithrar There is other things to change ?

elithrar:
No sorry! Busy week here. Will make an effort to review and hopefully
merge soon.

(It’s a large PR with multiple breaking changes and thus requires more
eyeballs)

zone_example_test.go
```diff
@@ -31,7 +31,7 @@ func ExampleAPI_ListZones_filter() {
 	}
 
 	// Fetch a slice of zones example.org and example.net.
-	zones, err := api.ListZones(1, "example.org", "example.net")
```
cbranch: This commit should be removed:

* you fix the call in a later commit
* your change no longer matches the comment

SublimeDamien:
@cbranch I tried to clean up, say me if it's OK

SublimeDamien:
@cbranch @elithrar Hi guys, what is missing for merge this PR ?

benjvi:
I've been reading through this and I'm not sure what the desired benefit is for returning response objects from Create methods. It seems simpler to pass just the object and error back, if that is possible. Is it the case that some resources don't have an ID and so passing the resource object back is insufficient to do lookups? (all the resource types I looked at so far do have IDs). I might be missing something else? @elithrar ?

elithrar:
@benjvi - I'm OK with your suggested approach here. The approach in #169 LGTM.

SublimeDamien:
@elithrar @benjvi If I understand what you want, for an example

actually there is:
```go
func (api *API) CreateCustomHostname(zoneID string, ch CustomHostname) (*CustomHostnameResponse, error) {
	uri := "/zones/" + zoneID + "/custom_hostnames"
	res, err := api.makeRequest("POST", uri, ch)
	if err != nil {
		return nil, errors.Wrap(err, errMakeRequestError)
	}

	var response CustomHostnameResponse
	err = json.Unmarshal(res, &response)
	if err != nil {
		return nil, errors.Wrap(err, errUnmarshalError)
	}

	return &response, nil
}
```

to

```go
func (api *API) CreateCustomHostname(zoneID string, ch CustomHostname) (*CustomHostname, error) {
	uri := "/zones/" + zoneID + "/custom_hostnames"
	res, err := api.makeRequest("POST", uri, ch)
	if err != nil {
		return nil, errors.Wrap(err, errMakeRequestError)
	}

	var response CustomHostnameResponse
	err = json.Unmarshal(res, &response)
	if err != nil {
		return nil, errors.Wrap(err, errUnmarshalError)
	}

	return &response.Result, nil
}
```

SublimeDamien:
@elithrar @benjvi If I understand what you want, for an example

actually there is:
```go
func (api *API) CreateCustomHostname(zoneID string, ch CustomHostname) (*CustomHostnameResponse, error) {
	uri := "/zones/" + zoneID + "/custom_hostnames"
	res, err := api.makeRequest("POST", uri, ch)
	if err != nil {
		return nil, errors.Wrap(err, errMakeRequestError)
	}

	var response CustomHostnameResponse
	err = json.Unmarshal(res, &response)
	if err != nil {
		return nil, errors.Wrap(err, errUnmarshalError)
	}

	return &response, nil
}
```

to

```go
func (api *API) CreateCustomHostname(zoneID string, ch CustomHostname) (*CustomHostname, error) {
	uri := "/zones/" + zoneID + "/custom_hostnames"
	res, err := api.makeRequest("POST", uri, ch)
	if err != nil {
		return nil, errors.Wrap(err, errMakeRequestError)
	}

	var response CustomHostnameResponse
	err = json.Unmarshal(res, &response)
	if err != nil {
		return nil, errors.Wrap(err, errUnmarshalError)
	}

	return &response.Result, nil
}
```

benjvi:
@SublimeDamien yes thats how I see it, passing just the pointer to the object and an error

PR #161, opened 2018-02-13
Rate limiting (for a zone)

benjvi:
Adds support for the rate limiting API. Mostly standard CRUD calls to API with marshalling etc. For listing, I allowed the caller to specify pagination options or call ListAll directly

### DIFF
--- a/rate_limiting.go
+++ b/rate_limiting.go
@@ -2,16 +2,17 @@ package cloudflare
 
 import (
 	"encoding/json"
-	"github.com/pkg/errors"
 	"net/url"
 	"strconv"
+
+	"github.com/pkg/errors"
 )
 
 // RateLimit is a policy than can be applied to limit traffic within a customer domain
 type RateLimit struct {
 	ID          string                  `json:"id,omitempty"`
-	Disabled    bool                    `json:"disabled,omitempty"` // api defaults to false
-	Description string                  `json:"description"`
+	Disabled    bool                    `json:"disabled,omitempty"`
+	Description string                  `json:"description,omitempty"`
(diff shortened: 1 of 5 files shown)


rate_limiting.go
```diff
@@ -0,0 +1,194 @@
+package cloudflare
+
+import (
+	"encoding/json"
+	"github.com/pkg/errors"
```
jamesog: Nit: the style elsewhere in this project (presumably because most people are using `goimports`) is to put external imports at the end, after a blank line.

rate_limiting_example_test.go
```diff
@@ -0,0 +1,109 @@
+package cloudflare_test
+
+import (
+	"fmt"
+	cloudflare "github.com/cloudflare/cloudflare-go"
```
jamesog: Same here.

rate_limiting.go
```diff
@@ -0,0 +1,194 @@
...
+	"net/url"
+	"strconv"
+)
+
+// RateLimit is a policy than can be applied to limit traffic within a customer domain
+type RateLimit struct {
+	ID          string                  `json:"id,omitempty"`
+	Disabled    bool                    `json:"disabled,omitempty"` // api defaults to false
```
jamesog: Not sure if the comment is necessary here as bool also defaults to false

benjvi:
Thanks guys. I addressed the comments from @jamesog, also I realised today that the empty struct values were different than the api defaults when completely omitting a field in a couple of places -
 which results in the wrong thing happening when empty values are given. I replaced these with pointers so the values get omitted completely and default values are used. Hopefully this should be ok now?

PR #162, opened 2018-02-15
Added CheckRegions field to LoadBalancerPool

FnuGk:
This adds the `CheckRegions` field to `LoadBalancerPool`.
This makes it possible to set the check regions via the API.
Without this modifying a LoadBalancerPool would set the check regions to `null` meaning all regions, resulting in the error `Probing from every data centre is not allowed with this subscription. Please chooseregions to probe from.: validation failed` when using a free account.

### DIFF
--- a/load_balancing.go
+++ b/load_balancing.go
@@ -18,7 +18,10 @@ type LoadBalancerPool struct {
 	Monitor           string               `json:"monitor,omitempty"`
 	Origins           []LoadBalancerOrigin `json:"origins"`
 	NotificationEmail string               `json:"notification_email,omitempty"`
-	CheckRegions      []string             `json:"check_regions"`
+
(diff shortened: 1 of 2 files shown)


load_balancing.go
```diff
@@ -18,6 +18,7 @@ type LoadBalancerPool struct {
 	Monitor           string               `json:"monitor,omitempty"`
 	Origins           []LoadBalancerOrigin `json:"origins"`
 	NotificationEmail string               `json:"notification_email,omitempty"`
+	CheckRegions      []string             `json:"check_regions"`
```
elithrar: Add a comment here?
```
// CheckRegions defines the geographic region(s) from where to run health-checks from - e.g. "WNAM", "WEU", "SAF", "SAM".
// Providing a null/empty value means "all regions", which may not be available to all plan types.
```

elithrar:
Thanks for this @FnuGk - just a minor documentation addition and we're good.

elithrar:
Thanks @FnuGK!

PR #163, opened 2018-02-20
Api client rate limiting

benjvi:
This allows the client to be well behaved w.r.t. api request rate and also to retry in case of errors. I've tried to put in some sensible defaults, hopefully those do make sense given typical usage

- rate limiting for clients based on golang rate.limiter (leaky bucket)
- no bursting, clients can specify a max rps as a float which is applied uniformly
  - at the moment, clients can specify anything (ultimately the api will enforce limits). are there separate limits we should put on this?
  - default: 4 rps
- in case of errors (connection error/5xx/429), the client will retry up to a specified number of times
  - this is also the default approach taken in the hashicorp retry http library, but I am unsure of this safety-wise. Thoughts??
- each time retrying, the client will exponentially backoff (pow 2), starting from a specified minimum time up to a limited specified maximum
  - defaults: 3 retries, min delay 1 second, max delay 30 seconds (config accepts delays as integer seconds)

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -8,7 +8,7 @@ import (
 	"io/ioutil"
 	"net/http"
 
-	"time"
+	"context"
 
 	"github.com/pkg/errors"
 	"golang.org/x/time/rate"
@@ -33,8 +33,7 @@ type API struct {
 	headers           http.Header
 	httpClient        *http.Client
 	authType          int
-	rateLimitRPS      int
-	rateLimitBurst    int
+	rateLimiter       *rate.Limiter
 	maxRetries        int
 }
 
@@ -45,14 +44,13 @@ func New(key, email string, opts ...Option) (*API, error) {
 	}
 
 	api := &API{
-		APIKey:         key,
-		APIEmail:       email,
-		BaseURL:        apiURL,
-		headers:        make(http.Header),
-		authType:       AuthKeyEmail,
-		rateLimitRPS:   20,
-		rateLimitBurst: 1200,
-		maxRetries:     3,
+		APIKey:      key,
+		APIEmail:    email,
+		BaseURL:     apiURL,
+		headers:     make(http.Header),
+		authType:    AuthKeyEmail,
+		rateLimiter: rate.NewLimiter(rate.Limit(4), 1), // 4rps equates to default api limit (1200 req/5 min)
+		maxRetries:  3,
 	}
 
 	err := api.parseOptions(opts...)
@@ -110,17 +108,15 @@ func (api *API) makeRequestWithAuthType(method, uri string, params interface{}, 
 	var resp *http.Response
 	var respErr error
 	var reqBody io.Reader
-	l := rate.NewLimiter(rate.Limit(api.rateLimitRPS), api.rateLimitBurst)
 	for i := 0; i < api.maxRetries; i++ {
 		if jsonBody != nil {
 			reqBody = bytes.NewReader(jsonBody)
 		}
 
-		reservation := l.Reserve()
-		if !reservation.OK() {
-			return nil, errors.New("Not able to make any request due to the current rate limit settings")
+		api.rateLimiter.Wait(context.TODO())
+		if err != nil {
+			return nil, errors.Wrap(err, "Error caused by request rate limiting")
 		}
-		time.Sleep(reservation.Delay())
 		resp, respErr = api.request(method, uri, reqBody, authType)
 		// retry if the server is rate limiting us or if it failed
 		// assumes server operations are rolled back on failure (!)
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -10,6 +10,10 @@ import (
 
 	"context"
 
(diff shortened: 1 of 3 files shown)


cloudflare.go
```diff
@@ -87,26 +101,72 @@ func (api *API) makeRequest(method, uri string, params interface{}) ([]byte, err
...
+			// expect the backoff introduced here on errored requests to dominate the effect of rate limiting
+			// dont need a random component here as the rate limiter should do something similar
+			// nb time duration could truncate an arbitrary float. Since our inputs are all ints, we should be ok
+			sleepDuration := time.Duration(math.Pow(2, float64(i-1)) * float64(api.retryPolicy.MinRetryDelay))
+
+			if sleepDuration > api.retryPolicy.MaxRetryDelay {
+				sleepDuration = api.retryPolicy.MaxRetryDelay
+			}
+			// useful to do some simple logging here, maybe introduce levels later
+			log.Printf("Sleeping %s before retry attempt number %d for request %s %s", sleepDuration.String(), i, method, uri)
```
elithrar: Strongly prefer that we didn't log directly in the library.
* Add a simple `Logger` interface to the `API` struct and provide a default no-op method.
* Accept this new `Logger` type as an `Option` - 

```
func Logger(logger *Logger) Option {
	return func(api *API) error {
		api.logger = logger
		return nil
	}
}
```

```
type Logger interface {
	Log(keyvals ...interface{}) error
}
```

cloudflare.go
```diff
@@ -87,26 +101,72 @@ func (api *API) makeRequest(method, uri string, params interface{}) ([]byte, err
...
+		}
+		resp, respErr = api.request(method, uri, reqBody, authType)
+
+		// retry if the server is rate limiting us or if it failed
+		// assumes server operations are rolled back on failure
+		if respErr != nil || resp.StatusCode == http.StatusTooManyRequests || resp.StatusCode >= 500 {
+			// if we got a valid http response, try to read body so we can reuse the connection
+			// see https://golang.org/pkg/net/http/#Client.Do
+			if respErr == nil {
+				// have to read the body before closing, but set a limit of how much to read
```
elithrar: Comment doesn't align with `ReadAll`. Did you mean to put `TODO` or use a `LimitReader` here?

cloudflare.go
```diff
@@ -87,26 +101,72 @@ func (api *API) makeRequest(method, uri string, params interface{}) ([]byte, err
...
+				respBody, err = ioutil.ReadAll(resp.Body)
+				respErr = errors.Wrap(err, "could not read response body")
+				resp.Body.Close()
+
+				log.Printf("Request: %s %s got an error response %d: %s\n", method, uri, resp.StatusCode,
+					strings.Replace(strings.Replace(string(respBody), "\n", "", -1), "\t", "", -1))
+			} else {
+				log.Printf("Error performing request: %s %s : %s \n", method, uri, respErr.Error())
```
elithrar: See comment above RE: logging.

cloudflare.go
```diff
@@ -234,3 +294,11 @@ type PaginationOptions struct {
 	Page    int `json:"page,omitempty"`
 	PerPage int `json:"per_page,omitempty"`
 }
+
+// RetryPolicy specifies number of retrys and min/max retry delays
```
elithrar: s/retrys/retries

options.go
```diff
@@ -31,6 +37,33 @@ func UsingOrganization(orgID string) Option {
...
+		// because ratelimiter doesnt do any windowing
+		// setting burst makes it difficult to enforce a fixed rate
+		// so setting it equal to 1 this effectively disables bursting
+		// this doesn't check for sensible values, ultimately the api will enforce that the value is ok
+		api.rateLimiter = rate.NewLimiter(rate.Limit(rps), 1)
+		return nil
+	}
+}
+
+// UsingRetry policy applies a non-default number of retrys and min/max retry delays
```
elithrar: `UsingRetryPolicy`

options.go
```diff
@@ -31,6 +37,33 @@ func UsingOrganization(orgID string) Option {
 	}
 }
 
+// RateLimit applies a non-default rate limit to client API requests
```
elithrar: `UsingRateLimit` (go vet / godoc format)

cloudflare.go
benjvi: I did originally mean to use a LimitReader here, however I'm not sure there is much benefit - given that we are still relying on the server sending fixed size responses for non-errored requests anyway

cloudflare.go
benjvi: I like this approach but would suggest aligning with the standard interface so that people can just pass straight in a log.Logger or a logrus.Logger (etc)

benjvi:
Just realized I missed one of the typos. @elithrar is this ok now?

nickithewatt:
Hi @elithrar, just checking if there is anything else which needs to be done here? Have all the changes been addressed?

PR #169, opened 2018-02-26
create page rule to return the page rule object - quick fix?

benjvi:
Does the simplest possible thing to get creation of page rules to be usable in terraform or other automation by returning the page rule object. I also added a bunch of tests while I'm here

I'm aware there is related some discussion and parallel work going on in #154, happy to align with that just trying to get something in quickly

Of course - this is still a breaking change

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -6,6 +6,12 @@ import (
 	cloudflare "github.com/cloudflare/cloudflare-go"
 )
 
+const (
+	user   = "cloudflare@example.org"
+	domain = "example.com"
+	apiKey = "deadbeef"
+)
+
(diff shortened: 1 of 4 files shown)


elithrar:
@benjvi approved.

Let me know when it also makes sense to tag a new release w/ the breaking changes, so you can vendor against a version (instead of a commit ref or HEAD) in the Terraform provider.

benjvi:
Awesome -thanks @elithrar . Will do - I'm not planning to make any more changes to this library at the moment but I still have a few things to look at so something else might turn up. Still could be worth tagging in the meantime? - up to you

PR #171, opened 2018-03-06
add pagination to load balancer + pool + monitor APIs

benjvi:
Noticed these APIs weren't paginating correctly, so this fixes that. This change should be non-breaking. I made the changes in much the same way as in #161 with the exception that I used the name "List[X]InPage" instead of just "List[X]" that takes a pagination argument. Given the lack of overrides in go it seems better to have a more explicit name for the one that needs pagination options, and maintain the previous (implied) behaviour of returning all resources on the List[X] call

It would be ideal to change the returned value to be a slice of pointers, but since that requires breaking changes, it probably more sense to do it as part of a larger refactoring.

Slightly unrelated, I was mislead for a while that the API docs suggest that List Page Rules returns pagination info which is not the case. Not sure how easy it would be to fix those docs but that would be helpful.

### DIFF
--- a/load_balancing.go
+++ b/load_balancing.go
@@ -4,6 +4,9 @@ import (
 	"encoding/json"
 	"time"
 
+	"net/url"
+	"strconv"
+
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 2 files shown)


load_balancing.go
```diff
@@ -193,20 +239,63 @@ func (api *API) CreateLoadBalancerMonitor(monitor LoadBalancerMonitor) (LoadBala
 	return r.Result, nil
 }
 
-// ListLoadBalancerMonitors lists load balancer monitors connected to an account.
+// ListLoadBalancerMonitorsInPage lists a single page of load balancer monitors connected to an account.
 //
 // API reference: https://api.cloudflare.com/#load-balancer-monitors-list-monitors
-func (api *API) ListLoadBalancerMonitors() ([]LoadBalancerMonitor, error) {
+func (api *API) ListLoadBalancerMonitorsInPage(pageOpts PaginationOptions) ([]LoadBalancerMonitor, ResultInfo, error) {
```
elithrar: What's the goal of having `xxxxInPage` methods?

benjvi: Allow clients to choose their own pagination strategy/ retrieve a single page of interest? Could also be useful if you just want the ResultInfo from a single call

Issue #173, opened 2018-03-22
custom_hostnames endpoint now returns HTTP 201 not 200 and error being thrown


prdonahue:
2018/03/22 13:50:27 error from makeRequest: HTTP status 201: content "{\n  \"result\": ...

PR #174, opened 2018-03-22
fix(api): accept 2xx responses without error

terinjokes:
Many of our API endpoints that create objects now return 201 Created
instead of 200 OK. The API implementation here was previously only
accepting http.StatusOK as valid.

This change allows all currently defined 2xx status codes.

Bug: #173

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -109,15 +109,18 @@ func (api *API) makeRequestWithAuthType(method, uri string, params interface{}, 
 		return nil, errors.Wrap(err, "could not read response body")
 	}
 
-	switch resp.StatusCode {
-	case http.StatusOK:
-		break
-	case http.StatusUnauthorized:
+	switch {
+	case resp.StatusCode >= http.StatusOK && resp.StatusCode < http.StatusMultipleChoices:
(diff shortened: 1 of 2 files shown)


cloudflare.go
```diff
@@ -109,22 +109,22 @@ func (api *API) makeRequestWithAuthType(method, uri string, params interface{},
...
-	case http.StatusForbidden:
-		return nil, errors.Errorf("HTTP status %d: insufficient permissions", resp.StatusCode)
-	case http.StatusServiceUnavailable, http.StatusBadGateway, http.StatusGatewayTimeout,
-		522, 523, 524:
-		return nil, errors.Errorf("HTTP status %d: service failure", resp.StatusCode)
-	default:
-		var s string
-		if body != nil {
-			s = string(body)
+	if resp.StatusCode < http.StatusOK || resp.StatusCode > http.StatusIMUsed {
```
terinjokes: Alternatively, I can do a true switch:

```go
switch {
	case resp.StatusCode >= http.StatusOK && resp.StatusCode <= http.StatusIMUsed:

elithrar: The switch would be preferred, if you can.

PR #179, opened 2018-05-04
[bugfix] Fix session_affinity none handling

elithrar:
This fixes a bug where the marshalled JSON request sets the "session_affinity" field to `null`, which is not supported by the API as per https://github.com/terraform-providers/terraform-provider-cloudflare/pull/57#issuecomment-386656370


elithrar:
Fixed in a simpler manner.

PR #180, opened 2018-05-07
Fix Zone Owner object

fholzer:
1. `owner_type` was renamed to `type`, though this isn't reflected in the documentation yet
2. When a zone is owned by an organization instead of a user, the `email` field is replaced by a `name` field


patryk:
Thanks for PR, but for backwards compatibility reasons I have renamed `Type` back to `OwnerType`.

Issue #135, opened 2017-06-13
SDK inconsistency in zone vs record APIs


ideahitme:
**Description:**

Zones sdk code does not retrieve all zones, only the first page. 
DNS records sdk code does retrieve all records

**Details:**
This is a snippet from `zone.go` `ListZones(z ...string)` method. 

```go
		// TODO: Paginate here. We only grab the first page of results.
		// Could do this concurrently after the first request by creating a
		// sync.WaitGroup or just a channel + workers.
		res, err = api.makeRequest("GET", "/zones", nil)
		if err != nil {
			return []Zone{}, errors.Wrap(err, errMakeRequestError)
		}
		err = json.Unmarshal(res, &r)
		if err != nil {
			return []Zone{}, errors.Wrap(err, errUnmarshalError)
		}
		zones = r.Result
```

As noted in TODOs it does not implement pagination, moreover the struct where JSON is unmarshaled does not define (as opposed to response from dns records API): 

```go
// ResultInfo contains metadata about the Response.
type ResultInfo struct {
	Page       int `json:"page"`
	PerPage    int `json:"per_page"`
	TotalPages int `json:"total_pages"`
	Count      int `json:"count"`
	Total      int `json:"total_count"`
}
```

1. I would assume `ResultInfo` is still part of the http response from `"https://api.cloudflare.com/client/v4/zones` and can be retrieved ? 
2. Are u planning to have this fixed, I could create a PR if it is not yet planned ?

Tenzer:
Related issues: #6 and #4.

ideahitme:
As per official documentation https://api.cloudflare.com/#zone-list-zones it has 
```
type ResultInfo struct {
	Page       int `json:"page"`
	PerPage    int `json:"per_page"`
	TotalPages int `json:"total_pages"`
	Count      int `json:"count"`
	Total      int `json:"total_count"`
}
```
as json, so in theory it should be pretty simple to implement

ideahitme:
@Tenzer I see,  feel free to close this if you think it is necessary

PR #182, opened 2018-05-16
[feat] ListZones internal pagination

Oldsouldier:
Collect all available zones sequentially and return them to the user,
rather than just the first page.  The maximum page size is used
to minimize API calls.

This matches the DNSRecords implementation, and allows simple
access to all zones until general pagination can be implemented.

### DIFF
--- a/zone.go
+++ b/zone.go
@@ -86,7 +86,8 @@ type ZoneResponse struct {
 // ZonesResponse represents the response from the Zone endpoint containing an array of zones.
 type ZonesResponse struct {
 	Response
-	Result []Zone `json:"result"`
+	Result     []Zone `json:"result"`
+	ResultInfo `json:"result_info"`
 }
 
 // ZoneIDResponse represents the response from the Zone endpoint, containing only a zone ID.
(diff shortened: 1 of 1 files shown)


Oldsouldier:
Related issues: #6, #4 , and #135

Oldsouldier:
@elithrar Any thoughts on this? Seems like the issue has been open a while. If progress has been made on a general pagination technique i can align to that instead.

elithrar:
Hey @Oldsouldier - I'm no longer at Cloudflare, so may not be able to help with this directly. @tomlikestorock may be able to take a look at this.

njuettner:
It would be nice to merge it :).

We use your library in one of our projects, see:
https://github.com/kubernetes-incubator/external-dns/blob/master/provider/cloudflare.go

Recently we were asked why it only returns the first page to list zones. Unfortunately the only option would be to use the API directly.

terinjokes:
I'd prefer not baking automatic pagination into individual API endpoints, but instead allow users to pass in the paging information. Then other functions could be written to wrap an automatic paginate.

Over in the JS bindings, I'd been thinking over using function options on the endpoint functions, to allow these to be pretty flexible. Below would be an example of this idea in Go.

```go
// ListZonesContext lists zones on an account with context and option support
func (api *API) ListZonesContext(ctx context.Context, z string, opt... Option) (ZonesResponse, error) {
  // Return a response type that includes the API meta information, that includes
  // the pagination information.
}

// WithZonesFilter takes a list of zone names
// to filter against.
func WithZonesFilter(z... string) Option {
  return func(opt requestOpt) {
    opt.zones = z
  }
}

// WithPagination allows the consumer to specify the pagination options
// for the request.
func WithPagination(page, count int) Option {
  return func(opt requestOpt) {
    opt.perPage = 25
    opt.page = 3
  }
}
```

Oldsouldier:
@terinjokes ,  I completely agree with that approach. However, I would suggest either replacing the existing ListZones, or merging this PR and adding the new semantics afterwards. The current implementation is not reliable.

njuettner:
@terinjokes is there any progress on this PR? I really like to see it gets merged. I agree you proposal is better than the change in this PR but I would also like to have solution which works right now than no change at all.

J-Ha:
Could we get some updates on this?
We have trouble with this issue since April and need this feature. 
Or we have to settle over to an another provider.

rvalitov:
6 months have passed and still it's not merged and no updates. That's more than enough to make a decision. This repository is no longer supported by CloudFlare?

rvalitov:
Thanks for the update!

patryk:
Since we've decided to go alternative route, I'm closing this PR, but feel free to ping me if it's not the case. Thanks all for insights.

PR #183, opened 2018-05-22
Add partial support for VirtualDNS analytics

seiffert:
We are implementing a Prometheus exporter for our internal monitoring tooling and want to monitor Virtual DNS metrics. For this use case, general metrics are sufficient right now which is why we didn't add dimensions or filters. However, we still wanted to contribute our changes in case they help somebody.

### DIFF
--- a/virtualdns.go
+++ b/virtualdns.go
@@ -2,6 +2,9 @@ package cloudflare
 
 import (
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


patryk:
Looks great, thanks for your contribution, and I apologize for late response. 🐢

PR #188, opened 2018-06-09
Added support for Workers and Worker Routes

### LINKED ISSUE #187: Implement Workers endpoints
- https://developers.cloudflare.com/workers/api/
- https://developers.cloudflare.com/workers/api/config-api-for-enterprise/
- https://api.cloudflare.com/#worker-script-properties

Sjeanpierre:
### Adds support for Cloudflare workers 

- https://developers.cloudflare.com/workers/api/
- https://api.cloudflare.com/#worker-script-properties

This does not yet support the enterprise routes[1] for workers as they are currently broken in the Cloudflare API

[1] - https://developers.cloudflare.com/workers/api/config-api-for-enterprise/

fixes #187


prdonahue:
Thanks, @Sjeanpierre! Will ask someone to take a look and see if we can get this merged in (or if changes are needed).

Can you elaborate on your comment 

> This does not yet support the enterprise routes[1] for workers as they are currently broken in the Cloudflare API?

What specifically is broken? Do you have a repro?

Sjeanpierre:
Hi @prdonahue thanks for commenting so quickly

yes as far as the enterprise routes go I have support ticket 1534191 open about that issue, but the gist of the matter is that the enterprise specific routes don't work, support has already confirmed that they are able to reproduce. reproduction steps below.

```when attempting to list scripts by doing the 
curl -X GET "https://api.cloudflare.com/client/v4/organization/ORG_ID/workers/scripts" -H "X-Auth-Email:USER_EMAIL" -H "X-Auth-Key:xxxREDACTEDxxx"

I get back the following error
{"success":false,"errors":[{"code":7000,"message":"No route for that URI"}],"messages":[],"result":null}

```

prdonahue:
@Sjeanpierre can you try .../organizations/... in lieu of .../organization/...?

Sjeanpierre:
@prdonahue that seems to work better, I guess the route is just incorrect in the documentation?

I'm getting a new error about `workers.api.error.entitlement`, but at least the route resolves now...

prdonahue:
@Sjeanpierre yes, the docs are incorrect. We'll fix, thanks(!) and we'll have someone test those routes before merging.

rita3ko:
@Sjeanpierre thanks for calling that out. I've updated the [docs](https://developers.cloudflare.com/workers/api/config-api-for-enterprise/)!

Sjeanpierre:
@jamesog thanks for the review, `golint` warnings now addressed in https://github.com/cloudflare/cloudflare-go/pull/188/commits/6f34a2b85f21db1fc4b8d870f111bb50896c5b4f

jamesog:
LGTM, maybe someone who knows Workers better than I would like to review.

rita3ko:
LGTM from a Workers standpoint. @Sjeanpierre if you could add the Enterprise endpoints, that would be awesome. 
Once that's done, I can also add Workers to flarectl.

Sjeanpierre:
@rita3ko - sure I will add the enterprise endpoints, one small issue is that I am not sure what the enterprise responses look like since my enterprise org is getting `workers.api.error.entitlement` when I try the request, and the API documentation does not include response data.

@garrettgalow can you elaborate on this point? are you saying I should be using `/accounts/ instead of `/organizations` when implementing the enterprise endpoints? any documentation on this?

garrettgalow:
@Sjeanpierre we are updating the workers docs to reflect this, but yes you can basically find/replace 'organizations' with 'accounts' anywhere you see it. The IDs are the same.

You should be able to see the accounts api response since it wouldn't be blocked AFAIK. I don't have any scripts on my account, so I just got an empty response payload.

Sjeanpierre:
When doing 
`curl -X GET "https://api.cloudflare.com/client/v4/accounts/:account_id/workers/scripts" -H "X-Auth-Email:YOUR_CLOUDFLARE_EMAIL" -H "Content-Type:application/javascript" -H "X-Auth-Key:ACCOUNT_AUTH_KEY"`

I still get
image

I'll reach out to support to determine if my account is somehow broken, does this request work correctly for you guys internally?

Sjeanpierre:
I am still getting the same errors from the API endpoint. My support ticket has been open for days and still no resolutions `1534191` maybe one of you can help me escalate?

Once I get things working I can add the enterprise endpoints to the code

thanks, @prdonahue @rita3ko @jamesog

rita3ko:
@Sjeanpierre I'll follow up on the ticket.

Sjeanpierre:
Added additional support for `accounts` routes which are intended for enterprise account use - https://developers.cloudflare.com/workers/api/config-api-for-enterprise/ 

@rita3ko @jamesog @garrettgalow   - This PR should be good to go now, please review the new functions for the enterprise calls and let me know if you'd like anything changed

thanks!

rita3ko:
Hi @Sjeanpierre -

A couple notes:
- Rather than having separate functions for`UploadWorkerWithName` for Enterprise, and `UploadWorker`, it may make more sense to take in an optional variable "name", and handle the logic based on that. @jamesog not sure if you have thoughts on best practices here.
- Can you also add support for ENT routes? I.e. with ENT, the payload contains the script name `"script": "helloworld"`, rather than `"enabled": true` (I'm currently updating the docs to reflect that).

Thanks!

Sjeanpierre:
@rita3ko Thanks for the feedback, Go doesn't really do optional parameters as easily as other languages, but let's see if @jamesog or anyone else has strong opinions on best practice here. 

Thanks for updating the docs I will make the changes needed to support enterprise routes

jamesog:
I think firewall.go does this because it takes the organization as a separate named parameter and uses a different route. As this one is using the same route but with a name appended to the URL I'd be inclined to make it one function that takes a `scriptName string`, which you can pass as `""` for non-Enterprise - or you can cheat and make it a variadic which makes it optional. I'm not sure if Go-purists would like that though :-)

Also just a note on the doc strings on the functions - please make sure they're full sentences with punctuation. I've just noticed some don't have the full stop at the end.

rita3ko:
One thing @elithrar suggested was to use a struct as the argument you pass in, and have it contain an additional field if it's ENT. This will also allow to easier support possible future additional fields.

workers.go
```diff
@@ -0,0 +1,259 @@
...
+	if err != nil {
+		return r, errors.Wrap(err, errUnmarshalError)
+	}
+	return r, nil
+}
+
+// DownloadWorker fetch raw script content for your worker returns []byte containing worker code js
+//
+// API reference: https://api.cloudflare.com/#worker-script-download-worker
+func (api *API) DownloadWorker(zoneID string) (WorkerScriptResponse, error) {
```
elithrar: To expand on my suggestion here:

`DownloadWorker(zoneID string, conf *WorkerConf) (WorkerScriptResponse, error)`

`*WorkerConf` can be something like -

```
type WorkerConf struct {
    ScriptName string
    // Future options here
}
```

This avoids having double the methods, which also has potential to change in the future if Cloudflare opened up multi-script/supported it as the default. If a non `nil` `*WorkerConf` is provided, use the name in there.

Keeping the API surface small, but making the default case "easy" -- which is single-script, zoneID based today - should strike a good balance.

Sjeanpierre:
@patryk should be good to go now

patryk:
Looks good, thanks Stevenson.

PR #194, opened 2018-06-21
Add some of the missing setting to 'page_rule' api

grogancolin:
This adds in (I think all) of the missing Page Rule settings.

I got the list of settings from the Create Page UI in my Cloudflare dashboard, so it may differ if some are hidden from there.

### DIFF
--- a/page_rules.go
+++ b/page_rules.go
@@ -23,26 +23,38 @@ type PageRuleTarget struct {
 PageRuleAction is the action to take when the target is matched.
 
 Valid IDs are:
-
   always_online
   always_use_https
   browser_cache_ttl
   browser_check
+  bypass_cache_on_cookie
+  cache_by_device_type
+  cache_deception_armor
   cache_level
+  cache_on_cookie
   disable_apps
   disable_performance
   disable_railgun
   disable_security
   edge_cache_ttl
   email_obfuscation
+  explicit_cache_control
   forwarding_url
+  host_header_override
   ip_geolocation
   mirage
+  opportunistic_encryption
+  origin_error_page_pass_thru
+  polish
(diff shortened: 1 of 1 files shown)


page_rules.go
```diff
@@ -52,26 +65,40 @@ type PageRuleAction struct {
...
+	"opportunistic_encryption":    "Opportunistic Encryption",    // Value of type string
+	"origin_error_page_pass_thru": "Origin Error Page Pass-thru", // Value of type string
+	"polish":                      "Polish",                      // Value of type string
+	"resolve_override":            "Resolve Override",            // Value of type string
+	"respect_strong_etag":         "Respect Strong ETags",        // Value of type string
+	"response_buffering":          "Response Buffering",          // Value of type string
+	"rocket_loader":               "Rocker Loader",               // Value of type string
+	"security_level":              "Security Level",              // Value of type string
+	"server_side_exclude":         "Server Side Excludes",        // Value of type string
+	"smart_errors":                "Smart Errors",                // Value of type string
```
dknecht: This doesn't exist anymore.

grogancolin:
All of them? They're available in my UI (or rather, they were when I checked last week).
Is there anywhere with a complete list of available page rule options and I get can working on correcting that!

Thanks!

page_rules.go
grogancolin: All of them? They're available in my UI (or rather, they were when I checked last week).
Is there anywhere with a complete list of available page rule options and I get can working on correcting that!

Thanks!

rita3ko: Just Smart Errors.

patryk:
Thanks, Colin!

PR #195, opened 2018-06-26
Add `Accounts` API endpoints

jacobbednarz:
This introduces the ability to manage Cloudflare Accounts via the API. 

I initially went to implement the Organisations however I noticed the API docs mention this is deprecated and instead we should use Accounts.

From https://api.cloudflare.com/#organizations-properties:

> NOTE: This API is deprecated, please use equivalent /accounts API endpoints where possible. Account APIs provide a broader range of features, and are backwards-compatible to organization APIs.

So instead, this PR implements (and tests) the following:

- List all accounts that the current logged in user has access to 
  (https://api.cloudflare.com/#accounts-list-accounts)
- A single account lookup based on account ID 
  (https://api.cloudflare.com/#accounts-account-details)
- Update an account based on the account ID 
 (https://api.cloudflare.com/#accounts-update-account)

I haven't added the functionality to `flarectl` as it seems slightly neglected and I'm unsure if that is still something that is being maintained. If we would like the functionality to be added, I'm more than happy to implement it.

### DIFF
--- a/accounts.go
+++ b/accounts.go
@@ -43,10 +43,10 @@ type AccountDetailResponse struct {
 	Result   Account  `json:"result"`
 }
 
-// ListAccounts returns all accounts the logged in user has access to.
+// Accounts returns all accounts the logged in user has access to.
 //
 // API reference: https://api.cloudflare.com/#accounts-list-accounts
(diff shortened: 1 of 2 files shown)


patryk:
Looks great. Thanks Jacob (again)! 🍻

PR #202, opened 2018-07-04
Add ssl verification details endpoint

patricelee:
Updates the library with the ability to fetch ssl verification details.

- ssl Verification Details https://api.cloudflare.com/#ssl-verification-ssl-verification-details

### DIFF
--- a/ssl_verification.go
+++ b/ssl_verification.go
@@ -1,0 +1,41 @@
+package cloudflare
+
+import (
+	"encoding/json"
+
+	"github.com/pkg/errors"
+)
+
+// ZoneSSLVerification represents custom SSL Verfication metadata.
+type ZoneSSLVerification struct {
+	CertificateStatus string `json:"certificate_status"`
+	CertPackUUID      string `json:"cert_pack_uuid"`
+	ValidationMethod  string `json:"validation_method"`
+	ValidationType    string `json:"validation_type"`
+	VerificationInfo  struct {
+		HTTPURL  string `json:"http_url"`
+		HTTPBody string `json:"http_body"`
+	} `json:"verification_info"`
+	Hostname string `json:"hostname"`
+}
+
+type zoneSSLVerificationResponse struct {
+	Response
+	Result []ZoneSSLVerification `json:"result"`
+}
+
+// ListSSLVerification get SSL Verification Info for a Zone
+//
+// API reference: https://api.cloudflare.com/#ssl-verification-ssl-verification-details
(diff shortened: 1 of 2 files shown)


ssl_verification.go
```diff
@@ -0,0 +1,41 @@
...
+
+type zoneSSLVerificationResponse struct {
+	Response
+	Result []ZoneSSLVerification `json:"result"`
+}
+
+// ListSSLVerification get SSL Verification Info for a Zone
+//
+// API reference: https://api.cloudflare.com/#ssl-verification-ssl-verification-details
+func (api *API) ListSSLVerification(zoneID string) ([]ZoneSSLVerification, error) {
```
jacobbednarz: According to the API documentation, this call supports an optional boolean parameter of `retry`. Is this something you'd be able to add to this function to ensure we get full coverage here?

ssl_verification_test.go
```diff
@@ -0,0 +1,60 @@
...
+				{
+					"certificate_status": "pending_validation",
+					"cert_pack_uuid": "4449b368-d10f-4a73-9d27-f3f2752443df",
+					"validation_method": "http",
+					"validation_type": "dv",
+					"verification_info": {
+						"http_url": "http:\/\/example.com\/.well-known\/pki-validation\/ca3-0fdf5b3dc4d0428fa81ac381a8c4e333.txt",
+						"http_body": "ca3-ff2d513629514e3b90675a982f9f0644"
+					},
+					"hostname": "example.com"
```
jacobbednarz: This response object seems to differ what is coming back from the API and is documented. Here is what I'm getting from the API with an active verification using some anonymised fields:

```json
{
  "certificate_status": "active",
  "signature": "ECDSAWithSHA256",
  "brand_check": true,
  "verification_info": {
    "record_name": "abc123.example.com",
    "record_target": "72AA6FC86BB81C0BA35B6DBB66481234.comodoca.com"
  },
  "verification_status": null,
  "verification_type": "cname"
}
```

The differences look to be:

- `brand_check` is missing and a boolean 
- `verification_info` object has different key values (`record` instead of `http`) which will make a difference in your `ZoneSSLVerification` struct. Perhaps we need to account for both of these structs separately and allow `VerificationInfo` to accept either?
- `cert_pack_uuid` is missing from my response and the documentation.

I'm unsure why there are differences but I've tested this response on an Enterprise and Personal paid plans with no differences to the response. Are you on an older SSL product or something along those lines that we may need to account for here?

Issue #33, opened 2016-05-03
Implement Zone Subscription endpoints


Tenzer:
https://api.cloudflare.com/#zone-subscription

PR #205, opened 2018-07-13
Adding Cloudflare Zone Subscription API calls.

justnom:
This is my attempt to add a subset of the [Zone Subscription API calls](https://api.cloudflare.com/#zone-subscription-zone-subscription-details).

There is an issue for this already #33.

### DIFF
--- a/zone.go
+++ b/zone.go
@@ -231,6 +231,31 @@ type PurgeCacheResponse struct {
 	} `json:"result"`
 }
 
+// ZoneSubscriptionDetails represents all the subscription details for the zone.
+type ZoneSubscriptionDetails struct {
+	ID       string `json:"id,omitempty"`
+	State    string `json:"state,omitempty"`
+	Price    int    `json:"price,omitempty"`
+	Currency string `json:"currency,omitempty"`
+	Zone     Zone   `json:"zone,omitempty"`
+	RatePlan struct {
+		ID                string `json:"id,omitempty"`
+		Name              string `json:"public_name,omitempty"`
+		Currency          string `json:"currency,omitempty"`
+		ExternallyManaged bool   `json:"externally_managed,omitempty"`
+		Scope             string `json:"scope,omitempty"`
+	} `json:"rate_plan,omitempty"`
+	Frequency          string    `json:"frequency,omitempty"`
+	CurrentPeriodStart time.Time `json:"current_period_end,omitempty"`
+	CurrentPeriodEnt   time.Time `json:"current_period_start,omitempty"`
+}
+
+// ZoneSubscriptionDetailsResponse represents the response from the zone subscription endpoint.
+type ZoneSubscriptionDetailsResponse struct {
+	Response
+	Result ZoneSubscriptionDetails `json:"result"`
+}
+
 // newZone describes a new zone.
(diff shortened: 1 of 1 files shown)


zone.go
```diff
@@ -231,6 +231,31 @@ type PurgeCacheResponse struct {
...
+// ZoneSubscriptionDetails represents all the subscription details for the zone.
+type ZoneSubscriptionDetails struct {
+	ID       string `json:"id,omitempty"`
+	State    string `json:"state,omitempty"`
+	Price    int    `json:"price,omitempty"`
+	Currency string `json:"currency,omitempty"`
+	Zone     Zone   `json:"zone,omitempty"`
+	RatePlan struct {
+		ID                string `json:"id,omitempty"`
+		Name              string `json:"public_name,omitempty"`
```
jacobbednarz: I'm unsure whether it's a good idea to introduce a field that doesn't match the JSON response's key as most of the existing structs follow a 1:1 mapping. Is there a particular reasoning behind this?

zone.go
```diff
@@ -231,6 +231,31 @@ type PurgeCacheResponse struct {
 	} `json:"result"`
 }
 
+// ZoneSubscriptionDetails represents all the subscription details for the zone.
+type ZoneSubscriptionDetails struct {
```
jacobbednarz: This struct appears to be missing the `component_values` section. Are you seeing differences in the response to what is documented or has this been left out intentionally?

zone.go
```diff
@@ -231,6 +231,31 @@ type PurgeCacheResponse struct {
...
+	RatePlan struct {
+		ID                string `json:"id,omitempty"`
+		Name              string `json:"public_name,omitempty"`
+		Currency          string `json:"currency,omitempty"`
+		ExternallyManaged bool   `json:"externally_managed,omitempty"`
+		Scope             string `json:"scope,omitempty"`
+	} `json:"rate_plan,omitempty"`
+	Frequency          string    `json:"frequency,omitempty"`
+	CurrentPeriodStart time.Time `json:"current_period_end,omitempty"`
+	CurrentPeriodEnt   time.Time `json:"current_period_start,omitempty"`
```
jacobbednarz: This looks like a typo. `s/CurrentPeriodEnt/CurrentPeriodEnd`

zone.go
```diff
@@ -231,6 +231,31 @@ type PurgeCacheResponse struct {
...
+	Zone     Zone   `json:"zone,omitempty"`
+	RatePlan struct {
+		ID                string `json:"id,omitempty"`
+		Name              string `json:"public_name,omitempty"`
+		Currency          string `json:"currency,omitempty"`
+		ExternallyManaged bool   `json:"externally_managed,omitempty"`
+		Scope             string `json:"scope,omitempty"`
+	} `json:"rate_plan,omitempty"`
+	Frequency          string    `json:"frequency,omitempty"`
+	CurrentPeriodStart time.Time `json:"current_period_end,omitempty"`
```
jacobbednarz: The key name doesn't match the correct JSON field. You'll need to swap these two around so that the start reads from the start JSON field.

jacobbednarz:
Would you mind adding some test coverage for these changes?

PR #208, opened 2018-07-17
Add correlate (nat support) to rate limiting

AshleyPoole:
Add correlate (nat support) to rate limiting so that the property can be configured.

I have another PR to add this to the Terraform Cloudflare provider.

### DIFF
--- a/rate_limiting.go
+++ b/rate_limiting.go
@@ -18,6 +18,7 @@ type RateLimit struct {
 	Threshold   int                     `json:"threshold"`
 	Period      int                     `json:"period"`
 	Action      RateLimitAction         `json:"action"`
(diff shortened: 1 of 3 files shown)


AshleyPoole:
@patryk would you be able to give this a glance over and merge if happy? Also, how do I go about updating the package version or is this handled by the build definition?

AshleyPoole:
This PR is to resolve issue 212 (https://github.com/cloudflare/cloudflare-go/issues/212)

patryk:
Hey @AshleyPoole. Looks good, merged. Sorry for the delay.

As for Terraform provider, we need to update `/vendor` subdirectory in https://github.com/terraform-providers/terraform-provider-cloudflare/tree/master/vendor with `govendor`. I can do it.

PR #210, opened 2018-07-20
Add support for WAF read and update

gerardc:
- Add support for WAF read and update
- Add test coverage for `ListWAFRules` and `ListWAFPackages` functions
- Update documentation for WAF endpoints


gerardc:
cc @patryk are to you review? 🙇

waf.go
```diff
@@ -39,14 +39,23 @@ type WAFRule struct {
...
-// WAFRulesResponse represents the response from the WAF rule endpoint.
+// WAFRulesResponse represents the response from the WAF rules endpoint.
 type WAFRulesResponse struct {
 	Response
 	Result     []WAFRule  `json:"result"`
 	ResultInfo ResultInfo `json:"result_info"`
 }
 
+// WAFRuleResponse represents the response from the WAF rule endpoint.
+type WAFRuleResponse struct {
```
jacobbednarz: In some other examples, the pattern as been to name the singular resource (in this case a WAF rule) as such but then the multiple with `List` in the name. For instance, [`AccessRuleResponse `](https://github.com/cloudflare/cloudflare-go/blob/master/firewall.go#L41-L45) (singular) and [`AccessRuleListResponse `](https://github.com/cloudflare/cloudflare-go/blob/master/firewall.go#L49-L53) (multiple in the response). Do you think it's worth updating here and version bumping to signify a change to the `cloudflare-go` API?

waf.go
```diff
@@ -95,3 +106,47 @@ func (api *API) ListWAFRules(zoneID, packageID string) ([]WAFRule, error) {
 	}
 	return rules, nil
 }
+
+// WAFRule returns a WAF rule from the given WAF package.
+//
+// API Reference: https://api.cloudflare.com/#waf-rules-rule-details
+func (api *API) WAFRule(zoneID, packageID, ruleID string) (WAFRule, error) {
+	var r WAFRuleResponse
```
jacobbednarz: I tend to define the variables as close as possible to the use of them to make things easier for reviewing but that's just a personal nitpick you can ignore if you'd like.

waf.go
```diff
@@ -95,3 +106,47 @@ func (api *API) ListWAFRules(zoneID, packageID string) ([]WAFRule, error) {
...
+}
+
+// UpdateWAFRule lets you update the mode of a WAF Rule.
+//
+// API Reference: https://api.cloudflare.com/#waf-rules-edit-rule
+func (api *API) UpdateWAFRule(zoneID, packageID, ruleID, mode string) (WAFRule, error) {
+	var r WAFRuleResponse
+	uri := "/zones/" + zoneID + "/firewall/waf/packages/" + packageID + "/rules/" + ruleID
+
+	params := struct {
```
jacobbednarz: I'm not sure how I feel about this inline definition as it's not documented and inside of a function as opposed to exportable (like most of the other parts of this library). Was there a particular reason for not allowing this to be exported? 

There are quite a [few examples within this library](https://github.com/cloudflare/cloudflare-go/blob/89e1fce42324c0da504a3405f38e2a3910be7091/zone.go#L353-L358) where the PATCH requests rely on an exported struct and I'm thinking it might be best maintain that unless we've got a good reason to go against it.

waf_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+func TestListWAFPackages(t *testing.T) {
+	setup()
+	defer teardown()
+
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, "GET", r.Method, "Expected method 'GET', got %s", r.Method)
+
+		w.Header().Set("content-type", "application/json")
+		// JSON data from: https://api.cloudflare.com/#zone-analytics-properties
```
jacobbednarz: This comment doesn't match what is getting outputted below. We can probably do without it unless there is some knowledge or caveat to be conveyed here.

waf_test.go
```diff
@@ -0,0 +1,263 @@
...
+      "result_info": {
+        "page": 1,
+        "per_page": 20,
+        "count": 1,
+        "total_count": 2000
+      }
+    }`)
+	}
+
+	mux.HandleFunc("/zones/"+testZoneID+"/firewall/waf/packages", handler)
```
jacobbednarz: `testZoneID` is used in a few places and looks like it's relying on the value coming from the rate limiting test file. Should we look to localise this to the tests it's running? I think this might be accidentally working and not something we want to rely on continuing to be there in the future.

waf_test.go
```diff
@@ -0,0 +1,263 @@
...
+
+func TestListWAFRules(t *testing.T) {
+	setup()
+	defer teardown()
+
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, "GET", r.Method, "Expected method 'GET', got %s", r.Method)
+
+		w.Header().Set("content-type", "application/json")
+		// JSON data from: https://api.cloudflare.com/#zone-analytics-properties
```
jacobbednarz: Same comment as https://github.com/cloudflare/cloudflare-go/pull/210/files#r204581841, it's not matching what is below.

gerardc:
@jacobbednarz thanks for the feedback I'll make those changes and ping you when ready for re-review 🙇

waf.go
gerardc: I've made the change 👍 note that this will be a breaking change

gerardc:
@jacobbednarz @patryk ready for re-review

waf.go
patryk: While I am fully sympathetic to the rename, I would recommend against it at this point. We'll have a better chance in the future when we'll do more substantial code cleanups.

Please keep `WAFRulesResponse` for now.

gerardc:
@patryk thanks for the feedback, I've made the change to use `WAFRulesResponse`

PR #213, opened 2018-08-04
Add support to makeRequest for passing a raw request body.

ejcx:
While adding support to `cf` for workers I noticed I could not
get UploadWorker to succeed. I tracked down the issue to be all
the confusing string interpolation going on here.

 - Script string is passed to UploadWorker
 - UploadWorker passes script string to makeRequest as bytes
 - makeRequest json.Marshals this byte array which mangles it

If we attempt to cast the params passed to makeRequest as a byte
array and just avoid marshalling when the params looks like bytes
then we can upload the script raw without issues.


ejcx:
Anything holding this up?

PR #215, opened 2018-08-08
Fix worker enterprise API behavior

jRiest:
Workers have a separate API for enterprise customers that supports multi-script functionality https://developers.cloudflare.com/workers/api/config-api-for-enterprise/

This PR addresses 2 issues with using the enterprise API
- Currently if an enterprise user tries to set a `ScriptName` but forgets to set their `organizationID`, calls to create/update/delete a worker will just silently ignore their `ScriptName` parameter which is probably not what they want. This PR changes it so that if a `ScriptName` is provided, it will return an error unless the `organizationID` is also set
- The enterprise API has different endpoints for setting routes that allows for mapping routes to a particular script. This PR will use those endpoints (`/routes/`) instead of the non-enterprise endpoints (`/filters/`) when an `organizationID` is set

### DIFF
--- a/workers.go
+++ b/workers.go
@@ -65,8 +65,8 @@ type WorkerScriptResponse struct {
 //
 // API reference: https://api.cloudflare.com/#worker-script-delete-worker
 func (api *API) DeleteWorker(requestParams *WorkerRequestParams) (WorkerScriptResponse, error) {
-	// if organization ID is defined and ScriptName is provided we will treat as org request
-	if api.organizationID != "" && requestParams.ScriptName != "" {
+	// if ScriptName is provided we will treat as org request
+	if requestParams.ScriptName != "" {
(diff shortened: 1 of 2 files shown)


avidal:
> This PR changes it so that if a ScriptName is provided, it will return an error unless the organizationID is also set

I don't see any changes to that effect in the PR. Did you miss this one or are you relying on the API itself to return the error?

jRiest:
@avidal those errors are already returned in the `deleteWorkerWithName`, `uploadWorkerWithName` and `downloadWorkerWithName` functions, but without this PR, it will only call those functions if `ScriptName` is set *AND* `organizationID` is set. This PR changes it to call those functions whenever `ScriptName` is set

rita3ko:
@jamesog can you help take a look?

PR #221, opened 2018-08-21
Added support for universal ssl setting

FnuGk:
Added support for getting and updating universal ssl setting for a specified zone.

API reference: https://api.cloudflare.com/#universal-ssl-settings-for-a-zone-properties


patryk:
Looks good. Thanks Jesper!

PR #222, opened 2018-08-21
Add settings object to the custom_hostnames payload

nickysemenza:
The custom_hostnames api can accept a `settings` object in the create and update methods. It also returns the `settings` object in its responses.
I've updated the tests (and manually tested against the real API)


patryk:
Looks good, thanks!

PR #224, opened 2018-08-27
worker bug fixes

jRiest:


workers.go
```diff
@@ -171,7 +171,11 @@ func (api *API) UploadWorker(requestParams *WorkerRequestParams, data string) (W
 		return api.uploadWorkerWithName(requestParams.ScriptName, data)
 	}
 	uri := "/zones/" + requestParams.ZoneID + "/workers/script"
-	api.headers.Add("Content-Type", "application/javascript")
+
+	defer func(ct string) {
+		api.headers.Set("Content-Type", ct)
+	}(api.headers.Get("Content-Type"))
+	api.headers.Set("Content-Type", "application/javascript")
```
deuill: I realize you're working with what's already here, but is it guaranteed this won't affect other calls to the shared `API` instance, possibly in other Goroutines? I'd think that creating a single instance and farming out work concurrently is a fairly common pattern.

It may be better to simply extend the `API.makeRequest` function or create a new function, called `makeRequestWithHeaders` (ugh) to allow for setting user-defined headers per-handler without having to resort to setting and re-setting shared variables (which are cloned down the line, anyways).

cloudflare.go
```diff
@@ -247,12 +257,10 @@ func (api *API) userBaseURL(accountBase string) string {
 
 // cloneHeader returns a shallow copy of the header.
```
avidal: doc comment still says `cloneHeader` but the func changed to `copyHeader`

PR #229, opened 2018-09-29
Improve API errors handling

jakub-bacic:
Return custom error in case of API error so it's possible to easily handle them later.

It will allow users to handle API errors (e.g. `81057 - The record already exists`) without the need to manually parse error strings:

```go
const (
    ErrRecordAlreadyExists = 81057
)

func AlreadyExistsError(err error) bool {
    if err, ok := errors.Cause(err).(*cloudflare.APIError); ok {
        return err.HasErrorCode(ErrRecordAlreadyExists)
    }
    return false
}


resp, err := api.CreateDNSRecord(zoneId, dnsRecord)
if err != nil && !AlreadyExistsError(err) {
    // handle all errors expect "The record already exists"
}
```

Change should be backward compatible as the `Error` function returns string in the same format as before.


jacobbednarz:
You'll want to rebase against `master` seeing how the addition of Cloudflare filters has thrown a bit of a spanner into the works and the default handling of HTTP errors now has a bit more of an edge case. See 3b565f11210000b087b305d0ea08c8f795506f24 for full details.

jakub-bacic:
@jacobbednarz 
I rebased the code and adjusted `ValidateFilterExpression` method accordingly. I simply return the error wrapping it with `errMakeRequestError` message to make it consistent with other API functions. User can extract validation errors later by analyzing returned error struct.

Regarding the comment about errors array, I guess it's made to be generic. I'm able to get two errors for example by calling the following curl:
```
$ curl -X GET "https://api.cloudflare.com/client/v4/zones/zone-id/dns_records" -H "Content-Type:application/json" | jq .

{
  "success": false,
  "errors": [
    {
      "code": 7003,
      "message": "Could not route to /zones/zone-id/dns_records, perhaps your object identifier is invalid?"
    },
    {
      "code": 7000,
      "message": "No route for that URI"
    }
  ],
  "messages": [],
  "result": null
}
```

Not sure if we're able to get two errors from `/filters/validate-expr` endpoint though. I haven't used that one myself.

patryk:
Hey @jakub-bacic, could you please rebase the PR, so we could push this forward? Thanks.

PR #231, opened 2018-10-04
Rename TestClient_Auth to Test_IPs and mock properly

### LINKED ISSUE #127: Mocking in TestClient_Auth is broken hence it ends up making actual HTTP request
This also results in not testing the `Auth` at all.

mihaitodor:
I believe this fixes #127 (cc @terinjokes).

In my view, this test got left behind after some refactoring work. Looking at the rest of the tests, the `X-Auth-Email` and `X-Auth-Token` asserts are validated [here](https://github.com/cloudflare/cloudflare-go/blob/e24ed152b4c8d947daef772b3505c3d2d3ab01bb/cloudflare_test.go#L45-L46), so this test should only be responsible for asserting that `IPs()` does the right thing, which is a special API call, that [does not require logging in to the Cloudflare API](https://github.com/cloudflare/cloudflare-go/blob/e24ed152b4c8d947daef772b3505c3d2d3ab01bb/ips.go#L25).

### DIFF
--- a/ips_test.go
+++ b/ips_test.go
@@ -1,0 +1,67 @@
+package cloudflare
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+type MockTransport struct {
+	http.Transport
+	Server *httptest.Server
+	Path   string
+}
+
+func (m *MockTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	url, err := url.Parse(m.Server.URL + m.Path)
+	if err != nil {
+		return nil, err
+	}
+
+	req.URL = url
+
+	return m.Transport.RoundTrip(req)
+}
+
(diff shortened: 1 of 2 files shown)


ips_test.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	req.URL = url
+
+	return m.Transport.RoundTrip(req)
+}
+
+func TestClient_IPs(t *testing.T) {
```
jacobbednarz: Seeing how `IPs()` isn't anything to do with the `Client` now, we can probably just leave this as `Test_IPs`.

mihaitodor:
@jacobbednarz I was undecided with the name of the test method. Thanks for the suggestion. I fixed it :)

patryk:
Thanks Mihai! Congrats on first contribution. Much appreciated.

mihaitodor:
You're welcome! Thanks for merging it :)

PR #233, opened 2018-10-05
Add support Cloudflare Access

jacobbednarz:
This introduces the initial support for Cloudflare Access via
`access_policy` and `access_application`. It provides full compatibility
with the public API and has been tested against our existing production
Access configuration.

While this provides support for the current API, I'm not entirely happy
with how unreadable some of the Access Policy configuration _can_ be.
For instance, to create a series of allow rules for individual email
addresses, the `AccessPolicy` would be:

```go
AccessPolicy{
  Name: "Allow devs",
  Include: []interface{}{
    AccessPolicyEmail{struct {
      Email string `json:"email"`
    }{Email: "test@example.com"}},
    AccessPolicyEmail{struct {
      Email string `json:"email"`
    }{Email: "test2@example.com"}},
  },
  Decision: "allow",
}
```

Same goes for `AccessPolicyIP` and `AccessPolicyAccessGroup`.
`AccessPolicyEveryone` actually seems pretty reasonable.

```go
AccessPolicy{
  Name: "Allow devs",
  Include: []interface{}{
    AccessPolicyEveryone{},
  },
  Decision: "allow",
}
```

I'd ideally like to expose an API that looked a bit like this:

```go
AccessPolicy{
  Name: "allow devs",
  Include: AccessPolicyConditions{
    IPs: []string{"1.2.3.4/18", "4.3.2.1"},
    Everyone: false,
    Emails: []string{"test@example.com", "test2@example.com"}
    EmailDomains: []string{"test.com"}
  },
}
```

However, every way I tried to slice this ended up with very convoluted
code. Unsure if the lack of :coffee: was to blame or I was overthinking
it so I welcome any thoughts or ideas in this space if anyone has any!

I don't think this should block the merge however revisiting this
particular piece of functionality would be ideal before we attempt
Terraform provider support to make the integration a bit smoother.

Those of you who have used Access before will notice there isn't support
here for Access Groups; which is intentional. While the UI does expose
the endpoints and Cloudflare are pretty consistent with the path +
method combinations, I didn't want to add support for functionality that
isn't publicly documented. Once that functionality gets some
documentation, we can add Access Groups but until then, YMMV.

### API Documentation links

Access applications: https://api.cloudflare.com/#access-applications-properties
Access policies: https://api.cloudflare.com/#access-policy-properties

### DIFF
--- a/access_policy.go
+++ b/access_policy.go
@@ -193,3 +193,21 @@ func (api *API) UpdateAccessPolicy(zoneID, applicationID string, accessPolicy Ac
 	return accessPolicyDetailResponse.Result, nil
 }
 
+// DeleteAccessPolicy deletes an access policy.
+//
+// API reference: https://api.cloudflare.com/#access-policy-update-access-policy
+func (api *API) DeleteAccessPolicy(zoneID, applicationID, accessPolicyID string) error {
+	uri := fmt.Sprintf(
+		"/zones/%s/access/apps/%s/policies/%s",
+		zoneID,
+		applicationID,
+		accessPolicyID,
+	)
+
+	_, err := api.makeRequest("DELETE", uri, nil)
+	if err != nil {
+		return errors.Wrap(err, errMakeRequestError)
+	}
+
+	return nil
+}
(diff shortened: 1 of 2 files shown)


patryk:
Sorry for the delay, merged! Thanks.

PR #236, opened 2018-10-16
Make `ZoneIDByName` safer to use for ambiguous zone names

### LINKED ISSUE #234: ZoneIDByName() assumes zero or one zone per zone name
Function ZoneIDByName() purports to retrieve a zone's ID from its name, but actually returns only the ID of the first matching zone returned by api.[ListZones](https://api.cloudflare.com/#zone-list-zones)().  

Were zone names unique for a given API user, this would not be a problem.  But a user may have multiple zones with the same name created under multiple accounts.  This function ignores any zones with matching names that are listed after the first match.

```go
// ZoneIDByName retrieves a zone's ID from the name.
func (api *API) ZoneIDByName(zoneName string) (string, error) {
    res, err := api.ListZones(zoneName) 
    if err != nil {
        return "", errors.Wrap(err, "ListZones command failed")
    }
    for _, zone := range res {
        if zone.Name == zoneName {
            return zone.ID, nil
        }
    }
    return "", errors.New("Zone could not be found")
}
```

### LINKED ISSUE #235: dnsCreate() creates records for arbitrary zone when multiple zones share same name
Function dnsCreate() will, in the case of an API user with multiple zones of the same name (e.g., one who has created the same zone in multiple associated CloudFlare accounts), create records a single zone, using the first relevant zone returned by [ListZones](https://api.cloudflare.com/#zone-list-zones)().  There is no option to disambiguate by specifying a zone by a unique identifier.

This is caused by dnsCreate()'s use of a buggy ZoneIDByName().  See #234.

```go
func dnsCreate(c *cli.Context) {

    [...]

    zone := c.String("zone")

    [...]

    zoneID, err := api.ZoneIDByName(zone)
```

jacobbednarz:
Updates the `ZoneIDByName` function to be more intuitive (and safe) in
the event that multiple zones are returned by the API. It does this by
ensuring that if we have a response that includes more than a single
zone, an account ID *must be* defined to distinguish which account we
are intending to operate on. This enforcement then feeds into the
following step where the filtering not only matches on `zone.Name` but
the account ID provided as well. Lookups where the zone is unique will
continue to operate as normal.

Included in this change is updating `Account.Settings` to be a pointer
within the struct to enable the `Zone` struct to omit it should it not
be defined.

Fixes #234
Fixes #235

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -90,11 +90,23 @@ func (api *API) ZoneIDByName(zoneName string) (string, error) {
 	if err != nil {
 		return "", errors.Wrap(err, "ListZones command failed")
 	}
+
+	if len(res) > 1 && api.OrganizationID == "" {
+		return "", errors.New("ambiguous zone name used without an account ID")
+	}
+
 	for _, zone := range res {
-		if zone.Name == zoneName {
-			return zone.ID, nil
+		if api.OrganizationID != "" {
+			if zone.Name == zoneName && api.OrganizationID == zone.Account.ID {
+				return zone.ID, nil
+			}
+		} else {
+			if zone.Name == zoneName {
+				return zone.ID, nil
+			}
 		}
 	}
+
 	return "", errors.New("Zone could not be found")
 }
 
--- a/zone.go
+++ b/zone.go
@@ -44,6 +44,7 @@ type Zone struct {
 	Betas       []string `json:"betas"`
 	DeactReason string   `json:"deactivation_reason"`
 	Meta        ZoneMeta `json:"meta"`
+	Account     Account  `json:"account"`
 }
 
 // ZoneMeta describes metadata about a zone.
--- a/accounts.go
+++ b/accounts.go
@@ -15,9 +15,9 @@ type AccountSettings struct {
 
 // Account represents the root object that owns resources.
 type Account struct {
-	ID       string          `json:"id,omitempty"`
-	Name     string          `json:"name,omitempty"`
-	Settings AccountSettings `json:"settings"`
+	ID       string           `json:"id,omitempty"`
+	Name     string           `json:"name,omitempty"`
+	Settings *AccountSettings `json:"settings"`
 }
 
 // AccountResponse represents the response from the accounts endpoint for a
--- a/accounts_test.go
+++ b/accounts_test.go
@@ -12,7 +12,7 @@ import (
 var expectedAccountStruct = Account{
 	ID:   "01a7362d577a6c3019a474fd6f485823",
 	Name: "Cloudflare Demo",
-	Settings: AccountSettings{
+	Settings: &AccountSettings{
 		EnforceTwoFactor: false,
 	},
 }
@@ -132,7 +132,7 @@ func TestUpdateAccount(t *testing.T) {
 	oldAccountDetails := Account{
 		ID:   "01a7362d577a6c3019a474fd6f485823",
 		Name: "Cloudflare Demo - Old",
-		Settings: AccountSettings{
+		Settings: &AccountSettings{
 			EnforceTwoFactor: false,
 		},
 	}
@@ -140,7 +140,7 @@ func TestUpdateAccount(t *testing.T) {
 	newAccountDetails := Account{
 		ID:   "01a7362d577a6c3019a474fd6f485823",
 		Name: "Cloudflare Demo - New",
-		Settings: AccountSettings{
+		Settings: &AccountSettings{
 			EnforceTwoFactor: false,
 		},
 	}
--- a/cloudflare_test.go
+++ b/cloudflare_test.go
@@ -201,3 +201,279 @@ func TestClient_RetryReturnsPersistentErrorResponse(t *testing.T) {
 	_, err := client.ListLoadBalancerPools()
 	assert.Error(t, err)
 }
+
+func TestZoneIDByNameWithNonUniqueZonesWithoutOrgID(t *testing.T) {
+	setup()
+	defer teardown()
+
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, r.Method, "GET", "Expected method 'GET', got %s", r.Method)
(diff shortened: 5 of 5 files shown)


pmocek:
The problem, here, is not that ZoneIDByName() is unsafe or non-intuitive.  The problem, as described in #234, is that it is documented to do something that is not possible: "ZoneIDByName retrieves a zone's ID from the name".

There is not a one-to-one mapping between zone names and zones.  ZoneIDByName() is designed and implemented as if this is the case.  This cannot be fixed.  However, the function's purpose could be redefined to something like, "retrieve a zone's ID from a name and account ID," or "retrieve a set of zone IDs from a name," etc.

jacobbednarz:
> The problem, here, is not that ZoneIDByName() is unsafe or non-intuitive. The problem, as described in #234, is that it is documented to do something that is not possible: "ZoneIDByName retrieves a zone's ID from the name".

I have to partially disagree here. The fact that there are cases where you can request an ID based on the zone name and you don't get back the correct or expected result suggests this is both unsafe **and** non-intuitive. If the library cannot determine which information to return based on the information you've provided, it shouldn't be giving it back. Until we determine whether the multiple active zone setup (your issue example) is something that is intended to work or it's just magically coming together, I don't think we should be changing this function's purpose as it has been functioning as intended for quite some time without any issues. 

I'm happy to consider adding a new function `ZoneIDByNameAndAccountID` which would just set the account ID and extract some of this additional logic if that would help here?

jacobbednarz:
@patryk Do you mind confirming if it's intended that a zone can be available on two different, active accounts?

pmocek:
Use case that reveals the bug described in #235:  Desiring to duplicate a large set of zones and records in a reliable and repeatable manner, you write a script to do it automatically using flarectl , and you want to test with some dry-runs before using it to make real changes for domains that matter.

Your script examines configuration of your DNS server, queries it for DNS records for all zones for which the server is configured, then creates equivalent zones and records in CloudFlare.

Configure it to use your development/testing CloudFlare account.  Run it.  Query applicable DNS servers to verify process worked as expected.

Reconfigure to use production CloudFlare account.  Run again.  Result should be very similar, with only the CloudFlare-supplied DNS server hostnames varying.  Instead, this second run results in creation of zones in production account as expected, but also in faulty creation of DNS records.  DNS record creation is either suppressed or unreliable, because due to #234, flarectrl issues requests to create each record in the first zone found with a matching name, and in cases where that match is the zone created during the test run, the record already exists, so no equivalent record is created in the newly-created zone in the production account.

pmocek:
Again:  This is not about the safety or intuitive nature of the current implementation of ZoneIDByName().  It is about flawed logic.  Querying CloudFlare for a set of zones using a non-unique identifier, then taking action on the first match while ignoring any other matches, *would* make sense if the desired action was, "Take action for a domain named example.com--just any one of them, not any particular one, and not all of them."  Order of results returned by the [list zones API call](https://api.cloudflare.com/#zone-list-zones) is undefined unless the optional `order` parameter is provided, and ZoneIDByName() does not provide it, so flarectl's action in this case is also undefined.

cloudflare.go
```diff
@@ -90,11 +90,23 @@ func (api *API) ZoneIDByName(zoneName string) (string, error) {
...
 	}
+
+	if len(res) > 1 && api.OrganizationID == "" {
+		return "", errors.New("ambiguous zone name used without an account ID")
+	}
+
 	for _, zone := range res {
-		if zone.Name == zoneName {
-			return zone.ID, nil
+		if api.OrganizationID != "" {
```
pmocek: Is the OrganizationID attribute of an API object ever set?  When we instantiate an API in cmd/flarectl/flarectl.go line 31, we provide only two parameters: API key and API e-mail address.  An organization ID will never be used in this case.

```go
api, err = cloudflare.New(os.Getenv("CF_API_KEY"), os.Getenv("CF_API_EMAIL"))
```

Further, multiple matches for a zone lookup by zone name is indication of a problem regardless of whether or not our API client has an OrganizationID attribute.

zone.go
```diff
@@ -44,6 +44,7 @@ type Zone struct {
 	Betas       []string `json:"betas"`
 	DeactReason string   `json:"deactivation_reason"`
 	Meta        ZoneMeta `json:"meta"`
+	Account     Account  `json:"account"`
```
pmocek: [In the CloudFlare API, a zone](https://api.cloudflare.com/#zone-properties)'s account is comprised of two strings: an ID and a name.  If we are going to redefine the Zone struct to include an Account element, it would seem wise to define it as a two-element struct, either defined inline (as with Host, which has Name and Website), or as a new Account type (as with Owner, which has ID, Email, and OwnerType).

cloudflare.go
patryk: Yes, `OrganizationID` is used in https://github.com/terraform-providers/terraform-provider-cloudflare, it may also be used by other third-party projects as SDK.

Possibility of returning same zone twice is because our state machine allows same zone name be signed up to many accounts. They will be separate instances (different zone IDs), but only one of them will be active at a time.

This is allowed to enable zone migration between accounts without need for downtime.

zone.go
patryk: I think we can leave it as is. It's technically the same object (`/account == /zones's "account" field`), just returning a different subset of data.

pmocek: @patryk: Apologies.  I misread.  The Account type is already defined elsewhere with name and ID, and this change adds an Account of that type to the Zone type, which was the second of my proposed alternatives.

pmocek:
An example of the cascading failure I described is flarectl, which does, in fact, create DNS records in one matching zone among multiple, apparently chosen at random, because as described in #235, flarectl relies on this function, ZoneIDByName(), to return *the ID* of *the domain* that matches a given name.  This function, however, simply returns the first match *and ignores any other matches*.  Flarectl's creation of DNS records works, coincidentally, when there is only one zone with a given name.

patryk:
Yes, but this is something we should address in flarectl.

PR #237, opened 2018-10-18
Add NewWithUserServiceKey API

coderanger:
This is a bit nicer than having to pass in intentionally fake creds to `New` and then modifying the API object.

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -42,20 +42,13 @@ type API struct {
 	logger            Logger
 }
 
-// New creates a new Cloudflare v4 API client.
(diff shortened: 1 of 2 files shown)


coderanger:
Looks like the test failure is a Travis blip, the worker never started. Probably just needs someone to force a rebuild.

coderanger:
Forcing a rebuild fixed the Travis blip, but now it looks like golint is failing to install on 1.8.

patryk:
Thanks for the PR. I have removed 1.8 as a test target. I believe it's old and unmaintained already, so no need to require it anymore.

PR #240, opened 2018-10-21
Add support for managing Custom Pages

jacobbednarz:
This introduces the ability to view and update custom pages for the zone
and account level. There are a couple of decisions I've made in this PR
that I'd like some thoughts on:

- Renamed `cpage.go` => `custom_pages.go`. I'm not sure of the initial
  reasoning here but it wasn't immediately clear what `cpage.go` was
  responsible for until I started digging. I swapped this out to make
  it explicit.
- Shared methods for zone and account interactions that accepts a
  `CustomPageParameter` to determine which level to operate on. Instead
  of duplicating each method for each level, I opted for a single,
  shared method that used a struct to determine where the API calls are
  to be made. This makes sense to me and looking forward to other
  integrations (like Terraform), I can see this working. Open to other
  ideas here though!

Zone API docs:

- https://api.cloudflare.com/#custom-pages-for-a-zone-list-available-custom-pages
- https://api.cloudflare.com/#custom-pages-for-a-zone-custom-page-details
- https://api.cloudflare.com/#custom-pages-for-a-zone-update-custom-page-url

Account API docs:

- https://api.cloudflare.com/#custom-pages-account--list-custom-pages
- https://api.cloudflare.com/#custom-pages-account--custom-page-details
- https://api.cloudflare.com/#custom-pages-account--update-custom-page


custom_pages.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	// Should the account ID be defined, treat this as an account level operation.
+	if options.AccountID != "" {
+		pageType = "accounts"
+		identifier = options.AccountID
+	} else {
+		pageType = "zones"
+		identifier = options.ZoneID
+	}
```
petervandoros: Did you consider having `pageType` and `identifier` in `CustomPageOptions` directly? I think it will simplify this code that is repeated a few times.

patryk:
Yeah, your reasoning in both respects is sound. File name should not be a concern regarding how Go resolves names within a package. Shared methods avoid boilerplate and provided they are just added pose no compatibility issues. Thanks!

PR #242, opened 2018-10-22
[cmd/flarectl] Handle errors from `CreateZone`

### LINKED ISSUE #241: flarectl zone create doesn't work with .club domains
flarectl.exe for Windows x64 does not create a new zone when the domain name has a .club extension.

There also is no error message.

jacobbednarz:
Should the `CreateZone` calls fail, it should provide some feedback to
the user as to why. This implements a fix whereby it returns the HTTP
response to stderr which includes the error message.

Fixes #241


patryk:
Thanks!

PR #245, opened 2018-10-31
Add support for audit log apis to cloudflare-go

ejcx:


auditlogs.go
```diff
@@ -0,0 +1,139 @@
...
+	if a.Page > 0 {
+		params += "&page=" + fmt.Sprintf("%d", a.Page)
+	}
+	return params
+}
+
+// GetOrganizationAuditLogs will return the audit logs of a specific
+// organization, based on the ID passed in. The audit logs can be
+// filtered based on any argument in the AuditLogFilter
+func (api *API) GetOrganizationAuditLogs(organizationID string, a AuditLogFilter) (AuditLogResponse, error) {
```
jacobbednarz: Is it possible to add the API reference in here? Similar to https://github.com/cloudflare/cloudflare-go/blob/aa07f8c57bcb06db0b1d4dfe7b685140826822c6/accounts.go#L46-L48?

auditlogs.go
```diff
@@ -0,0 +1,139 @@
...
+		return auditResponse, err
+	}
+	return auditResponse, nil
+}
+
+// GetUserAuditLogs will return your user's audit logs. The audit logs can be
+// filtered based on any argument in the AuditLogFilter
+func (api *API) GetUserAuditLogs(a AuditLogFilter) (AuditLogResponse, error) {
+	uri := "/user/audit_logs" + fmt.Sprintf("%s", a)
+	res, err := api.makeRequest("GET", uri, nil)
```
jacobbednarz: Is it possible to add the API reference in here? Similar to https://github.com/cloudflare/cloudflare-go/blob/aa07f8c57bcb06db0b1d4dfe7b685140826822c6/accounts.go#L46-L48?

jacobbednarz:
Thanks @ejcx! Do we also want to include some tests for these new endpoints?

auditlogs.go
ejcx: Good call.

ejcx:
@jacobbednarz I actually don't know what these tests are good for, to be quite honest. It looks like it's just testing that the JSON responses can be marshalled properly? I don't really understand the mocking happening and why it's there

jacobbednarz:
That's fair and due to the simplicity here, I don't think it's a big deal if we were to leave it out. My question was mainly focused on the `String()` function and whether we needed some tests there to confirm particular behavior for the endpoints. I.e. if `a.ActorEmail` is empty but that still ends up appended to the URL, does it break functionality?

This was a minor niggle and I'm happy either way 😄

ejcx:
@jacobbednarz I added a test for stringifying.

ejcx:
Thanks @patryk, I updated the names.

PR #246, opened 2018-11-01
Vendored dependancies using `go mod`

tresni:
Wanted to build a Homebrew Formulae but they require vendored dependancies for golang now.  This setups up those depedancies and allows an easy Homebrew Formulae.


jacobbednarz:
Thanks for the PR @tresni!

I think this is fine as majority of the dependencies are only for `flarectl`. I'm a little concerned `go mod` is [still experimental](https://github.com/golang/go/wiki/Modules) and this could introduce some vendoring weirdness until it becomes stable in an upcoming release but I'm willing to give it a go.

There are a couple of things we'll need to address to make sure CI is running using vendoring:

- Enabling modules support. Right now, it's not using modules (see https://travis-ci.org/cloudflare/cloudflare-go/jobs/449191157#L483). We can fix this by defining a new `env` key/value for `GO111MODULE=on`. I.e.

```yaml
language: go
go: 
 - "1.10.x"
 - "1.11.x"

env:
  - GO111MODULE=on
```

- The script runner will need to be updated as we're currently using [`go test -v -race ./...`](https://github.com/cloudflare/cloudflare-go/blob/master/.travis.yml#L26) which doesn't use the vendor directory. To do this, we need to add `-mod=vendor` to our script _but_ we can only apply this to go 1.11.x since it's not supported in other versions. I'd propose we provide an additional `env` value here that is defined which can be injected into the command.

```yaml
language: go

matrix:
  include:
  - go: "1.10.x" # and all the other versions we test but don't have `mod` support
    env: 
      - TEST_ARGS="-v -race"
  - go: "1.11.x"
    env: 
      - TEST_ARGS="-v -race -mod=vendor"

script:
  # .. other steps
  - go test $TEST_ARGS ./...
```

Let me know if you need a hand with the Travis CI stuff.

tresni:
@jacobbednarz Alright, I think this is all working now.  Took a little more then I expected, but please let me know if that doesn't seem to be doing what you expect.

`go get` should be a no-op (or completely unnecessary) for anything [1.5 or later](https://blog.gopheracademy.com/advent-2015/vendor-folder/), but I've left it in case you think I'm missing something.  Did need some additional uses of `-mod=vendor` during `go test`, `go list`, and `go vet` to prevent `go` from trying to download packages we don't want it to.  I think it's good to go now.

jacobbednarz:
Really nice job on getting this working! 👏 Couple of questions:

- It looks like the [`before_install`](https://github.com/cloudflare/cloudflare-go/blob/8790350e2ee0f7e94540ec58d44f9e44e6e7848d/.travis.yml#L29-L31) isn't using vendored libraries. I think this is 👌 as they are CI only libraries and aren't required for the `cloudflare-go` library to operate. Is this your understanding too?
- The `go vet` command on [1.11](https://travis-ci.org/cloudflare/cloudflare-go/jobs/450123336#L516) is 3 times slower than [1.9](https://travis-ci.org/cloudflare/cloudflare-go/jobs/450123334#L494). Looking at the CI timings, this step can be wildly varied. I don't think it's a blocker but just wanted to call it out.

tresni:
* Yeah, I took the `before_install` piece to be build requirements vs actual requirements for the tool.  e.g. I have successfully built without using `gox`, it just parallelizes multiple builds.  Same with `golint`, it's being used to test builds, it's not actually a dependency of flarectl or cloudflare-go.
* `go vet` seems to vary a lot.  If you look over the past builds for this branch, that's where a lot (but not all) of the variance was.  The other was in `go get`, which shouldn't be used when using module support.  I've checked each piece of the `go vet`/`go list` command, it doesn't appear to be grabbing anything it shouldn't.

jacobbednarz:
ping @patryk for your thoughts 💭

jacobbednarz:
TIL, thanks @deuill!

patryk:
Great work @tresni, I love it. Thanks @deuill for some insights Re: go vet, for now the slowdown doesn't hurt us (it's just few secs).

PR #259, opened 2018-11-29
Add support for using custom ports with monitors

Chreo:
This PR add support for creating monitors testing against customer ports which is a requirement for making basic layer 4 TCP connection monitors


jacobbednarz:
Closes #262

patryk:
Thanks!

PR #261, opened 2018-11-30
Add session_affinity_ttl option to the libs.

georgethomas111:
## Description

This change enables using PersistenceTTL to specify the time in seconds the cookie is going to stay alive. This will provide the user the flexibility to set the cookie expiry time if the choose to.

Yes. Unit test added. 

## Types of changes

### DIFF
--- a/load_balancing.go
+++ b/load_balancing.go
@@ -54,18 +54,19 @@ type LoadBalancerMonitor struct {
 
 // LoadBalancer represents a load balancer's properties.
 type LoadBalancer struct {
-	ID           string              `json:"id,omitempty"`
-	CreatedOn    *time.Time          `json:"created_on,omitempty"`
-	ModifiedOn   *time.Time          `json:"modified_on,omitempty"`
-	Description  string              `json:"description"`
-	Name         string              `json:"name"`
-	TTL          int                 `json:"ttl,omitempty"`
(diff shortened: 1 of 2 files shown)


jacobbednarz:
Have you got some links to the API endpoint or documentation that supports this change? I don't see anything about a session affinity TTL in the API, docs or UI itself to suggest this is a thing.

georgethomas111:
The api documentation and ui hasn't been updated yet. You could expect that to happen within one or two weeks. I made the PR to make sure the option is available here.

jacobbednarz:
Thanks for the additional context. It would be 💯 if we could include that in the initial PR description too so that everyone is on the same page. 

I assume we're holding off merging until this feature is released?

georgethomas111:
I am fine either way. The feature can be used by someone who wants to try it without documentation.

jacobbednarz:
Let's hold off until it's all released to avoid any confusion and we can include a link to the documentation here at the same time.

georgethomas111:
@jacobbednarz The documentation has been updated. Let me know if any changes are needed for merging this. 

https://api.cloudflare.com/#load-balancers-create-load-balancer

patryk:
👊 thanks @georgethomas111 !

PR #263, opened 2018-12-10
Added support for WorkersKV

aplunk:
# Adds support for Workers KV

* https://developers.cloudflare.com/workers/kv/
* https://api.cloudflare.com/#workers-kv-namespace-properties

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -163,19 +163,19 @@ func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, u
 	var reqBody io.Reader
 	var err error
 
-	if params != nil {
-		switch p := params.(type) {
-		case []byte:
-			reqBody = bytes.NewReader(p)
-		case io.Reader:
-			reqBody = p
-		default:
-			jsonBody, err := json.Marshal(p)
-			if err != nil {
-				return nil, errors.Wrap(err, "error marshalling params to JSON")
-			}
-			reqBody = bytes.NewReader(jsonBody)
+	switch p := params.(type) {
+	case []byte:
+		reqBody = bytes.NewReader(p)
+	case io.Reader:
+		reqBody = p
+	case nil:
+		// empty request body
+	default:
+		jsonBody, err := json.Marshal(p)
+		if err != nil {
+			return nil, errors.Wrap(err, "error marshalling params to JSON")
 		}
+		reqBody = bytes.NewReader(jsonBody)
 	}
 
 	var resp *http.Response
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -160,29 +160,30 @@ func (api *API) makeRequestWithAuthType(ctx context.Context, method, uri string,
 
 func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, uri string, params interface{}, authType int, headers http.Header) ([]byte, error) {
 	// Replace nil with a JSON object if needed
-	var reqBody io.Reader
+	var jsonBody []byte
 	var err error
 
-	switch p := params.(type) {
-	case []byte:
-		reqBody = bytes.NewReader(p)
-	case io.Reader:
-		reqBody = p
-	case nil:
-		// empty request body
-	default:
-		jsonBody, err := json.Marshal(p)
-		if err != nil {
-			return nil, errors.Wrap(err, "error marshalling params to JSON")
+	if params != nil {
+		if paramBytes, ok := params.([]byte); ok {
+			jsonBody = paramBytes
+		} else {
+			jsonBody, err = json.Marshal(params)
+			if err != nil {
+				return nil, errors.Wrap(err, "error marshalling params to JSON")
+			}
 		}
-		reqBody = bytes.NewReader(jsonBody)
+	} else {
+		jsonBody = nil
 	}
 
 	var resp *http.Response
 	var respErr error
+	var reqBody io.Reader
 	var respBody []byte
 	for i := 0; i <= api.retryPolicy.MaxRetries; i++ {
-
+		if jsonBody != nil {
+			reqBody = bytes.NewReader(jsonBody)
+		}
 		if i > 0 {
 			// expect the backoff introduced here on errored requests to dominate the effect of rate limiting
 			// dont need a random component here as the rate limiter should do something similar
@@ -196,16 +197,6 @@ func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, u
 			api.logger.Printf("Sleeping %s before retry attempt number %d for request %s %s", sleepDuration.String(), i, method, uri)
 			time.Sleep(sleepDuration)
 
-			// seek to the beginning of the request body if supported because a previous attempt may have
-			// left it in a partially drained state
-			seeker, ok := reqBody.(io.Seeker)
-			if ok {
-				if _, err := seeker.Seek(0, io.SeekStart); err != nil {
-					return nil, errors.Wrap(err, "Error seeking to the beginning of the request body")
-				}
-			} else if reqBody != nil {
-				return nil, errors.New("Request body does not support io.Seek and can't be retried")
-			}
 		}
 		api.rateLimiter.Wait(context.TODO())
 		if err != nil {
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -160,29 +160,28 @@ func (api *API) makeRequestWithAuthType(ctx context.Context, method, uri string,
 
 func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, uri string, params interface{}, authType int, headers http.Header) ([]byte, error) {
 	// Replace nil with a JSON object if needed
-	var jsonBody []byte
+	var reqBody io.Reader
 	var err error
+
 	if params != nil {
-		if paramBytes, ok := params.([]byte); ok {
-			jsonBody = paramBytes
-		} else {
-			jsonBody, err = json.Marshal(params)
+		switch p := params.(type) {
+		case []byte:
+			reqBody = bytes.NewReader(p)
+		case io.Reader:
+			reqBody = p
+		default:
+			jsonBody, err := json.Marshal(p)
 			if err != nil {
 				return nil, errors.Wrap(err, "error marshalling params to JSON")
 			}
+			reqBody = bytes.NewReader(jsonBody)
 		}
-	} else {
-		jsonBody = nil
 	}
 
 	var resp *http.Response
 	var respErr error
-	var reqBody io.Reader
 	var respBody []byte
 	for i := 0; i <= api.retryPolicy.MaxRetries; i++ {
-		if jsonBody != nil {
-			reqBody = bytes.NewReader(jsonBody)
-		}
 
 		if i > 0 {
 			// expect the backoff introduced here on errored requests to dominate the effect of rate limiting
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -195,6 +195,17 @@ func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, u
 			// useful to do some simple logging here, maybe introduce levels later
 			api.logger.Printf("Sleeping %s before retry attempt number %d for request %s %s", sleepDuration.String(), i, method, uri)
 			time.Sleep(sleepDuration)
+
+			// seek to the beginning of the request body if supported because a previous attempt may have
+			// left it in a partially drained state
+			seeker, ok := reqBody.(io.Seeker)
+			if ok {
+				if _, err := seeker.Seek(0, io.SeekStart); err != nil {
+					return nil, errors.Wrap(err, "Error seeking to the beginning of the request body")
+				}
+			} else if reqBody != nil {
+				return nil, errors.New("Request body does not support io.Seek and can't be retried")
+			}
 		}
 		api.rateLimiter.Wait(context.TODO())
 		if err != nil {
--- a/workers_kv.go
+++ b/workers_kv.go
@@ -121,13 +121,13 @@ func (api *API) UpdateWorkersKVNamespace(ctx context.Context, namespaceID string
 	return result, err
 }
 
-// CreateWorkersKV writes a value identified by a key.
+// WriteWorkersKV writes a value identified by a key.
 //
 // API reference: https://api.cloudflare.com/#workers-kv-namespace-write-key-value-pair
 func (api *API) WriteWorkersKV(ctx context.Context, namespaceID, key string, value []byte) (Response, error) {
 	uri := fmt.Sprintf("/accounts/%s/storage/kv/namespaces/%s/values/%s", api.OrganizationID, namespaceID, key)
 	res, err := api.makeRequestWithAuthTypeAndHeaders(
-		ctx, http.MethodPut, uri, value, api.authType, http.Header{"Content-Type": []string{"binary/octet-stream"}},
+		ctx, http.MethodPut, uri, value, api.authType, http.Header{"Content-Type": []string{"application/octet-stream"}},
 	)
 	if err != nil {
 		return Response{}, errors.Wrap(err, errMakeRequestError)
--- a/workers_kv.go
+++ b/workers_kv.go
@@ -124,7 +124,7 @@ func (api *API) UpdateWorkersKVNamespace(ctx context.Context, namespaceID string
 // CreateWorkersKV writes a value identified by a key.
 //
 // API reference: https://api.cloudflare.com/#workers-kv-namespace-write-key-value-pair
-func (api *API) CreateWorkersKV(ctx context.Context, namespaceID, key string, value []byte) (Response, error) {
+func (api *API) WriteWorkersKV(ctx context.Context, namespaceID, key string, value []byte) (Response, error) {
 	uri := fmt.Sprintf("/accounts/%s/storage/kv/namespaces/%s/values/%s", api.OrganizationID, namespaceID, key)
 	res, err := api.makeRequestWithAuthTypeAndHeaders(
 		ctx, http.MethodPut, uri, value, api.authType, http.Header{"Content-Type": []string{"binary/octet-stream"}},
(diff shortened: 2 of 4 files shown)


cloudflare.go
```diff
@@ -160,29 +160,28 @@ func (api *API) makeRequestWithAuthType(ctx context.Context, method, uri string,
...
-		jsonBody = nil
+		reqBody = bytes.NewReader(jsonBody)
 	}
 
 	var resp *http.Response
 	var respErr error
-	var reqBody io.Reader
 	var respBody []byte
 	for i := 0; i <= api.retryPolicy.MaxRetries; i++ {
-		if jsonBody != nil {
```
aplunk: If reqBody is an `io.Seeker` we may need to call [Seek](https://golang.org/src/io/io.go?s=4702:4774#L105) because the Reader could have been partially drained.

jacobbednarz:
Appreciate the KV support! I've got a few questions related to this PR that we don't really touch on anywhere and I think are important for the reviewers to understand the intentions behind some of the changes.

- I can only see internal functions where we rely on `context`. Is there any reason why we're doing this for public/exported functions now? What benefits are we looking to gain from it that we currently need?
- Similarly for the `io.Reader` change, I can't think of any (current) use cases where this would warrant the change. Do you have some ideas/plans for it that you could elaborate on here to better understand why we'd want this change now? It seems like extra complexity as it stands.

Small cosmetic thing, as we're exporting most of these structs and functions, we probably want to include `Worker` somewhere in the name as right now it's not clear what "storage" or "namespace" it's referring to. I.e. `StorageNamespaceRequest` => `WorkerKVStorageNamespaceRequest`, `CreateStorageNamespace` => `CreateWorkerKVStorageNamespace`

Thanks!

workers_kv.go
```diff
@@ -0,0 +1,189 @@
...
+		return result, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return result, err
+}
+
+// DeleteStorageNamespace deletes the namespace corresponding to the given ID
+//
+// API reference: https://api.cloudflare.com/#workers-kv-namespace-remove-a-namespace
+func (api *API) DeleteStorageNamespace(ctx context.Context, namespace string) (Response, error) {
```
ags799: can this just return `error`?

workers_kv.go
```diff
@@ -0,0 +1,189 @@
...
+		return result, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return result, err
+}
+
+// UpdateStorageNamespace modifies a namespace's title
+//
+// API reference: https://api.cloudflare.com/#workers-kv-namespace-rename-a-namespace
+func (api *API) UpdateStorageNamespace(ctx context.Context, namespace string, req *StorageNamespaceRequest) (Response, error) {
```
ags799: can this just return `error`?

ags799: `namespace` -> `namespaceID`

workers_kv.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	return result, err
+}
+
+// CreateStorageKV writes a value identified by a key. Body should be the value to be stored.
+// Existing values will be overwritten.
+//
+// API reference: https://api.cloudflare.com/#workers-kv-namespace-write-key-value-pair
+func (api *API) CreateStorageKV(ctx context.Context, namespace, key string, value io.Reader) (Response, error) {
```
ags799: `CreateStorageKV` -> `Put`?

workers_kv.go
```diff
@@ -0,0 +1,189 @@
...
+
+	result := Response{}
+	if err := json.Unmarshal(res, &result); err != nil {
+		return result, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return result, err
+}
+
+// CreateStorageKV writes a value identified by a key. Body should be the value to be stored.
```
ags799: You can remove the sentence `Body should be the value to be stored`, since this is a CLI tool.

workers_kv.go
```diff
@@ -0,0 +1,189 @@
...
+		return result, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return result, err
+}
+
+// ReadStorageKV returns the value associated with the given key in the given namespace
+//
+// API reference: https://api.cloudflare.com/#workers-kv-namespace-read-key-value-pair
+func (api API) ReadStorageKV(ctx context.Context, namespace, key string) ([]byte, error) {
```
ags799: rename to `Get`?

workers_kv.go
ags799: return just `error`?

cloudflare.go
```diff
@@ -160,29 +160,28 @@ func (api *API) makeRequestWithAuthType(ctx context.Context, method, uri string,
...
+	case []byte:
+		reqBody = bytes.NewReader(p)
+	case io.Reader:
+		reqBody = p
+	case nil:
+		// empty request body
+	default:
+		jsonBody, err := json.Marshal(p)
+		if err != nil {
+			return nil, errors.Wrap(err, "error marshalling params to JSON")
```
ags799: do we have a `errMarshalError`?

cloudflare.go
ags799: Can we always expect a `reqBody` to be an `io.Seeker`? What if it isn't?

Is there some way we can make a copy of the `io.Reader` for each attempt? We wouldn't copy the backing value, just the pointer to the start of that value.

workers_kv_example_test.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+func ExampleAPI_CreateStorageNamespace() {
+	api, err := cloudflare.New(apiKey, user, cloudflare.UsingOrganization(org))
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	req := &cloudflare.StorageNamespaceRequest{Title: "test_namespace2"}
+	snr, err := api.CreateStorageNamespace(context.Background(), req)
```
ags799: `snr` is very terse, maybe `response`?

workers_kv_example_test.go
```diff
@@ -0,0 +1,137 @@
...
+	key := "test_key"
+
+	resp, err := api.CreateStorageKV(context.Background(), namespace, key, payload)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	if payload.Len() > 0 {
+		log.Fatalf("Reader was not drained, remaining bytes: %d", payload.Len())
+	}
```
ags799: is this necessary? shouldn't this be reported in `CreateStorageKV`'s error?

workers_kv_example_test.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func ExampleAPI_DeleteStorageKV() {
+	api, err := cloudflare.New(apiKey, user, cloudflare.UsingOrganization(org))
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	key := "test_key"
+	resp, err := api.ReadStorageKV(context.Background(), namespace, key)
```
ags799: `DeleteStorageKV`

workers_kv_test.go
```diff
@@ -0,0 +1,369 @@
...
+	res, err := client.CreateStorageNamespace(context.Background(), &StorageNamespaceRequest{Title: "Namespace"})
+	want := StorageNamespaceResponse{
+		successResponse,
+		"3aeaxxxxee014exxxx4cf66xxxxc0448",
+		"test_namespace",
+	}
+
+	if assert.NoError(t, err) {
+		assert.Equal(t, want.Response, res.Response)
+	}
```
ags799:
```suggestion
	require.NoError(t, err)
	require.Equal(t, want.Response, res.Response)
```

workers_kv_test.go
```diff
@@ -0,0 +1,369 @@
...
+		assert.Equal(t, want.ResultInfo, res.ResultInfo)
+
+		sort.Slice(res.Result, func(i, j int) bool {
+			return res.Result[i].ID < res.Result[j].ID
+		})
+		sort.Slice(want.Result, func(i, j int) bool {
+			return want.Result[i].ID < want.Result[j].ID
+		})
+		assert.Equal(t, res.Result, want.Result)
+	}
```
ags799: since you're controlling the order of output, can't you just `require.Equal(t, want, res)`?

workers_kv_test.go
```diff
@@ -0,0 +1,369 @@
...
+		w.Header().Set("content-type", "binary/octet-stream")
+		fmt.Fprintf(w, response)
+	})
+
+	res, err := client.CreateStorageKV(context.Background(), namespace, key, bytes.NewReader(value))
+	want := successResponse
+
+	if assert.NoError(t, err) {
+		assert.Equal(t, want, res)
+	}
```
ags799: i recommend moving `want` definition above the `CreateStorageKV` call. should make reading easier, to be able to see error checking right after call.

workers_kv_test.go
```diff
@@ -0,0 +1,369 @@
...
+			try++
+		}
+
+		w.Header().Set("content-type", "binary/octet-stream")
+		fmt.Fprintf(w, response)
+	})
+
+	reader := bytes.NewReader(value)
+	totalLen := reader.Len()
+	assert.True(t, totalLen > 0)
```
ags799: this check shouldn't be necessary, you've defined `value` to be non-empty

workers_kv_test.go
```diff
@@ -0,0 +1,369 @@
...
+
+	res, err := client.CreateStorageKV(context.Background(), namespace, key, reader)
+
+	if assert.NoError(t, err) {
+		assert.Equal(t, successResponse, res)
+	}
```
ags799: i don't understand this, shouldn't you get an error?

workers_kv_test.go
```diff
@@ -0,0 +1,369 @@
...
+	key := "test_key"
+	namespace := "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
+
+	mux.HandleFunc(fmt.Sprintf("/accounts/foo/storage/kv/namespaces/%s/values/%s", namespace, key), func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusInternalServerError)
+	})
+
+	_, err := client.CreateStorageKV(context.Background(), namespace, key, noopReader{})
+
+	assert.Error(t, err, "Request body does not support io.Seek and can't be retried")
```
ags799: also confused here

workers_kv_test.go
```diff
@@ -0,0 +1,369 @@
...
+
+		sort.Slice(res.Result, func(i, j int) bool {
+			return res.Result[i].Name < res.Result[j].Name
+		})
+
+		sort.Slice(want.Result, func(i, j int) bool {
+			return want.Result[i].Name < want.Result[j].Name
+		})
+		assert.Equal(t, want.Result, res.Result)
+	}
```
ags799: i also think you can use a plain `require.Equal(t, want, res)` here

cloudflare.go
aplunk: It will not always be an io.Seeker. We can't copy the `io.Reader` because the whole point is to stream data instead of allocating a buffer up front. If customers want a buffer allocated up front, they can pass in a `[]byte` instead. If they use [strings.NewReader()](https://golang.org/pkg/strings/#NewReader) or [bytes.NewReader()](https://golang.org/pkg/bytes/#NewReader) they get `Seek` implemented for free. The purpose of the io.Seeker interface is to provide a way to reset the "pointer" to the start of the reader.

workers_kv.go
```diff
@@ -0,0 +1,189 @@
...
+
+// StorageNamespaceRequest provides parameters for creating and updating storage namespaces
+type StorageNamespaceRequest struct {
+	Title string
+}
+
+// StorageNamespaceResponse is the response received when creating storage namespaces
+type StorageNamespaceResponse struct {
+	Response
+	ID, Title string
```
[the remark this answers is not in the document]
aplunk: I did test this manually using the example tests... let me verify to see if my test was invalid.

workers_kv.go
```diff
@@ -0,0 +1,189 @@
...
+
+	result := StorageNamespaceResponse{}
+	if err := json.Unmarshal(res, &result); err != nil {
+		return result, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return result, err
+}
+
+// ListStorageNamespaces lists storage namespaces
```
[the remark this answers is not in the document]
aplunk: I was following function comment patterns introduced in workers.go. Should I break the pattern?

cloudflare.go
ags799: can we require the parameter to be an `io.Seeker`? would like to avoid branches in logic

workers_kv.go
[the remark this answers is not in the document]
ags799: Nope never mind!

cloudflare.go
aplunk: We could but that prevents users from streaming data who don't want to buffer, which is a valid use case IMO. This logic is only necessary because retry logic was added inside of this rest client... I would prefer to either conditionally support io.Seeker, or to always throw an error if this library attempts to retry with an io.Reader payload. The code currently implemented gives the caller the flexibility to pass in a buffer or a stream and handles retrying gracefully when possible.

workers_kv_test.go
aplunk: Will fix.

workers_kv_test.go
aplunk: It only errors the first time. The retry logic in `makeRequestWithAuthTypeAndHeaders` calls the mock handler a second time according to the retry policy which succeeds.

cloudflare.go
aplunk: I'm not sure, checking.

workers_kv.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+// CreateStorageNamespace creates a namespace under the given title.
+// A 400 is returned if the account already owns a namespace with this title.
+// A namespace must be explicitly deleted to be replaced.
+//
+// API reference: https://api.cloudflare.com/#workers-kv-namespace-create-a-namespace
+func (api *API) CreateStorageNamespace(ctx context.Context, req *StorageNamespaceRequest) (StorageNamespaceResponse, error) {
+	uri := fmt.Sprintf("/accounts/%s/storage/kv/namespaces", api.OrganizationID)
+	res, err := api.makeRequestContext(ctx, "POST", uri, req)
```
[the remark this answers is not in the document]
aplunk: Updating.

workers_kv.go
aplunk: If you would prefer. I'll update this now.

workers_kv_example_test.go
aplunk: This is an example test. We don't have to check anything here it was just for debugging. I can remove.

aplunk: >  shouldn't this be reported in CreateStorageKV's error?

I don't think so because that's the underlying http client's responsibility. I'm open to suggestions if you disagree.

workers_kv_example_test.go
aplunk: Fixed.

workers_kv_test.go
aplunk: controlling the order of output? Are you asking about sorting? I'm doing that because json.Unmarshal is not guaranteed to maintain list order.

cloudflare.go
aplunk: Not that I can find. This is a direct copy of the error that this library already returned.

workers_kv.go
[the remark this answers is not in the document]
aplunk: Fixed.

workers_kv.go
aplunk: Do you want everything that returns (Response, error) to return error instead? Or just certain functions?

workers_kv.go
aplunk: Updating.

workers_kv.go
aplunk: I don't understand the question.

aplunk: Are you asking me to rename to `PutStorageKV`?

workers_kv_example_test.go
aplunk: Will update.

workers_kv_test.go
aplunk: I don't see a require dependency installed in this repository. Are you asking me to install something?

aplunk: Ahh it's part of testify. I understand now. Updating.

workers_kv_test.go
aplunk: I can use require instead.

cloudflare.go
aplunk: Because large values will be supported in the future, we should fail retrying here. Remove io.Seeker support.

workers_kv.go
aplunk: Discussed IRL, we are going to return a response here.

cloudflare.go
aplunk: @ags799 the error I would have to throw in this case must happen at runtime, even if we remove the `io.Seeker` support. Because of this, I don't think the error will be less confusing by removing `io.Seeker` support. I think we should keep this in because I think that most customers are going to pass in `bytes.NewReader()` or `strings.NewReader()` which already support `io.Seeker`... giving them retry support for free.

aplunk: Switching back to `[]byte` because this API likely won't need to support large files.

workers_kv.go
```diff
@@ -0,0 +1,188 @@
...
+	return result, err
+}
+
+// CreateWorkersKV writes a value identified by a key.
+//
+// API reference: https://api.cloudflare.com/#workers-kv-namespace-write-key-value-pair
+func (api *API) CreateWorkersKV(ctx context.Context, namespaceID, key string, value []byte) (Response, error) {
+	uri := fmt.Sprintf("/accounts/%s/storage/kv/namespaces/%s/values/%s", api.OrganizationID, namespaceID, key)
+	res, err := api.makeRequestWithAuthTypeAndHeaders(
+		ctx, http.MethodPut, uri, value, api.authType, http.Header{"Content-Type": []string{"binary/octet-stream"}},
```
aplunk: @ags799 is `binary/octet-stream` correct? Should this be `test/plain` in certain situations?

ags799: `application/octet-stream` please

aplunk: Fixed.

ags799:
I'm a developer for Cloudflare KV Storage and I approve this pull request.

PR #266, opened 2018-12-18
Added Priority field to ZoneLockdown type

mbrowne:
## Description

This adds the Priority field for ZoneLockdown. It's currently undocumented but it seems to be already functional for both reading and writing.

Yes, I ran the code in lockdown_example_test.go against a real Cloudflare account and it worked.

## Types of changes

## Checklist:

Note: I am under the impression that the documentation is automatically generated in this case so that's why I marked yes for "I have updated the documentation accordingly"

### DIFF
--- a/lockdown_example_test.go
+++ b/lockdown_example_test.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"log"
 	"strings"
-	"testing"
 
 	cloudflare "github.com/cloudflare/cloudflare-go"
 )
@@ -31,7 +30,7 @@ func ExampleAPI_ListZoneLockdowns_all() {
 	}
 }
 
-func ExampleAPI_CreateZoneLockdown(t *testing.T) {
+func ExampleAPI_CreateZoneLockdown() {
 	api, err := cloudflare.New("deadbeef", "test@example.org")
 	if err != nil {
 		log.Fatal(err)
--- a/lockdown_example_test.go
+++ b/lockdown_example_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"strings"
+	"testing"
 
 	cloudflare "github.com/cloudflare/cloudflare-go"
 )
@@ -28,4 +29,37 @@ func ExampleAPI_ListZoneLockdowns_all() {
 	for _, r := range rules.Result {
 		fmt.Printf("%s: %s\n", strings.Join(r.URLs, ", "), r.Configurations)
 	}
+}
+
+func ExampleAPI_CreateZoneLockdown(t *testing.T) {
+	api, err := cloudflare.New("deadbeef", "test@example.org")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	zoneID, err := api.ZoneIDByName("example.org")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	newZoneLockdown := cloudflare.ZoneLockdown{
+		Description: "Test Zone Lockdown Rule",
+		URLs: []string{
+			"*.example.org/test",
+		},
+		Configurations: []cloudflare.ZoneLockdownConfig{
+			cloudflare.ZoneLockdownConfig{
+				Target: "ip",
+				Value:  "127.0.0.1",
+			},
+		},
+		Paused:   false,
+		Priority: 1,
+	}
+
+	response, err := api.CreateZoneLockdown(zoneID, newZoneLockdown)
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Println("Response: ", response)
 }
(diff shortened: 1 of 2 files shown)


jacobbednarz:
:wave: @mbrowne - thanks for the effort in getting this change together!

While I too can see this in the UI, I don't think we want to make a habit of introducing undocumented functionality in case it hasn't yet been intended for public API usage. Instead, do you mind raising a support ticket with Cloudflare (via http://support.cloudflare.com/) and they can pass through the information to the appropriate team? If not, please let me know and I'll raise one.

Once we have an answer on that, we can proceed with adding this one.

mbrowne:
@jacobbednarz Given that this repo is owned by Cloudflare, I figured one of the developers here would know the status of this field. But I did also submit a support ticket yesterday. This is the relevant part of the response I received so far:

> Yes, we have escalated this issue to our documentation team to check further.
>
> Meanwhile, can you please try to add this field  "priority":1 in the API call and see if sets the priority while creating zone lockdown rules and let us know is it giving any error?
>
> example:
>
```bash
curl -X POST "https://api.cloudflare.com/client/v4/zones/023e105f4ecef8ad9ca31a8372d0c353/firewall/lockdowns" \
     -H "X-Auth-Email: user@example.com" \
     -H "X-Auth-Key: c2547eb745079dac9320b638f5e225cf483cc5cfdda41" \
     -H "Content-Type: application/json" \
     --data '{"paused":false,"priority":1,"description":"Restrict access to these endpoints to requests from a known IP address","urls":["api.mysite.com/some/endpoint*"],"configurations":[{"target":"ip","value":"198.51.100.4"}]}'
```

So apparently they are telling customers it's OK to use this, but we still need an official answer. So feel free to keep this PR on hold until I hear back from the "documentation team". But it seems likely this was just an oversight and not deliberately left out of the documentation.

mbrowne:
I received a new response from Cloudflare:

> As far as I'm concerned the "priority" parameter is meant to be documented. While you create a zone lockdown rule with priority in the dashboard, if you open the developer tools of browser, you will see the corresponding POST request with payload, which includes "priority" field as well. Hope this helps verify the API call you want to make.
>
> We have an internal ticket to follow up the documentation update and we will keep you posted once we have updates.
>
> Thanks for your patience.

So I think it would be fine to go ahead and merge this, but it's your call of course. Thanks :)

garrettgalow:
Priority is considered an 'advanced' feature and doesn't show up in the UI unless configured by API. Once set by the API it actually will populate in the UI. It is an officially supported field that should be included in here.

Here are some official docs about priority: https://developers.cloudflare.com/firewall/cf-firewall-rules/order-priority/

jacobbednarz:
@garrettgalow Have you mixed up issues here? 😛 This issue is about Zone Lockdown whereas your link is for Filter based firewall. Or have I missed something here?

mbrowne:
Looks like it's now documented:
https://api.cloudflare.com/#zone-lockdown-properties

Can this be merged now (when you have a moment)?

gofish:
@mbrowne have you tested this with a priority of 0? i.e. the zero value of this new field. I checked some code internally and it looks like the "priority" should either be null, missing, or in the range [1,999999]. If correct, then this change would break clients that do not provide a value for "priority".

Either setting `omitempty` in the json struct tag, or using a `*int`, would likely work. I don't have an account to test this against at the moment. Let me know if you can confirm. Cheers!

mbrowne:
Sorry for the delay, and good catch—I've now added `omitempty` to the `Priority` field and tested adding a rule both with and without a Priority, and it's working well.

jacobbednarz:
@mbrowne Looks like CI is failing due to your change (see 1.11.x https://travis-ci.org/cloudflare/cloudflare-go/jobs/533886264#L528). Could you take a look?

lockdown_example_test.go
```diff
@@ -29,3 +30,36 @@ func ExampleAPI_ListZoneLockdowns_all() {
 		fmt.Printf("%s: %s\n", strings.Join(r.URLs, ", "), r.Configurations)
 	}
 }
+
+func ExampleAPI_CreateZoneLockdown(t *testing.T) {
```
jacobbednarz: Should resolve CI failures (you don't need `t` here at all).

```suggestion
func ExampleAPI_CreateZoneLockdown() {
```

mbrowne: Ah, sorry, I had renamed the test to `TestExampleAPI_CreateZoneLockdown` in order to quickly run it and forgot to remove the `t` argument afterwards. Thanks

jacobbednarz:
cc @patryk @zackproser

mbrowne:
@patryk or @zackproser, can you approve? Just a friendly reminder that this should be ready to go now. Thanks @jacobbednarz.

patryk:
Oh yes, absolutely! Sorry for the delay, @mbrowne. I appreciate your patience and thanks for help.

PR #268, opened 2019-01-25
flarectl: dnsCreateOrUpdate() improvements

joepurdy:
## Description

👋 These changes address a couple of issues with the `flarectl dns create-or-update` command. Originally I was focused on https://github.com/cloudflare/cloudflare-go/issues/206, but also saw an opportunity to fix https://github.com/cloudflare/cloudflare-go/issues/249 and https://github.com/cloudflare/cloudflare-go/issues/250 while I was at it. Happy to deconstruct this into multiple PRs if that's the preferred route. 

### Fixes https://github.com/cloudflare/cloudflare-go/issues/206
I assume the motivation for `Name: name + "." + zone` was to handle cases where naked subdomains are passed to the `--name` flag.
EX:
```
flarectl dns create-or-update --zone example.com --type A --name app --content 192.0.2.70

instead of...

flarectl dns create-or-update --zone example.com --type A --name app.example.com --content 192.0.2.70
```
In that case, attempting to search for only `app` will always return no matches even when a record does exist for `app.example.com`.

However always appending the zone to whatever is passed for name makes it impossible to work on root records (`example.com`). To get around this I changed the behavior to first search for the raw value of the passed `--name` flag and only append the zone if the first search fails to find a record. If both searches fail to find records the name is set back to whatever was passed via the `--name` flag and execution continues as before with creating a new record.

### Fixes https://github.com/cloudflare/cloudflare-go/issues/249
Added a couple `return` statements to halt execution when errors occur

### Fixes https://github.com/cloudflare/cloudflare-go/issues/250
In the case a record was found and `api.UpdateDNSRecord()` was called to update the existing record there was a lovely `panic: runtime error: invalid memory address or nil pointer dereference` due to the `resp` variable never being initialized.

Simple fix, assign `resp.Result` manually after a successful update call.

## Has your change been tested?

Doesn't look like there's a automated testing in the flarectl commands (testing CLIs is weird). That said I ran some fun before and afters against one of my test domains on Cloudflare to validate these changes. I've made some redactions/alterations to output for privacy:

### Attempt to add a root DNS record:
Before:
```
flarectl d o --zone example.com --type A --proxy --name example.com --content 192.0.2.70
                 ID                |                         NAME                         | TYPE  |            CONTENT            | TTL | PROXIABLE | PROXY | LOCKED
+----------------------------------+------------------------------------------------------+-------+-------------------------------+-----+-----------+-------+--------+
[...]

Issue #250, opened 2018-11-16
flarectl: dnsCreateOrUpdate() attempts to format nonexistent response after updating record


pmocek:
In flarectl's dnsCreateOrUpdate(), a variable `resp` is declared as pointer to a cloudflare.DNSRecordResponse.  If the record to be created or updated is found not to exist, then api.CreateDNSRecord() is called and `resp` is pointed to the response object.  If instead that record is found to exist, then api.UpdateDNSRecord() is called instead, no response is captured, and `resp` remains a nil pointer.  In either case, `resp` is then passed to formatDNSRecord(), which fails after an update because `resp` remains nil.

PR #270, opened 2019-01-30
flarectl: dnsCreateOrUpdate() fix nil pointer exception

joepurdy:
More context in the original PR here: https://github.com/cloudflare/cloudflare-go/pull/268

This spins off the fix for https://github.com/cloudflare/cloudflare-go/issues/250 into its' own PR.


ironsalsa:
Any motion on this one? I ran into the same problem, and I'd love to use this command in CI without a NPE.

PR #276, opened 2019-02-19
Add Minify in PageRuleActions

dtupitsin:
Add Auto-Minify in PageRuleActions

Need this action for create Auto-Minify in PageRules by terraform provider (terraform-providers/terraform-provider-cloudflare#120)

Also made the flarectl output more readable on pagerules with minify.
Before:
```
$ ~/go/bin/flarectl p l --zone example.org
Pri ID                               Status   URL
  2 ad668eaaa0de1ea66dc26939491cced0 active   https://example.org/*/test/_static_/*
    : %!f(<nil>) - %!s(<nil>)
  1 b1ea70598683166ebf41b256ab9368b9 active   https://example.org/test/*.js
    : %!f(<nil>) - %!s(<nil>)
```
After:
```
$ ~/go/bin/flarectl p l --zone example.org
Pri ID                               Status   URL
  2 ad668eaaa0de1ea66dc26939491cced0 active   https://example.org/*/test/_static_/*
    Minify: [js:on html:off css:on]
  1 b1ea70598683166ebf41b256ab9368b9 active   https://example.org/test/*.js
    Minify: [js:on html:off css:off]
```

I'm not sure that everything is done correctly. Waiting for check

### DIFF
--- a/cmd/flarectl/misc.go
+++ b/cmd/flarectl/misc.go
@@ -147,7 +147,11 @@ func pageRules(c *cli.Context) {
 			case float64:
 				s = fmt.Sprintf("%s: %.f", cloudflare.PageRuleActions[a.ID], v)
 			case map[string]interface{}:
-				s = fmt.Sprintf("%s: %.f - %s", cloudflare.PageRuleActions[a.ID], v["status_code"], v["url"])
+				if a.ID == "minify" {
+					s = fmt.Sprintf("%s: [js:%s html:%s css:%s]", cloudflare.PageRuleActions[a.ID], v["js"], v["html"], v["css"])
+				} else {
+					s = fmt.Sprintf("%s: %.f - %s", cloudflare.PageRuleActions[a.ID], v["status_code"], v["url"])
+				}
 			case nil:
 				s = fmt.Sprintf("%s", cloudflare.PageRuleActions[a.ID])
 			default:
--- a/page_rules.go
+++ b/page_rules.go
@@ -43,6 +43,7 @@ Valid IDs are:
   forwarding_url
   host_header_override
   ip_geolocation
+  minify
   mirage
   opportunistic_encryption
   origin_error_page_pass_thru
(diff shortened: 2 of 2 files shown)


cmd/flarectl/misc.go
```diff
@@ -147,7 +147,11 @@ func pageRules(c *cli.Context) {
 			case float64:
 				s = fmt.Sprintf("%s: %.f", cloudflare.PageRuleActions[a.ID], v)
 			case map[string]interface{}:
-				s = fmt.Sprintf("%s: %.f - %s", cloudflare.PageRuleActions[a.ID], v["status_code"], v["url"])
+				if a.ID == "minify" {
+					s = fmt.Sprintf("%s: [js:%s html:%s css:%s]", cloudflare.PageRuleActions[a.ID], v["js"], v["html"], v["css"])
+				} else {
```
jacobbednarz: This is a little hap-hazardous as we're using an `if`/`else` and checking for the `minify` ID as if it's a special case amongst another group of common values (which it's not as it's a 50/50 between the two options). I think we might be better off either:

- Defining a custom formatter that is reusable for all `map[string]interface{}`. Perhaps a field to value output? Example, `[status_code: 301 url: https://example.com]` and `[js: true html:true css: true]` type of thing.
- Explicitly calling out the two conditions in their own `if`s to avoid any accidental classifications.

I think the first one is a more appropriate action to ensure we get some consistency but I'm not too fussed. I just foresee some accidents if we leave it as is.

jacobbednarz:
Thanks for this one @dtupitsin! I've left a small suggestion to (hopefully) eliminate any accidental classification of the `if`/`else` condition you've introduced by being more explicit about the intentions.

Sing out if you need any help implementing it.

jacobbednarz:
I've opened #297 (without the flarectl changes) as we are looking to add this in the Terraform provider and are currently blocked.

PR #279, opened 2019-02-28
flarectl: Add support for organization scopes

fasmide:
## Description

This merge requests adds a global flag `--orgid` used when users want to change to an organization scope. 

I'm working on some `loadbalancer` and `pool` subcommands where users are required to change scope. I will open a pull request when ready.

Also: Cleaning the code up a bit - this way we don't have to push around `checkEnv` function calls everywhere

I have tested that `--orgid` is used - `flarectl ips` still works as expected even when not providing any credentials and all the other sub commands are working as before when given credentials

## Types of changes

- [x] Refactoring (move code and functions around in a way that does not break functionality)

## Checklist:

### DIFF
--- a/cmd/flarectl/dns.go
+++ b/cmd/flarectl/dns.go
@@ -24,10 +24,6 @@ func formatDNSRecord(record cloudflare.DNSRecord) []string {
 }
 
 func dnsCreate(c *cli.Context) {
-	if err := checkEnv(); err != nil {
-		fmt.Println(err)
-		return
-	}
 	if err := checkFlags(c, "zone", "name", "type", "content"); err != nil {
 		return
 	}
@@ -65,10 +61,6 @@ func dnsCreate(c *cli.Context) {
 }
 
 func dnsCreateOrUpdate(c *cli.Context) {
-	if err := checkEnv(); err != nil {
-		fmt.Println(err)
-		return
-	}
 	if err := checkFlags(c, "zone", "name", "type", "content"); err != nil {
 		fmt.Println(err)
 		return
@@ -136,10 +128,6 @@ func dnsCreateOrUpdate(c *cli.Context) {
 }
 
 func dnsUpdate(c *cli.Context) {
-	if err := checkEnv(); err != nil {
-		fmt.Println(err)
-		return
-	}
 	if err := checkFlags(c, "zone", "id"); err != nil {
 		fmt.Println(err)
 		return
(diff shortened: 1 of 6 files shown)


cmd/flarectl/flarectl.go
```diff
@@ -14,6 +14,13 @@ func main() {
 	app.Name = "flarectl"
 	app.Usage = "Cloudflare CLI"
 	app.Version = "2017.10.0"
+	app.Flags = []cli.Flag{
+		cli.StringFlag{
+			Name:  "orgid",
```
jacobbednarz:
```suggestion
			Name:  "org-id",
```

Perhaps `org-id` is a touch easier to read?

cmd/flarectl/flarectl.go
```diff
@@ -14,6 +14,13 @@ func main() {
...
+			Usage: "Optional organization ID",
+			Value: "",
```
jacobbednarz: Instead of only relying on a flag, have you considered using `EnvVar` to capture this value too?

```suggestion
			Value: "",
			EnvVar: "CF_ORG_ID"
```

cmd/flarectl/dns.go
```diff
@@ -24,10 +24,6 @@ func formatDNSRecord(record cloudflare.DNSRecord) []string {
 }
 
 func dnsCreate(c *cli.Context) {
-	if err := checkEnv(); err != nil {
```
jacobbednarz: I think this cleanup 💯 but do wonder if we could move it to a single `app.Before` instead of each individual flag 🤔

fasmide: I tried it and while looking a lot cleaner, it breaks the usecase of `flarectl ips` which does not require any credentials so i think we are left with sticking it onto the groups of subcommands that does require credentials :)

cmd/flarectl/flarectl.go
fasmide: I think your right about that :)

cmd/flarectl/flarectl.go
fasmide: This is really a nice addition

fasmide:
@jacobbednarz do you know why the builds fail? - i see no reason why this MR should have anything todo with being able to `go get` the `golint` package

```
go get: error loading module requirements
The command "go get -u golang.org/x/lint/golint" failed and exited with 1 during .
```

jacobbednarz:
Doesn't look like `golint` is in the module file. You should be able to add it to the module file, commit and try the CI run again.

```
export GO111MODULE=on
go get -u golang.org/x/lint/golint
```

patryk:
Thanks @fasmide !

garrettgalow:
we should be using account id as the descriptor not org id. Orgs are deprecated. Note the ids themselves are the same, so this is really just about the naming

patryk:
Good point, will change that.

patryk:
Done. @fasmide please check everything still works fine for you.

PR #288, opened 2019-03-27
zone: Paginate `ListZones` for all zones

jacobbednarz:
The `ListZones` function has two distinct code paths. The first is for
fetching zones explicitly. You do this by passing in the zones you wish
to return.

```go
ListZones("example.org", "example.net")
```

This will return two `Zone` structs for `example.org` and `example.net`.

The other code path is returning all the zones the user has access to.
Previously, this would only return the first page of results (generally
20 zones). Many customers wouldn't have noticed this was the case but
some larger customers have hundreds or even thousands and would need to
perform multiple requests within their client to get return all pages.

This change addresses the latter by paginating through all pages. It
does this using a `sync.WaitGroup` and sending off concurrent requests
based on the number of `total_pages` returned in the API response.
Channels were considered here but introduced too much complexity and
gun-pointing-at-foot scenarios where it really wasn't needed. The only
use of channels here is to keep track of any errors that pop up during
the request parsing cycle.

### DIFF
--- a/zone.go
+++ b/zone.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


jacobbednarz:
I did consider swapping both code paths to use the pagination but I don't really see a use case for passing in > 20 zones. Could be wrong but seems unlikely 🤷‍♂️

ton31337:
When will this be released? We have ~45 zones in Cloudflare and Terraform just does not work when listing all zones. 20 is the default.

jacobbednarz:
This was released many months ago and Terraform also handles automatically paginating the results. If you're seeing different behaviours, please open a new issue with a reproduction test case in the appropriate repository.

PR #292, opened 2019-04-09
Add OwnershipChallenge token field to LogpushJob struct

zackproser:
To facilitate forthcoming support for Cloudflare logpush jobs in the Cloudflare Terraform provider, this change adds an ```OwnershipChallenge``` token field to the LogpushJob struct. 

LogpushJob create and update tasks require that this token be passed in the body of the API request.

This was tested locally and works alongside the forthcoming Terraform ```cloudflare_logpush_job``` resource.

## Types of changes

## Checklist:

### DIFF
--- a/logpush.go
+++ b/logpush.go
@@ -15,7 +15,7 @@ type LogpushJob struct {
 	Name               string     `json:"name"`
 	LogpullOptions     string     `json:"logpull_options"`
 	DestinationConf    string     `json:"destination_conf"`
-	OwnershipChallenge string     `json:"ownership_challenge"`
+	OwnershipChallenge string     `json:"ownership_challenge,omitempty"`
 	LastComplete       *time.Time `json:"last_complete,omitempty"`
 	LastError          *time.Time `json:"last_error,omitempty"`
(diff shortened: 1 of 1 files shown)


logpush.go
```diff
@@ -10,14 +10,15 @@ import (
...
-	DestinationConf string     `json:"destination_conf"`
-	LastComplete    *time.Time `json:"last_complete,omitempty"`
-	LastError       *time.Time `json:"last_error,omitempty"`
-	ErrorMessage    string     `json:"error_message,omitempty"`
+	ID                 int        `json:"id,omitempty"`
+	Enabled            bool       `json:"enabled"`
+	Name               string     `json:"name"`
+	LogpullOptions     string     `json:"logpull_options"`
+	DestinationConf    string     `json:"destination_conf"`
+	OwnershipChallenge string     `json:"ownership_challenge"`
```
jacobbednarz: Do we want to omit this field when empty? I see a few of the operations don't have this field set (for example, getting a logpush job) where an empty string might be a touch confusing.

zackproser: Good point - I believe we do. Will update. Thanks for reviewing!

PR #293, opened 2019-04-09
Support Resource Bindings in UploadWorker

grobinson-cloudflare:
This PR makes it possible to bind Worker scripts to KV namespaces when uploading scripts via the API.

## Types of changes

## Checklist:

### DIFF
--- a/workers.go
+++ b/workers.go
@@ -1,17 +1,32 @@
 package cloudflare
 
 import (
+	"bytes"
 	"encoding/json"
+	"io"
+	"mime/multipart"
 	"net/http"
 	"time"
 
 	"github.com/pkg/errors"
 )
 
+type WorkerMetadataParams struct {
+	BodyPart string           `json:"body_part"`
+	Bindings []WorkerResource `json:"bindings"`
+}
+
 // WorkerRequestParams provides parameters for worker requests for both enterprise and standard requests
 type WorkerRequestParams struct {
 	ZoneID     string
 	ScriptName string
+}
+
+// WorkerResource binds worker scripts to resources
+type WorkerResource struct {
+	Type        string `json:"type"`
+	Name        string `json:"name"`
+	NamespaceID string `json:"namespace_id"`
 }
 
 // WorkerRoute aka filters are patterns used to enable or disable workers that match requests.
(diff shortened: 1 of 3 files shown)


workers.go
```diff
@@ -167,15 +182,41 @@ func (api *API) ListWorkerScripts() (WorkerListResponse, error) {
 // UploadWorker push raw script content for your worker
 //
 // API reference: https://api.cloudflare.com/#worker-script-upload-worker
-func (api *API) UploadWorker(requestParams *WorkerRequestParams, data string) (WorkerScriptResponse, error) {
+func (api *API) UploadWorker(requestParams *WorkerRequestParams, data string, metadataParams *WorkerMetadataParams) (WorkerScriptResponse, error) {
```
sevki: there is no requirement to upload metadata and changing the public facing interface would require people to update their code to pass in a second nil parameter. 

What do you think about updating this PR to keep this function as is and adding a new `UploadWorkerWithResourceBindings` function to so we don't change API?

workers.go
```diff
@@ -167,15 +182,41 @@ func (api *API) ListWorkerScripts() (WorkerListResponse, error) {
...
+
+func (api *API) uploadWorker(requestParams *WorkerRequestParams, data string, metadataParams *WorkerMetadataParams) (WorkerScriptResponse, error) {
 	uri := "/zones/" + requestParams.ZoneID + "/workers/script"
+	if metadataParams != nil {
+		return api.doMultipartUploadWorker(uri, data, *metadataParams)
+	}
+	return api.doUploadWorker(uri, data)
+}
+
+// UploadNamedWorker push raw script content for your worker
```
sevki: Is this referred to as a `NamedWorker` anywhere else?
@deuill?

workers.go
```diff
@@ -186,19 +227,31 @@ func (api *API) UploadWorker(requestParams *WorkerRequestParams, data string) (W
...
+	if err != nil {
+		return r, errors.Wrap(err, errMakeRequestError)
 	}
-	uri := "/accounts/" + api.OrganizationID + "/workers/scripts/" + scriptName
 	headers := make(http.Header)
-	headers.Set("Content-Type", "application/javascript")
-	res, err := api.makeRequestWithHeaders("PUT", uri, []byte(data), headers)
-	var r WorkerScriptResponse
+	headers.Set("Content-Type", writer.FormDataContentType())
+	res, err := api.makeRequestWithHeaders("PUT", uri, body.Bytes(), headers)
```
sevki: should we consider switching this to [`http.MethodPut`](https://golang.org/pkg/net/http/#pkg-constants) too to make it more consistent with line 219?

workers.go
```diff
@@ -186,19 +227,31 @@ func (api *API) UploadWorker(requestParams *WorkerRequestParams, data string) (W
...
-//
-// API reference: https://api.cloudflare.com/#worker-script-upload-worker
-func (api *API) uploadWorkerWithName(scriptName string, data string) (WorkerScriptResponse, error) {
-	if api.OrganizationID == "" {
-		return WorkerScriptResponse{}, errors.New("organization ID required for enterprise only request")
+func (api *API) doMultipartUploadWorker(uri, data string, metadata WorkerMetadataParams) (WorkerScriptResponse, error) {
+	var r WorkerScriptResponse
+	body := &bytes.Buffer{}
+	writer := multipart.NewWriter(body)
+	metadataPart, err := writer.CreateFormFile("metadata", "metadata.json")
```
sevki: we should probably not hardcode these as  `metadata.json` and `script.js` are not the only available options 
https://developers.cloudflare.com/workers/api/resource-bindings/webassembly-modules/

patryk:
Merged a similar solution in #363. Thanks @grobinson-cloudflare !

PR #294, opened 2019-04-15
zone: Add support for partial zone management

jacobbednarz:
Updates the `CreateZone` function to allow the creation of a partially
managed zone within Cloudflare. Prior to this change, all zones were
assumed as full setups.

This unblocks some work in
terraform-providers/terraform-provider-cloudflare#280 where we would
like the ability to manage all zone types instead of just full zones.


jacobbednarz:
cc @zackproser @patryk @farberjd

farberjd:
Thanks for your work on this Jacob!

PR #295, opened 2019-04-16
page_rules: Send `rule` as payload

jacobbednarz:
Updates `UpdatePageRule` to send the full page rule payload for updates. Prior to this, no payload is sent using the `UpdatePageRule` function.


jacobbednarz:
cc @patryk

PR #301, opened 2019-05-09
URL Escape the URI

ironsalsa:
Currently, the path is used for submitting data that may contain special characters. For example, if you try to set up bulk redirects in Workers with KV per https://developers.cloudflare.com/workers/recipes/bulk-redirects/, the key '/pathtoreplace' is stored in KV as 'pathtoreplace'. Currently, the caller has to URL encode the submission to the api.WriteWorkersKV() call to ensure the slashes get through.

I can't think of a situation where we wouldn't want to URL encode a path, since it's built with a simple Sprintf() call in the higher level functions - adding an encode would the right thing instead of passively allowing data loss.

Provide a general summary of your changes in the title above. You should
remove this overview, any sections and any section descriptions you
don't need below before submitting. There isn't a strict requirement to
use this template if you can structure your description and still cover
these points.

## Description

Describe your changes in detail through motivation and context. Why is
this change required? What problem does it solve? If it fixes an open
issue, link to the issue using GitHub's closing issues keywords[1].

required.

## Types of changes

## Checklist:

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -267,6 +267,7 @@ func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, u
 // *http.Response, or an error if one occurred. The caller is responsible for
 // closing the response body.
 func (api *API) request(ctx context.Context, method, uri string, reqBody io.Reader, authType int, headers http.Header) (*http.Response, error) {
+	uri = url.PathEscape(uri)
 	req, err := http.NewRequest(method, api.BaseURL+uri, reqBody)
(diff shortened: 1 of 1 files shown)


ironsalsa:
Sorry I didn't fill out the whole form - I believe this should be a simple change that corrects some nasty unwanted behavior, but let me know if you want the full process for it.

ironsalsa:
I opened https://github.com/cloudflare/cloudflare-go/pull/302 for a more targeted fix since this was a mess.

PR #309, opened 2019-05-22
Bulk workers kv endpoint bindings

Michael9127:
Adds bindings for new bulk workers KV endpoint: https://api.cloudflare.com/#workers-kv-namespace-write-multiple-key-value-pairs

Check err return of rate limiting

Testing:
* Manually tested against personal CF namespace
* Added unit test to cover new method

PR #310, opened 2019-05-27
[BREAKING] Workers KV refactor, add bulk write and key expiry API

Electroid:
## Description
In trying to implement expiring keys and looking at the PR (#309) for bulk endpoints, I quickly realized the Workers KV module needs some refactoring. If we plan on having features like access control in the future, we'll want to do this sooner than later. Since KV is still recently GA and we (storage team) know how many people are using KV, the API disruption should be minimal.

### New
* Added `WorkersKVPair` to unify parameters for read/write and singleton/bulk APIs
* Add new bulk write API
  * `WriteWorkersKVPairBulk(ns, key, []*struct)`

### Backwards-Compatible
* Expose `http.Header` to `makeRequest` method
* Ensure consistent nomenclature to `WorkersKV<Type><Action>`
  * Pure struct renames, with backwards compatible aliases
    * `ListWorkersKVNamespacesResponse` -> `WorkersKVNamespaceListResponse`
    * `StorageKey` -> `WorkersKVKey`
  * Method renames and parameter changes, old methods now deprecated
     * `WriteWorkersKV(ns, key, value)` -> `WriteWorkersKVPair(ns, *struct)`
    * `ReadWorkersKV(ns, key) -> value` -> `ReadWorkersKVPair(ns, key) -> struct`

### Breaking
* Redo key listing API
   * Previous API was coded and mocked incorrectly, there was no cursor implementation
   * Because of this, it shows no one was using it, so we can make this a breaking change
     * Added struct `WorkersKVKeyListRequest`
     * Added struct `WorkersKVKeyListResponse`
     * Added struct `WorkersKVKeyPagination`
     * Added method `ListWorkersKVKeys(ns, *request) -> response`

### Misc
* Test suite fully updated
* All changes tested with production Workers KV
* If merged, add @Michael9127 as co-author for his bulk work

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -159,6 +159,11 @@ func (api *API) makeRequestWithAuthType(ctx context.Context, method, uri string,
 }
 
 func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, uri string, params interface{}, authType int, headers http.Header) ([]byte, error) {
+	res, _, err := api.makeRequestWithAuthTypeAndHeadersTuple(ctx, method, uri, params, authType, headers)
+	return res, err
+}
+
+func (api *API) makeRequestWithAuthTypeAndHeadersTuple(ctx context.Context, method, uri string, params interface{}, authType int, headers http.Header) ([]byte, http.Header, error) {
 	// Replace nil with a JSON object if needed
 	var jsonBody []byte
 	var err error
@@ -169,7 +174,7 @@ func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, u
 		} else {
 			jsonBody, err = json.Marshal(params)
 			if err != nil {
-				return nil, errors.Wrap(err, "error marshalling params to JSON")
+				return nil, nil, errors.Wrap(err, "error marshalling params to JSON")
 			}
 		}
 	} else {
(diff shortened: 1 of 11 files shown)


jacobbednarz:
There is alot going on in this PR and I'm unsure if we want to couple this discussions together. Do we want to break this into separate PRs and look at handling them individually?

Personally, I think we're getting pretty carried away with how we manage the authentication to the API and what type of function signatures we're relying on (for instance `makeRequestWithAuthTypeAndHeadersAlsoAttachHeaders` is a mouth full and covers a bit too much). I think we might want to take a look at cutting back on what we handle at the client/auth level and maybe expose a centralised, configurable way to perform the desired authentication instead. It looks like we've started down this path but got a bit tangled up with the specifics of what the Workers KV stuff requires.  Depending on how deep you want to go with this, perhaps something more focused in [this](https://github.com/cloudflare/cloudflare-go/blob/master/cloudflare.go#L291-L293) area of the client creation would be slimmer and get the desired result without relying so much on the existing `authType*` functions? One option I can think of is a map of URLs to the content type they need and then check if the current URL is present in that map and what content type it needs otherwise default to `application/json`.

Electroid:
Agree on the mouthful, for sure. At the same time, we probably don't want to make the KV changes coupled with another refactor. I slimmed down the client changes, see what you think -- I think it's a good compromise before that part of the code needs some attention.

Electroid:
Slimmed some more diffs and tested with a library that currently uses cloudflare-go to test for compatibility. If we're okay with the header exposure, the KV is ready for review.

jacobbednarz:
This seems like a reasonable compromise. When this goes downstream (to things like Terraform) we will probably need to do a fair bit of reworking to update to the non-deprecated versions but the interface shouldn't change so should be fine as a backwards compatible change 👍

PR #311, opened 2019-05-27
Add Worker ResourceBinding API

Electroid:
### Description

Adds functionality to upload worker scripts with the multi-part method, as well as listing bindings of an existing script. Fully backwards compatible and tested in production. Hopefully this will give way to add full Worker support in Terraform.

### Functions
  * `UploadWorkerMultiPart` 
  * `ListWorkerBindings`
### Structs
  * `WorkerResource`
    * `WorkerResourceMetaData`
  * `WorkerResourceListResponse`
  * `WorkerUploadRequest`
    * `WorkerUploadPart`
### Refactoring
  * Add helper functions to reduce boilerplate code `uriOfWorker`

### DIFF
--- a/workers.go
+++ b/workers.go
@@ -1,8 +1,14 @@
 package cloudflare
 
 import (
+	"bytes"
 	"encoding/json"
+	"fmt"
+	"io"
+	"mime/multipart"
 	"net/http"
+	"net/textproto"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


Electroid:
@jacobbednarz Are you internal at Cloudflare? Or do you happen to know the point of contact for this project?

jacobbednarz:
I'm not sorry, @patryk or @zackproser should be able to point you in the right direction though.

avidal:
@Electroid let's hold off on this for a bit. As it's the public API for our Go library we want to get the interfaces right, and I don't think that making them manually construct the multipart requests is the right way to go about it. We should be able to define an abstraction over it which allows you to create a `Worker` that has attached resources and upload it. No need to explicitly upload multipart, etc. We should be able to provide a single mechanism to callers.

patryk:
Thanks @Electroid, I merged a similar solution in #363, so I'm going to close this one.

PR #314, opened 2019-06-13
Support for worker resource bindings

### LINKED ISSUE #313: Add support for worker resource bindings
Provide support for the Resource Bindings API for Worker Scripts as detailed here https://developers.cloudflare.com/workers/api/resource-bindings/. This will allow programatic access to bind scripts to KV namespaces and WASM modules.

## Expected behavior

Implement a `UploadWorkerWithBindings` method to attach extra data to a request including KV namespaces and WASM bytecode in the form of an `io.Reader`. This will then perform a multipart/form-data request to the CF api.

## Current behavior

Not currently supported.

## Possible solution

I have a PR ready :)

## Context

My company needs to be able to manage it's CF Workers via Terraform and we need to be able to bind KV namespaces. This change will enable the corresponding changes within the Terraform provider to be made.

adampointer-form3:
resolves #313 

## Description

To enable the management of worker resource bindings in Terraform.

Automated test coverage.

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [x ] My code follows the code style of this project.
- [ x] All new and existing tests passed.

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -181,30 +181,35 @@ func (api *API) makeRequestWithAuthType(ctx context.Context, method, uri string,
 
 func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, uri string, params interface{}, authType int, headers http.Header) ([]byte, error) {
 	var err error
-	var reqBody io.Reader
-
+	var jsonBody []byte
 	if params != nil {
 		switch v := params.(type) {
 		case []byte:
-			// If params is []byte then pass through to the request body untouched as a io.Reader
-			reqBody = bytes.NewReader(v)
+			// If params is []byte then pass through to the request body untouched
+			jsonBody = v
 		case io.Reader:
-			// If params is already an io.Reader, e.g. from a multipart form then pass through
-			reqBody = v
+			// If params is already an io.Reader, e.g. from a multipart form then consume into bytes
+			jsonBody, err = ioutil.ReadAll(v)
+			if err != nil {
+				return nil, errors.Wrap(err, "error reading request body")
+			}
 		default:
 			// Anything else we try to JSON encode
-			jsonBody, err := json.Marshal(params)
+			jsonBody, err = json.Marshal(params)
 			if err != nil {
 				return nil, errors.Wrap(err, "error marshalling params to JSON")
 			}
-			reqBody = bytes.NewReader(jsonBody)
 		}
 	}
 
 	var resp *http.Response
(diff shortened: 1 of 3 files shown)


jacobbednarz:
Is this different to the work in #311 and #310? At first glance, there looks to be some overlap.

adampointer-form3:
Hi @jacobbednarz yes I did not see #311 there is some overlap. I would say that as you asked for more abstraction over the creation of the multipart message, mine fulfils those criteria better? There is no overlap with #310 I have been watching that one as I am also looking for KV support in Terraform :) Mine is just for creating WASM and KV bindings rather than managing KV namespaces.

adampointer-form3:
Actually it duplicates #293 too

Electroid:
I like the higher abstraction here. @adampointer-form3, is there a way we can abstract the `part` from WASM as well? Seems like an implementation detail from the multi-part upload we would want to hide.

adampointer-form3:
Is that better?

adampointer-form3:
@Electroid are there any other requirements here? This is starting to block internal projects. I don't want to resort to using forks really.

cloudflare.go
```diff
@@ -159,31 +159,31 @@ func (api *API) makeRequestWithAuthType(ctx context.Context, method, uri string,
 }
 
 func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, uri string, params interface{}, authType int, headers http.Header) ([]byte, error) {
-	// Replace nil with a JSON object if needed
-	var jsonBody []byte
 	var err error
+	var reqBody io.Reader
```
avidal: I believe that by moving the io.Reader setup outside of the retry loop you run the risk of retries being broken, since the first attempt may consume the reader and subsequent attempts will then have no request body to read from.

If that's the case it may make more sense to consume `params` into a byte slice outside of the retry loop and create a reader out of it within the retry loop, similar to the original implementation.

If params *is* an io.Reader, maybe just ioutil.ReadAll to read it into the byte slice.

adampointer-form3: Ok I have made the change as suggested.

jacobbednarz:
....and now duplicates #363 too 🙂 By the looks of things, we can go with #363 and once that is merged, build off it as it should cover the previous use cases mentioned in other PRs too.

patryk:
Closing on behalf of #363 which I merged. Thank you @adampointer-form3 for the effort, I appreciate that.

PR #315, opened 2019-06-16
Deprecate use of `Organization`

jacobbednarz:
For a while now the Organization endpoints have been [deprecated in
favour of Account based endpoints](https://api.cloudflare.com/#organizations-properties). This introduces the cleanup which
removes all references to the deprecated endpoints as well as replacing
the associated helper methods such as `UsingOrganization` with
`UsingAccount`.

See terraform-providers/terraform-provider-cloudflare#227 for further
details on this breaking change.

cc @patryk @garrettgalow


user.go
```diff
@@ -9,20 +9,20 @@ import (
...
+	Username   string     `json:"username,omitempty"`
+	Telephone  string     `json:"telephone,omitempty"`
+	Country    string     `json:"country,omitempty"`
+	Zipcode    string     `json:"zipcode,omitempty"`
+	CreatedOn  *time.Time `json:"created_on,omitempty"`
+	ModifiedOn *time.Time `json:"modified_on,omitempty"`
+	APIKey     string     `json:"api_key,omitempty"`
+	TwoFA      bool       `json:"two_factor_authentication_enabled,omitempty"`
+	Betas      []string   `json:"betas,omitempty"`
+	Accounts   []Account  `json:"organizations,omitempty"`
```
jacobbednarz: I've updated the field name to be correct however the JSON response is still for `organizations` so this needs to stay until the API responses are either updated or conditionally respond with different values based on the incoming request.

zone.go
```diff
@@ -265,24 +265,24 @@ type newZone struct {
 	Type      string `json:"type"`
 	// We use a pointer to get a nil type when the field is empty.
 	// This allows us to completely omit this with json.Marshal().
-	Organization *Organization `json:"organization,omitempty"`
+	Account *Account `json:"organization,omitempty"`
```
jacobbednarz: I've updated the field name to be correct however the JSON response is still for `organizations` so this needs to stay until the API responses are either updated or conditionally respond with different values based on the incoming request.

options.go
```diff
@@ -28,11 +28,11 @@ func Headers(headers http.Header) Option {
 	}
 }
 
-// UsingOrganization allows you to apply account-level changes (Load Balancing,
```
jacobbednarz: I'd really like to deprecate this method entirely due to the amount of edge cases and additional code paths it introduces. Ideally explicit use would trump a helper method like this (including in the terraform provider).

PR #340, opened 2019-08-21
Add support for Access Identity Providers

jacobbednarz:
Updates the library to have full support for Access Identity Providers.

API reference: https://api.cloudflare.com/#access-identity-providers-properties
Identity provider reference: https://developers.cloudflare.com/access/configuring-identity-providers/

### DIFF
--- a/access_identity_provider.go
+++ b/access_identity_provider.go
@@ -1,0 +1,331 @@
+package cloudflare
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/pkg/errors"
+)
+
+// AccessIdentityProvider is the structure of the provider object.
+type AccessIdentityProvider struct {
+	ID     string      `json:"id,omitemtpy"`
+	Name   string      `json:"name"`
+	Type   string      `json:"type"`
+	Config interface{} `json:"config"`
+}
+
+// AzureADConfiguration is the representation of the Azure AD identity
+// provider.
+//
+// API reference: https://developers.cloudflare.com/access/configuring-identity-providers/azuread/
+type AzureADConfiguration struct {
+	ClientID      string `json:"client_id"`
+	ClientSecret  string `json:"client_secret"`
+	DirectoryID   string `json:"directory_id"`
+	SupportGroups bool   `json:"support_groups"`
+}
+
+// CentrifyConfiguration is the representation of the Centrify identity
+// provider.
+//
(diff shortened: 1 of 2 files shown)


access_identity_provider.go
```diff
@@ -0,0 +1,327 @@
...
+	Type   string      `json:"type"`
+	Config interface{} `json:"config"`
+}
+
+// AzureADConfiguration is the representation of the Azure AD identity
+// provider.
+//
+// API reference: https://developers.cloudflare.com/access/configuring-identity-providers/azuread/
+type AzureADConfiguration struct {
+	ClientID      string `json:"client_id"`
```
jacobbednarz: I made a design decision not to embed these as a shared struct as it becomes too hard working with it and testing it.

access_identity_provider.go
```diff
@@ -0,0 +1,327 @@
...
+	ClientID     string `json:"client_id"`
+	ClientSecret string `json:"client_secret"`
+}
+
+// JumpCloudSAMLConfiguration is the representation of the Jump Cloud
+// identity provider using SAML.
+//
+// API reference: https://developers.cloudflare.com/access/configuring-identity-providers/jumpcloud-saml/
+type JumpCloudSAMLConfiguration struct {
+	IssuerURL          string   `json:"issuer_url"`
```
jacobbednarz: Same comment as https://github.com/cloudflare/cloudflare-go/pull/340/files#r316000938 here despite the SAML authentication sources all being pretty similar.

jacobbednarz:
cc @deuill

access_identity_provider.go
```diff
@@ -0,0 +1,327 @@
...
+	ClientSecret  string `json:"client_secret"`
+	DirectoryID   string `json:"directory_id"`
+	SupportGroups bool   `json:"support_groups"`
+}
+
+// CentrifyConfiguration is the representation of the Centrify identity
+// provider.
+//
+// API reference: https://developers.cloudflare.com/access/configuring-identity-providers/centrify/
+type CentrifyConfiguration struct {
```
jacobbednarz: I haven't prefixed these structs with `Access` however I'm thinking I might. WDYT?

patryk: I'd say yes, because it would be hard to find it around later, let's prefix with something like: `AccessProviderCentrify`

jacobbednarz: Updated in 2110054

PR #348, opened 2019-09-26
Add support for IDN zone names in Punycode form

### LINKED ISSUE #347: Accesssing IDN zones fails with Punycode domains
## Expected behavior

Given `with-ümlaut.de` as a zone in my Cloudflare. Should `(*cloudflare.API).ZoneIDByName()` work for both the Unicode and the Punycode representation (`xn--with-mlaut-eeb.de`)?

```go
const token = "..."
func show(zoneName string) {
  api, _ := cloudflare.NewWithAPIToken(token)
  if id, err := api.ZoneIDByName(zoneName); err != nil {
    fmt.Println("err:", err)
  } else {
    fmt.Println("id:", id)
  }
}
```

## Current behavior

```go
show("with-ümlaut.de")
//=> id: ...
show("xn--with-mlaut-eeb.de")
//=> err: Zone could not be found
```

## Possible solution

Apply [`x/net/idna.ToUnicode`](https://godoc.org/golang.org/x/net/idna#ToUnicode) before sending the request to the API server.

This assumes that your database records do this kind of normalization as well. In my experiments   with the Cloudflare Dashboard I was not able to create a zone named `xn--whatever.com` (https://dash.cloudflare.com/api/v4/zones always returns the Unicode name). I assume https://api.cloudflare.com is the same application (modulo access control)?

## Context

I've encountered this issue while obtaining Let's Encrypt certificates for an IDN using [lego](https://github.com/go-acme/lego). lego sanitizes domains even before speaking with the LE servers (which involves `idna.ToASCII`), and passes the sanitized domain to  `ZoneIDByName`.

## References

- https://github.com/go-acme/lego/issues/831#issuecomment-535115322

## Your Environment

- Version used: `github.com/cloudflare/cloudflare-go v0.10.0`, see [go.mod](https://github.com/go-acme/lego/blob/master/go.mod)
- Environment name and version: Go 1.13, Debian 10 (though this shouldn't matter)

dmke:
## Description

Fixes #347. See there for details.

Local tests pass. Integrating in [lego](github.com/go-acme/lego) succeeds.

## Types of changes

-  Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

-  My change requires a change to the documentation.
-  I have updated the documentation accordingly.


virtualdns_test.go
```diff
@@ -21,8 +21,9 @@ func TestVirtualDNSUserAnalytics(t *testing.T) {
 	setup()
 	defer teardown()
 
-	since := time.Now().Add(-1 * time.Hour)
-	until := time.Now()
+	now := time.Now().UTC()
```
jacobbednarz: This isn't related to this change and I generally recommend one discussed change per PR. Would you mind splitting this one off so it gets the proper review it deserves?

zone.go
```diff
@@ -738,3 +740,18 @@ func (api *API) UpdateFallbackOrigin(zoneID string, fbo FallbackOrigin) (*Fallba
...
 }
+
+// normalizeZoneName tries to convert IDNs (international domain names)
+// from Punycode to Unicode form. If the given zone name is not represented
+// as Punycode, or converting fails (for invalid representations), it
+// is returned unchanged.
+//
+// Note: conversion errors are silently discarded.
+func normalizeZoneName(name string) string {
+	if strings.Contains(name, "xn--") {
```
jacobbednarz: I don't think we need to check if the string contains the "xn--" characters. Instead we can run all values through `idna.ToUnicode`.

zone_test.go
```diff
@@ -960,3 +960,35 @@ func TestFallbackOrigin_UpdateFallbackOrigin(t *testing.T) {
 		assert.Equal(t, want, response)
 	}
 }
+
+func Test_normalizeZoneName(t *testing.T) {
+	tests := []struct {
```
jacobbednarz: Nice table tests!

zone.go
dmke: Yeah, you're right. I thougth it might avoid some unnecessary processing overhead, but I should have benchmarked it first.

Here are my current results ([code](https://gist.github.com/dmke/d0ae099ab00873ca5281350bed91a6b3)):

```
ToUnicode(punycode) w/ prefix check       3000000	       472 ns/op
ToUnicode(punycode) w/o prefix check      3000000	       472 ns/op
ToUnicode(unicode) w/ prefix check       200000000	         9.99 ns/op
ToUnicode(unicode) w/o prefix check      200000000	        10.00 ns/op
```

The prefix check saves (at most) a few dozen picoseconds :-)

zone_test.go
dmke: Thanks :-)

virtualdns_test.go
dmke: See #349

jacobbednarz:
cc @zackproser @patryk

patryk:
Let's just fix merge conflicts and it's good to go!

dmke:
Sure, give me a moment :-)

dmke:
I have rebased my changes onto master.

dmke:
@patryk: Thanks for merging!

Not to sound impatient, but can we expect a new release/tag anytime soon? Are you waiting on other PRs, where I might be able to help?

patryk:
We make cloudflare-go releases more or less as needed. If you need an urgent release, let me know.

PR #349, opened 2019-09-27
Fix TZ dependent test case

dmke:
## Description

Running the test-suite in a non-UTC timezone (say in UTC+2), will cause a test to fail.

Yes.

## Types of changes

-  New feature (non-breaking change which adds functionality)
-  Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

-  My change requires a change to the documentation.
-  I have updated the documentation accordingly.

### DIFF
--- a/virtualdns_test.go
+++ b/virtualdns_test.go
@@ -21,8 +21,9 @@ func TestVirtualDNSUserAnalytics(t *testing.T) {
 	setup()
 	defer teardown()
 
-	since := time.Now().Add(-1 * time.Hour)
-	until := time.Now()
+	now := time.Now().UTC()
+	since := now.Add(-1 * time.Hour)
+	until := now
 
 	handler := func(w http.ResponseWriter, r *http.Request) {
 		expectedMetrics := "queryCount,uncachedCount,staleCount,responseTimeAvg,responseTimeMedia,responseTime90th,responseTime99th"


dmke:
Side note: Running [`golangci-lint`](https://github.com/golangci/golangci-lint) against this code base yields a ton of problems, ranging from duplicate code, non `gofmt -s`'ed files, maligned structs, unused variables, etc.

What are the QA requirements for this library? Should I bother to create PRs to cleanup these issues?

jacobbednarz:
Previous maintainers have ensured that things like `golint` pass at a minimum but nothing specific that I’ve found. 

As a general rule of thumb, the PRs I’ve committed and come across follow the [scout rule](https://www.matheus.ro/2017/12/11/clean-code-boy-scout-rule/) of leaving the code you touch in a better state than you found it in. 

I’m personally not a fan of mass cleanups as in some cases, you cause hidden breakages. I prefer to focus on one section at a time, understand the code in question and then refactor/tidy up.

jacobbednarz:
cc @patryk @zackproser

dmke:
> As a general rule of thumb, the PRs I’ve committed and come across follow the scout rule of leaving the code you touch in a better state than you found it in.

Indeed, that's why I'm asking.

> I’m personally not a fan of mass cleanups as in some cases, you cause hidden breakages. I prefer to focus on one section at a time, understand the code in question and then refactor/tidy up.

True. And in some cases, the change is obvious:

```
rate_limiting.go:156:3: assignOp: replace `pageOpts.Page = pageOpts.Page + 1` with `pageOpts.Page++` (gocritic)
                pageOpts.Page = pageOpts.Page + 1
                ^
virtualdns.go:166:18: underef: could simplify (*o.Since).UTC to o.Since.UTC (gocritic)
                v.Set("since", (*o.Since).UTC().Format(time.RFC3339))
                               ^
cmd/flarectl/firewall.go:119:15: Error return value of `errors.Wrap` is not checked (errcheck)
                        errors.Wrap(err, errCreating)
                                   ^
custom_pages.go:71:14: string `accounts` has 3 occurrences, make it a constant (goconst)
                pageType = "accounts"
                           ^
auditlogs.go:109:62: S1025: should use String() instead of fmt.Sprintf (gosimple)
        uri := "/organizations/" + organizationID + "/audit_logs" + fmt.Sprintf("%s", a)
                                                                    ^
auditlogs.go:137:30: S1025: should use String() instead of fmt.Sprintf (gosimple)
        uri := "/user/audit_logs" + fmt.Sprintf("%s", a)
                                    ^
cmd/flarectl/firewall.go:293:9: nilness: impossible condition: nil != nil (govet)
        if err != nil {
               ^
```

patryk:
More linting is something we might explore sometime in the future.

patryk:
Thanks @dmke !

PR #362, opened 2019-10-15
zone: Swap to using subscription API for setting plans

jacobbednarz:
While debugging an [issue in the Terraform provider](https://github.com/terraform-providers/terraform-provider-cloudflare/pull/488) I've come across
a broken endpoint whereby attempting to set a zone's rate plan
silently fails using this method. After a chat with some Cloudflare
Engineers, we're going to move to using the Subscription API instead
which has this functionality and is the intended approach to updating
a zone's rate plan.

### DIFF
--- a/zone.go
+++ b/zone.go
@@ -288,6 +288,14 @@ type FallbackOriginResponse struct {
 	Result FallbackOrigin `json:"result"`
 }
 
+// zoneSubscriptionRatePlanPayload is used to build the JSON payload for
+// setting a particular rate plan on an existing zone.
+type zoneSubscriptionRatePlanPayload struct {
+	RatePlan struct {
+		ID string `json:"id"`
+	} `json:"rate_plan"`
+}
+
 // CreateZone creates a zone on an account.
 //
 // Setting jumpstart to true will attempt to automatically scan for existing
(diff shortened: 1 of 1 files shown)


jacobbednarz:
@martijngonlag would you mind casting your eyes over this and make sure this is what the intended use is?

zone.go
```diff
@@ -486,20 +494,29 @@ func (api *API) ZoneSetVanityNS(zoneID string, ns []string) (Zone, error) {
...
+// "CF_ENT".
+//
+// API reference: https://api.cloudflare.com/#zone-subscription-update-zone-subscription
+func (api *API) ZoneSetPlan(zoneID string, planType string) error {
+	zonePayload := zoneSubscriptionRatePlanPayload{}
+	zonePayload.RatePlan.ID = planType
+
+	uri := fmt.Sprintf("/zones/%s/subscription", zoneID)
+
+	_, err := api.makeRequest("POST", uri, zonePayload)
```
garrettgalow: I think subscriptions will require POST on new zone plans, but PUT for updates once a subscription already exists

garrettgalow: I don't know if this begs the question of whether a subscription is a resource that should be controlled by TF. Seems like overkill, but I'm not sure.

jacobbednarz: > I think subscriptions will require POST on new zone plans, but PUT for updates once a subscription already exists

Gotcha. I might look to also add a method for `ZoneUpdateRatePlan` too. Thoughts?

> I don't know if this begs the question of whether a subscription is a resource that should be controlled by TF. Seems like overkill, but I'm not sure.

It's definitely a fair question and my preference is to have as much as possible available in Terraform where we aren't doing anything that is unmaintainable or difficult to work with and then let the consumer decide how much they put into code. FWIW, we manage all of our infrastructure in a repository of some kind (Terraform, Cloudformation, bash scripts); this just means one less thing we handle outside of Terraform for Cloudflare.

garrettgalow: Having an update makes sense. You will just need the subscription id for that.

Yeah, if the subscription state is kept in the zone object that might work best for now. If at some point we add full support for subscriptions, then we can maybe move it out.

jacobbednarz: Sweet, have added an update method via 2182e7e

PR #367, opened 2019-10-30
Changed the API endpoint used for Worker route deletion

marc-barry:
## Description

Changed the API endpoint used for Worker route deletion to use `zones/:zone_id/workers/routes/:route_id` from https://api.cloudflare.com/#worker-routes-delete-route.

I was trying to use `DeleteWorkerRoute()` but kept encountering the following error:

```
HTTP status 501: content "{\"success\":false,\"errors\":[{\"code\":10000,\"message\":\"API Tokens are not supported by this API for now\"}]}\n"
```

I dug into the code and noticed that this library is using a deprecated API endpoint documented here: https://api.cloudflare.com/#worker-filters-deprecated--properties. This pull request updates the `DeleteWorkerRoute()` function to use https://api.cloudflare.com/#worker-routes-delete-route.

I did notice that there is an attempt in the other `*WorkerRoute()` functions to switch endpoints between the deprecated ones and the currently supported based on some conditions. But I couldn't see an easy way to do this with the `DeleteWorkerRoute()` without adding complication which seemed unnecessary considering that the old API should no longer be used.

I updated the unit tests to use the new route.

N/A

## Types of changes

## Checklist:

### DIFF
--- a/workers.go
+++ b/workers.go
@@ -411,7 +411,7 @@ func formatMultipartBody(params *WorkerScriptParams) (string, []byte, error) {
 
 // CreateWorkerRoute creates worker route for a zone
 //
-// API reference: https://api.cloudflare.com/#worker-filters-create-filter
+// API reference: https://api.cloudflare.com/#worker-filters-create-filter, https://api.cloudflare.com/#worker-routes-create-route
 func (api *API) CreateWorkerRoute(zoneID string, route WorkerRoute) (WorkerRouteResponse, error) {
 	// Check whether a script name is defined in order to determine whether
 	// to use the single-script or multi-script endpoint.
@@ -438,11 +438,9 @@ func (api *API) CreateWorkerRoute(zoneID string, route WorkerRoute) (WorkerRoute
 
 // DeleteWorkerRoute deletes worker route for a zone
 //
-// API reference: https://api.cloudflare.com/#worker-filters-delete-filter
+// API reference: https://api.cloudflare.com/#worker-routes-delete-route
 func (api *API) DeleteWorkerRoute(zoneID string, routeID string) (WorkerRouteResponse, error) {
-	// For deleting a route, it doesn't matter whether we use the
-	// single-script or multi-script endpoint
-	uri := "/zones/" + zoneID + "/workers/filters/" + routeID
+	uri := "/zones/" + zoneID + "/workers/routes/" + routeID
 	res, err := api.makeRequest("DELETE", uri, nil)
 	if err != nil {
 		return WorkerRouteResponse{}, errors.Wrap(err, errMakeRequestError)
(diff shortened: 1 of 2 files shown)


workers.go
```diff
@@ -411,7 +411,7 @@ func formatMultipartBody(params *WorkerScriptParams) (string, []byte, error) {
 
 // CreateWorkerRoute creates worker route for a zone
 //
-// API reference: https://api.cloudflare.com/#worker-filters-create-filter
+// API reference: https://api.cloudflare.com/#worker-filters-create-filter, https://api.cloudflare.com/#worker-routes-create-route
```
marc-barry: Added the non-deprecated API endpoint to the documentation.

workers.go
```diff
@@ -438,11 +438,9 @@ func (api *API) CreateWorkerRoute(zoneID string, route WorkerRoute) (WorkerRoute
 
 // DeleteWorkerRoute deletes worker route for a zone
 //
-// API reference: https://api.cloudflare.com/#worker-filters-delete-filter
+// API reference: https://api.cloudflare.com/#worker-routes-delete-route
```
marc-barry: Updated the documentation to have the correct API reference.

workers.go
```diff
@@ -457,7 +455,7 @@ func (api *API) DeleteWorkerRoute(zoneID string, routeID string) (WorkerRouteRes
 
 // ListWorkerRoutes returns list of worker routes
 //
-// API reference: https://api.cloudflare.com/#worker-filters-list-filters
+// API reference: https://api.cloudflare.com/#worker-filters-list-filters, https://api.cloudflare.com/#worker-routes-list-routes
```
marc-barry: Added the non-deprecated API endpoint to the documentation.

workers.go
```diff
@@ -487,7 +485,7 @@ func (api *API) ListWorkerRoutes(zoneID string) (WorkerRoutesResponse, error) {
 
 // UpdateWorkerRoute updates worker route for a zone.
 //
-// API reference: https://api.cloudflare.com/#worker-filters-update-filter
+// API reference: https://api.cloudflare.com/#worker-filters-update-filter, https://api.cloudflare.com/#worker-routes-update-route
```
marc-barry: Added the non-deprecated API endpoint to the documentation.

jRiest:
Thanks for the PR @marc-barry!

Historically the `routes` endpoints were only available to users on the Cloudflare enterprise plan, which is why the code would default to the `filters` endpoints in the situations where either endpoint could be used. However, now that the `routes` endpoints are available to everyone and the `filters` endpoints have been deprecated, we should do exactly what you're doing here and use the `routes` endpoint whenever possible.

You may run into similar issues using API tokens with the list/create/update route operations if you hit one of the conditions where it would default to the `filters` endpoints, but that should be fixed in #369.

patryk:
Thanks @marc-barry !

PR #379, opened 2019-11-18
Add support for JSON output via --json

mfontani:
## Description

This PR adds a global `--json` support to `flarectl`, which allows the output that's currently output via the `writeTable` function (which is pretty much all of it, aside from the IP list) to be output as JSON instead of as a table.

This ensures that long lines (i.e. the likes of a SPF record) aren't needlessly split on whitespace, as well as allowing the output to be fed to other tools via pipes without having to needlessly parse textual tabular format: a simple `| jq .` will be enough to start the pipe with.

Locally, yes, via something along the lines of:

    CF_DOMAIN=example.com CF_API_KEY="$( /usr/bin/secret-tool lookup cf_api_key "$CF_DOMAIN" )"  CF_API_EMAIL="$( /usr/bin/secret-tool lookup cf_api_email "$CF_DOMAIN" )" go run . --json d l --zone $CF_DOMAIN

I've *not* added any tests, as there weren't any in the flarectl directory.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/flarectl/dns.go
+++ b/cmd/flarectl/dns.go
@@ -57,7 +57,7 @@ func dnsCreate(c *cli.Context) {
 		formatDNSRecord(resp.Result),
 	}
 
-	writeTable(output, "ID", "Name", "Type", "Content", "TTL", "Proxiable", "Proxy", "Locked")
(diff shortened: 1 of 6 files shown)


patryk:
Nice addition, thanks @mfontani !

PR #386, opened 2019-11-29
Return proper error code on failure #373

### LINKED ISSUE #373: Return non-zero exit code on incorrect usage
Running an incorrect command, such as `flarectl zone purge --xyz` should lead to a non-zero exit code for scripting.

## Expected behavior

`flarectl zone purge --xyz && echo "failed"` should print `failed`

## Current behavior

Running `flarectl zone purge --xyz && echo "purged"` prints 
```
Incorrect Usage: flag provided but not defined: -xyz

NAME:
   flarectl zone purge - (Selectively) Purge the cache for a zone

USAGE:
   flarectl zone purge [command options] [arguments...]

OPTIONS:
   --zone value   zone name
   --everything   purge everything from cache for the zone
   --hosts value  a list of hostnames to purge the cache for
   --tags value   the cache tags to purge (Enterprise only)
   --files value  a list of [exact] URLs to purge
   
purged
```

It is also somewhat confusing that the error message includes one hyphen instead of two.

## Context

We have a script that purges the cache for many files. Based upon linter recommendations, a shell variable was accidentally quoted when it should not have been (thus running a command like  `flarectl zone purge "--files http://1 --files http://2"` instead of `flarectl zone purge --files http://1 --files http://2`. Since the script runs inside a container of a kubernetes pod, a non-zero exit code would have been reported. Instead, the error was silently discarded.

## Your Environment

the bug in:

* Version used (e.g. git sha or specific tag): flarectl version 2017.10.0
* Environment name and version (e.g. Go 1.11, MacOS 10.14.1): Go 1.10, Google Cloud COS (no clear version)

mfontani:
A linter also complained about the Run() method not being error checked:

    [gometalinter][W][] error return value not checked (app.Run(os.Args)) (errcheck)

By checking the error returned, we can ensure the user gets a non-zero
return code on argument parse failure, as well as on command failure,
both of which are instances where returning a non-zero error code seems
like a good idea.

Closes/Fixes #373

I would tentatively consider this a "breaking change" in that it affects scripts which have, so far, interacted with the command line without having to bother about error checking, rather than a "bug fix" (which it sort of is, as it's a bug IMVHO that it _didn't_ return a non-zero exit code on failure).

Yup, locally.

Also, given the exact invocation given on #373 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/flarectl/flarectl.go
+++ b/cmd/flarectl/flarectl.go
@@ -672,5 +672,9 @@ func main() {
 			},
 		},
 	}
-	app.Run(os.Args)
+	err := app.Run(os.Args)
+	if err != nil {
+		os.Exit(1)
+	}
+	os.Exit(0)
 }


patryk:
Thanks @mfontani ! I agree to treat it as a bug. Command line tools should behave unix-like, return "failed" error code when they didn't succeed. Thanks again!

PR #387, opened 2019-12-01
Fix unmarshalling of OriginCACertificate

### LINKED ISSUE #190: JSON Unmarshall error when using CreateOriginCertificate
Hello,
Upon requesting to Create new CA certificate I always get this: 

`error unmarshalling the JSON response: parsing time ""2019-06-19 11:19:00 +0000 UTC"" as ""2006-01-02T15:04:05Z07:00"": cannot parse " 11:19:00 +0000 UTC"" as "T"`

I have verified that I submit all parameters correctly. 
Thanks

player-two:
Fixes #190 

## Description

The format of the `expires_on` field differs from what is documented, and the json parser makes `CreateOriginCertificate()` always return an error, even if the certificate was created.  This change introduces custom timestamp parsing that works for both the actual and documented formats.

I wrote a tiny test program that calls `CreateOriginCertificate()` and checked that the cert object is populated and the error is `nil`.

N/A

## Types of changes

## Checklist:

### DIFF
--- a/origin_ca.go
+++ b/origin_ca.go
@@ -34,6 +34,7 @@ type originCACertificateResult struct {
 	CSR             string    `json:"csr,omitempty"`
 }
 
+// UnmarshalJSON handles custom parsing from an API response to an OriginCACertificate
 func (c *OriginCACertificate) UnmarshalJSON(data []byte) error {
 	var res originCACertificateResult
 	err := json.Unmarshal(data, &res)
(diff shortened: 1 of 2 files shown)


player-two:
The other way to fix this is introduce a new type for just the field and modify the struct definition to 
```go
type OriginCACertificate struct {
	ExpiresOn certificateExpiresOn `json:"expires_on,omitempty"`
```
That would remove the need to overload `MarshalJSON()` for the struct, but it would mean any code accessing the field would have to explicitly convert to `time.Time`.  I opted to introduce a little more mess here to make usage easier.

patryk:
Thanks @player-two ! I have modified it slightly with help of the aliasing method described in http://choly.ca/post/go-json-marshalling/

PR #392, opened 2019-12-05
Add RevokedAt field to OriginCACertificate

player-two:
## Description

Given a cert ID, I want to know if it has been revoked or not.  The list call already filters out revoked certs, but that could require multiple calls to traverse pages when looking for the absence of an ID.

This field was originally added in #387 and pulled out (guessing unintentionally), which is fine because it wasn't very obvious nor related to the rest of that change.

Just via a small script that prints the struct after revoking and re-fetching by the ID

## Types of changes

## Checklist:


patryk:
Ops! I didn't notice you added new field, sorry!

player-two:
No problem at all!

PR #389, opened 2019-12-03
waf: read all pages for ListWAFRules, ListWAFGroups and ListWAFPackages

XaF:
## Description

The ListWAF* methods were not following the pagination, and just returning the first page of the results. Inspired from the `dns.go` file, I added the pagination support for those methods in order to list all the WAF rules, groups and packages when calling the corresponding methods.

## Has your change been tested?

Tests added to verify that the list is processed properly in all cases.

## Screenshots (if appropriate):

Screen Shot 2019-12-03 at 5 30 13 PM

## Types of changes

What sort of change does your code introduce/modify?

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

[1]: https://help.github.com/articles/closing-issues-using-keywords/

PR #377, opened 2019-11-13
Increase per_page for DNS records up to 100

Constantin07:
This PR updates the per_page max value to 100 as per docs here https://api.cloudflare.com/#dns-records-for-a-zone-list-dns-records

## Description

Reason: making lesser API requests to REST API to avoid `Throttling: Rate exceeded` error messages when listing very large DNS zones.

## Has your change been tested?

```
$ go test -v
=== RUN   TestAccessApplications
--- PASS: TestAccessApplications (0.00s)
=== RUN   TestAccessApplication
--- PASS: TestAccessApplication (0.00s)
=== RUN   TestCreateAccessApplications
--- PASS: TestCreateAccessApplications (0.00s)
=== RUN   TestUpdateAccessApplication
--- PASS: TestUpdateAccessApplication (0.00s)
=== RUN   TestUpdateAccessApplicationWithMissingID
--- PASS: TestUpdateAccessApplicationWithMissingID (0.00s)
=== RUN   TestDeleteAccessApplication
--- PASS: TestDeleteAccessApplication (0.00s)
=== RUN   TestRevokeAccessApplicationTokens
--- PASS: TestRevokeAccessApplicationTokens (0.00s)
=== RUN   TestAccessGroups
--- PASS: TestAccessGroups (0.00s)
=== RUN   TestAccessGroup
--- PASS: TestAccessGroup (0.00s)
=== RUN   TestCreateAccessGroup
--- PASS: TestCreateAccessGroup (0.00s)
=== RUN   TestUpdateAccessGroup
--- PASS: TestUpdateAccessGroup (0.00s)
=== RUN   TestUpdateAccessGroupWithMissingID
--- PASS: TestUpdateAccessGroupWithMissingID (0.00s)
=== RUN   TestDeleteAccessGroup
--- PASS: TestDeleteAccessGroup (0.00s)
=== RUN   TestAccessIdentityProviders
--- PASS: TestAccessIdentityProviders (0.00s)
=== RUN   TestAccessIdentityProviderDetails
--- PASS: TestAccessIdentityProviderDetails (0.00s)
=== RUN   TestCreateAccessIdentityProvider
--- PASS: TestCreateAccessIdentityProvider (0.00s)
=== RUN   TestUpdateAccessIdentityProvider
--- PASS: TestUpdateAccessIdentityProvider (0.00s)
=== RUN   TestDeleteAccessIdentityProvider
--- PASS: TestDeleteAccessIdentityProvider (0.00s)
[...]

Issue #324, opened 2019-07-09
Add context to API methods


johanbrandhorst:
## Expected behavior

When making API requests it is generally good to be able to cancel requests. In Go this is done via `context.Context` parameters. I would like to be able to cancel requests.

## Current behavior

There is no way to cancel pending requests.

## Possible solution

Either change all the methods to take a context (probably undesirable), or add `XContext` methods with the same parameters as `X` but taking a context (see ListZonesContext).

johanbrandhorst:
This is still needed

johanbrandhorst:
This is still needed.

jacobbednarz:
released in [v0.14.0](https://github.com/cloudflare/cloudflare-go/releases/tag/v0.14.0)

PR #394, opened 2019-12-11
Add pagination for listing KV worker namespaces

jbergknoff-rival:
## Description

Listing KV worker namespaces only brings back the first page of (20) results. While that's the default quota for a CF account, it is possible to have more than 20 namespaces, in which case the behavior is inadequate. We're seeing issues with this downstream in the Cloudflare Terraform provider, which is incorrectly trying to create a namespace that already exists (on the second page of results).

This PR updates `ListWorkersKVNamespaces` to return all namespaces by paging through the result set. This was modeled after the work done in ~~#288 and #377~~ #389 .

This changes the signature of the `ListWorkersKVNamespaces` function. Does that require a change to the docs? Where would that be made?

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change) (**changes the signature of `ListWorkersKVNamespaces`**)

## Checklist:

### DIFF
--- a/workers_kv.go
+++ b/workers_kv.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"sync"
 
 	"github.com/pkg/errors"
 )
@@ -83,18 +84,53 @@ func (api *API) CreateWorkersKVNamespace(ctx context.Context, req *WorkersKVName
 //
 // API reference: https://api.cloudflare.com/#workers-kv-namespace-list-namespaces
 func (api *API) ListWorkersKVNamespaces(ctx context.Context) (ListWorkersKVNamespacesResponse, error) {
-	uri := fmt.Sprintf("/accounts/%s/storage/kv/namespaces", api.AccountID)
+	uri := fmt.Sprintf("/accounts/%s/storage/kv/namespaces?per_page=100", api.AccountID)
 	res, err := api.makeRequestContext(ctx, http.MethodGet, uri, nil)
 	if err != nil {
 		return ListWorkersKVNamespacesResponse{}, errors.Wrap(err, errMakeRequestError)
 	}
 
-	result := ListWorkersKVNamespacesResponse{}
-	if err := json.Unmarshal(res, &result); err != nil {
-		return result, errors.Wrap(err, errUnmarshalError)
+	overallResult := ListWorkersKVNamespacesResponse{}
+	if err := json.Unmarshal(res, &overallResult); err != nil {
+		return overallResult, errors.Wrap(err, errUnmarshalError)
 	}
 
-	return result, err
+	totalPageCount := overallResult.ResultInfo.TotalPages
+	var wg sync.WaitGroup
+	wg.Add(totalPageCount - 1)
+	errc := make(chan error)
+
+	for i := 2; i <= totalPageCount; i++ {
+		go func(pageNumber int) error {
+			res, err = api.makeRequestContext(ctx, http.MethodGet, fmt.Sprintf("%s&page=%d", uri, pageNumber), nil)
+			if err != nil {
+				errc <- err
+			}
+
+			pageResult := ListWorkersKVNamespacesResponse{}
+			err = json.Unmarshal(res, &pageResult)
+			if err != nil {
+				errc <- err
+			}
+
+			for _, namespace := range pageResult.Result {
+				overallResult.Result = append(overallResult.Result, namespace)
+			}
+
+			select {
+			case err := <-errc:
+				return err
+			default:
+				wg.Done()
+			}
+
+			return nil
+		}(i)
+	}
+
+	wg.Wait()
+
+	return overallResult, nil
 }
 
 // DeleteWorkersKVNamespace deletes the namespace corresponding to the given ID
--- a/workers_kv.go
+++ b/workers_kv.go
@@ -83,7 +83,7 @@ func (api *API) CreateWorkersKVNamespace(ctx context.Context, req *WorkersKVName
 // ListWorkersKVNamespaces lists storage namespaces
 //
 // API reference: https://api.cloudflare.com/#workers-kv-namespace-list-namespaces
-func (api *API) ListWorkersKVNamespaces() ([]WorkersKVNamespace, error) {
+func (api *API) ListWorkersKVNamespaces(ctx context.Context) ([]WorkersKVNamespace, error) {
 	v := url.Values{}
 	v.Set("per_page", "100")
 
@@ -93,7 +93,7 @@ func (api *API) ListWorkersKVNamespaces() ([]WorkersKVNamespace, error) {
 	for {
 		v.Set("page", strconv.Itoa(page))
 		uri := fmt.Sprintf("/accounts/%s/storage/kv/namespaces?%s", api.AccountID, v.Encode())
-		res, err := api.makeRequest(http.MethodGet, uri, nil)
+		res, err := api.makeRequestContext(ctx, http.MethodGet, uri, nil)
 		if err != nil {
 			return []WorkersKVNamespace{}, errors.Wrap(err, errMakeRequestError)
 		}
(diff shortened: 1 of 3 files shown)


workers_kv.go
```diff
@@ -83,18 +84,53 @@ func (api *API) CreateWorkersKVNamespace(ctx context.Context, req *WorkersKVName
 //
 // API reference: https://api.cloudflare.com/#workers-kv-namespace-list-namespaces
 func (api *API) ListWorkersKVNamespaces(ctx context.Context) (ListWorkersKVNamespacesResponse, error) {
-	uri := fmt.Sprintf("/accounts/%s/storage/kv/namespaces", api.AccountID)
+	uri := fmt.Sprintf("/accounts/%s/storage/kv/namespaces?per_page=100", api.AccountID)
```
jbergknoff-rival: Per #377, increasing this to reduce the number of API calls made.

workers_kv.go
```diff
@@ -83,18 +84,53 @@ func (api *API) CreateWorkersKVNamespace(ctx context.Context, req *WorkersKVName
...
 	res, err := api.makeRequestContext(ctx, http.MethodGet, uri, nil)
 	if err != nil {
 		return ListWorkersKVNamespacesResponse{}, errors.Wrap(err, errMakeRequestError)
 	}
 
-	result := ListWorkersKVNamespacesResponse{}
-	if err := json.Unmarshal(res, &result); err != nil {
-		return result, errors.Wrap(err, errUnmarshalError)
+	overallResult := ListWorkersKVNamespacesResponse{}
```
jbergknoff-rival: This function returns a `ListWorkersKVNamespacesResponse` as opposed to, for example, `[]WorkersKVNamespace`. This makes it awkward to return the full list. In the current changeset, I'm hijacking the first `ListWorkersKVNamespacesResponse` and augmenting its `Result` list, but leaving `ResultInfo` and `Response` alone.

Would it make more sense to move the paging into a separate method which returns a `[]WorkersKVNamespace`.

workers_kv.go
```diff
@@ -83,18 +84,53 @@ func (api *API) CreateWorkersKVNamespace(ctx context.Context, req *WorkersKVName
...
+		return overallResult, errors.Wrap(err, errUnmarshalError)
 	}
 
-	return result, err
+	totalPageCount := overallResult.ResultInfo.TotalPages
+	var wg sync.WaitGroup
+	wg.Add(totalPageCount - 1)
+	errc := make(chan error)
+
+	for i := 2; i <= totalPageCount; i++ {
```
jbergknoff-rival: In #288, we seem to duplicate the request for the first page of zones. Maybe there's a reason for that? Here, we don't duplicate the request, so I'm calling that out.

workers_kv.go
```diff
@@ -83,18 +84,53 @@ func (api *API) CreateWorkersKVNamespace(ctx context.Context, req *WorkersKVName
...
+				errc <- err
+			}
+
+			for _, namespace := range pageResult.Result {
+				overallResult.Result = append(overallResult.Result, namespace)
+			}
+
+			select {
+			case err := <-errc:
+				return err
```
jbergknoff-rival: This error handling is modeled on #288. I don't do much work in golang, so please advise if this is inadequate. I'm not sure that these errors are actually being consumed in any way.

workers_kv.go
```diff
@@ -83,18 +84,53 @@ func (api *API) CreateWorkersKVNamespace(ctx context.Context, req *WorkersKVName
...
+				wg.Done()
+			}
+
+			return nil
+		}(i)
+	}
+
+	wg.Wait()
+
+	return overallResult, nil
```
jbergknoff-rival: Previously was doing `return result, err`. To my eye, any error would have been returned prior to reaching that point, so this returns `nil` for the error. Is that a mistake?

workers_kv.go
jacobbednarz: I'm fine with this for now but if we get any more query parameters, we are going to need to swap to using `url.Values` to ensure we don't thrash the others.

workers_kv.go
jacobbednarz: Yeah, I would expect this to match the other methods like it and return `[]WorkersKVNamespace`. I'm not sure if we have use for the `ListWorkersKVNamespacesResponse` as is.

workers_kv.go
jacobbednarz: If I recall, it was for requests that only had a single page which wasn't getting returned otherwise. Memory is fuzzy though. An approach like #389 might be good to follow here. We can get #288 updated in another PR.

workers_kv.go
```diff
@@ -83,18 +84,53 @@ func (api *API) CreateWorkersKVNamespace(ctx context.Context, req *WorkersKVName
...
 
-	return result, err
+	totalPageCount := overallResult.ResultInfo.TotalPages
+	var wg sync.WaitGroup
+	wg.Add(totalPageCount - 1)
+	errc := make(chan error)
+
+	for i := 2; i <= totalPageCount; i++ {
+		go func(pageNumber int) error {
+			res, err = api.makeRequestContext(ctx, http.MethodGet, fmt.Sprintf("%s&page=%d", uri, pageNumber), nil)
```
jacobbednarz: I don't think we really need the `makeRequestContext` here and `makerequest` should suffice.

workers_kv.go
jbergknoff-rival: Would it make sense to have that breaking change in this PR?

workers_kv.go
```diff
@@ -82,19 +83,39 @@ func (api *API) CreateWorkersKVNamespace(ctx context.Context, req *WorkersKVName
 // ListWorkersKVNamespaces lists storage namespaces
 //
 // API reference: https://api.cloudflare.com/#workers-kv-namespace-list-namespaces
-func (api *API) ListWorkersKVNamespaces(ctx context.Context) (ListWorkersKVNamespacesResponse, error) {
-	uri := fmt.Sprintf("/accounts/%s/storage/kv/namespaces", api.AccountID)
-	res, err := api.makeRequestContext(ctx, http.MethodGet, uri, nil)
-	if err != nil {
-		return ListWorkersKVNamespacesResponse{}, errors.Wrap(err, errMakeRequestError)
-	}
+func (api *API) ListWorkersKVNamespaces() ([]WorkersKVNamespace, error) {
```
jbergknoff-rival: I've changed the signature of this function, which is a breaking change for the library, after the discussion on the first iteration of the PR. Is this acceptable? Now this function is a bit unique in workers_kv.go, but it's a cleaner implementation than my first attempt, and, IMO, more useful functionality for most users of the library.

workers_kv.go
```diff
@@ -82,19 +83,39 @@ func (api *API) CreateWorkersKVNamespace(ctx context.Context, req *WorkersKVName
...
-	result := ListWorkersKVNamespacesResponse{}
-	if err := json.Unmarshal(res, &result); err != nil {
-		return result, errors.Wrap(err, errUnmarshalError)
+	var namespaces []WorkersKVNamespace
+	page := 1
+
+	for {
+		v.Set("page", strconv.Itoa(page))
+		uri := fmt.Sprintf("/accounts/%s/storage/kv/namespaces?%s", api.AccountID, v.Encode())
+		res, err := api.makeRequest(http.MethodGet, uri, nil)
```
jbergknoff-rival: I've replaced `api.makeRequestContext` for `api.makeRequest` per the comments on the first version of this PR. I also dropped context as an argument to this function.

patryk: I would keep context here, we might want to eventually make all API calls context-aware (issue #324).

workers_kv.go
patryk: I'd keep context, reasoning in comment below. Changing return type is fine.

jbergknoff-rival:
Thanks for taking a look, @patryk. I've reinstated the request context.

patryk:
Thanks!

PR #395, opened 2019-12-11
update SpectrumApplication to support string value for proxy_protocol

acmanderson:
## Description

Spectrum recently deprecated use of the `spp` field and the boolean value for the `proxy_protocol` field in its configuration API in favor of a string enum for the `proxy_protocol` field. This change updates the `SpectrumApplication` struct to support reading deprecated values from the API and writing with the updated schema.

- Breaking change: the type of the `ProxyProtocol` field on the `SpectrumApplication` struct has changed from a `bool` to a `string`. Clients using this library that manipulate the `spp`/`proxy_protocol` fields will need to update to use the new schema upon upgrading to the latest version of the library. 
- Adds the `SPP` boolean field to the `SpectrumApplication` struct, will be removed following the EOL date for support for the `proxy_protocol` boolean and the `spp` field.

Added automated test for validating new deserializing logic for `spp` and `proxy_protocol` fields, as well as manual testing against the production API.

## Types of changes

## Checklist:

### DIFF
--- a/spectrum.go
+++ b/spectrum.go
@@ -8,6 +8,31 @@ import (
 	"github.com/pkg/errors"
 )
 
+// ProxyProtocol implements json.Unmarshaler in order to support deserializing of the deprecated boolean
+// value for `proxy_protocol`
+type ProxyProtocol string
+
+func (p *ProxyProtocol) UnmarshalJSON(data []byte) error {
+	var raw interface{}
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return err
+	}
+
+	switch pp := raw.(type) {
+	case string:
+		*p = ProxyProtocol(pp)
+	case bool:
+		if pp {
+			*p = "v1"
+		} else {
+			*p = "off"
+		}
+	default:
+		return fmt.Errorf("invalid type for proxy_protocol field: %T", pp)
+	}
+	return nil
+}
+
 // SpectrumApplication defines a single Spectrum Application.
 type SpectrumApplication struct {
 	ID            string                        `json:"id,omitempty"`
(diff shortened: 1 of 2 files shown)


jacobbednarz:
I wonder if we can just skip straight to the desired enum values for `proxy_protocol` and disregard the need for accounting for `spp` 🤔 We can cut a new major for this library and then change tools like the terraform provider to have the correct values.

acmanderson:
> I wonder if we can just skip straight to the desired enum values for `proxy_protocol` and disregard the need for accounting for `spp` 🤔 We can cut a new major for this library and then change tools like the terraform provider to have the correct values.

The reason I included `spp` was that it's being folded into the `proxy_protocol` field as an enum option instead of its own field, but we're still supporting the old values in the API until the EOL date has passed. In circumstances where an application had `"proxy_protocol": false` and `"spp": true`, the logic in this PR would incorrectly set `ProxyProtocol` to `"off"` instead of `"simple"` when updating the application (see [API reference documentation](https://developers.cloudflare.com/spectrum/api-reference/) for enum values). Let me know if I can make that more clear.

I have an update ready to go for the Terraform provider pending a new `cloudflare-go` release, are there other tools I should be aware of?

pouysorvanndet:
[](url) Android/AppLinkPasteboardSample/build.xml

patryk:
If it's a breaking change then we shouldn't add fields, just to have them removed soon later. Just implement new behaviour already.

acmanderson:
Thanks for the feedback, updated to remove the SPP field.

patryk:
All good, thanks.

PR #400, opened 2019-12-19
api+flarectl: allow updating dns record type

mfontani:
## Description

The Cloudflare REST API apparently allows one to change both a record's usual properties, such as the content; but also the record's Type.  This wasn't reflected in this repository's API subroutine which implements the `UpdateDNSRecord` call.   Consequently, it wasn't implemented in the `flarectl dns update ...` command, either.

It now is.

This allows one to swiftly change a record's _value_, _type_ and _proxy status_ in one go. Particularly useful (my own use case) when wanting to migrate an `A` record to a `CNAME` when implementing a Cloudflare Load Balancer on a domain.

Locally, as shown in the commit messages.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/flarectl/dns.go
+++ b/cmd/flarectl/dns.go
@@ -140,6 +140,7 @@ func dnsUpdate(c *cli.Context) {
 	zone := c.String("zone")
 	recordID := c.String("id")
 	name := c.String("name")
+	rtype := c.String("type")
 	content := c.String("content")
 	ttl := c.Int("ttl")
 	proxy := c.Bool("proxy")
@@ -153,6 +154,7 @@ func dnsUpdate(c *cli.Context) {
 	record := cloudflare.DNSRecord{
 		ID:      recordID,
 		Name:    name,
+		Type:    strings.ToUpper(rtype),
 		Content: content,
 		TTL:     ttl,
 		Proxied: proxy,
--- a/cmd/flarectl/flarectl.go
+++ b/cmd/flarectl/flarectl.go
@@ -328,6 +328,10 @@ func main() {
 							Name:  "content",
 							Usage: "record content",
 						},
+						cli.StringFlag{
+							Name:  "type",
+							Usage: "record type",
+						},
 						cli.IntFlag{
 							Name:  "ttl",
 							Usage: "TTL (1 = automatic)",
--- a/dns.go
+++ b/dns.go
@@ -141,7 +141,9 @@ func (api *API) UpdateDNSRecord(zoneID, recordID string, rr DNSRecord) error {
 	if rr.Name == "" {
 		rr.Name = rec.Name
 	}
-	rr.Type = rec.Type
+	if rr.Type == "" {
+		rr.Type = rec.Type
+	}
 	uri := "/zones/" + zoneID + "/dns_records/" + recordID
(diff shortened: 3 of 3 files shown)


jacobbednarz:
I have some reservations about the edge cases this will introduce (code itself is fine). The largest being if someone goes from (what I'll dub) a simple record configuration (A/CNAME) to a complex record configuration (SRV/LOC/DNSKEY) thinking this will be a straight swap and having them left with incomplete or broken records.

There are explicit rules in the Terraform provider [to force recreation of the rule](https://github.com/terraform-providers/terraform-provider-cloudflare/blob/8387ec354dfef2db334c3cd32cd2a68fa73b3445/cloudflare/resource_cloudflare_record.go#L47-L51) when the type changes to mitigate a bunch of these scenarios.

We can take two approaches here:

- Don't ship this which forces the longer and more verbose approach to change the type however it does prevent people getting stuck in weird spots; or
- Ship this with the knowledge that it's a potential for edge cases and those edge cases won't all be supported and work out of the box.

Thoughts @patryk?

mfontani:
My 2c: 2, then 1.

I'm sure I'm not the only one who's had to use another tool in order to reliably change an A record into a CNAME, so having this feature available at all is IMVHO more important than the feature going as far as disallowing the user shooting themselves in the foot.

As far as that goes, these couple commits can easily be merged as is, and if you provide me with a link or something as to which switches ("from" type to "to" type) should be disallowed, I'm more than happy to have a go at it and provide a pull request for it once done.

I've had to use my own branch in order to reliably go and switch quite a few records in these past few days, and I'm happy I didn't have to use curl for it. I still had that chill and anticipation of "what if I screwed the ID up, and ended up changing a TXT instead of the A record" - but the onus was on _me_, the user, to get it right.   If the tool can stop common mistakes, it's IMVHO a "nice to have".

patryk:
Originally `ForceNew` was introduced in https://github.com/terraform-providers/terraform-provider-cloudflare/commit/7c71b2ad77bcddf5015f612847573b967424ce48 (over 3 years ago). It is entirely possible that API allows changing types in-place now.

The API code always performs request validation first, then executes. So I doubt that it would leave the record in broken state.

I think we're fine with this change, it's a bit niche use case, but ultimately should be harmless. We'll keep the current behaviour in Terraform, though.

pouysorvanndet:
> I have some reservations about the edge cases this will introduce (code itself is fine). The largest being if someone goes from (what I'll dub) a simple record configuration (A/CNAME) to a complex record configuration (SRV/LOC/DNSKEY) thinking this will be a straight swap and having them left with incomplete or broken records.
> 
> There are explicit rules in the Terraform provider [to force recreation of the rule](https://github.com/terraform-providers/terraform-provider-cloudflare/blob/8387ec354dfef2db334c3cd32cd2a68fa73b3445/cloudflare/resource_cloudflare_record.go#L47-L51) when the type changes to mitigate a bunch of these scenarios.
> 
> We can take two approaches here:
> 
> * Don't ship this which forces the longer and more verbose approach to change the type however it does prevent people getting stuck in weird spots; or
> * Ship this with the knowledge that it's a potential for edge cases and those edge cases won't all be supported and work out of the box.
> 
> Thoughts @patryk?

PR #403, opened 2020-01-07
access_identity_provider: Consolidate configuration fields

jacobbednarz:
While looking integration Access Identity Providers into Terraform, I
found that the configuration fields are super awkward to work with as we
either:

- Disregard Go's type system and make everything an `interface`; or
- Restrict the ability to use configuration options in a single step as
  we don't know what Type the incoming data actually is.

To rectify this, I've consolidated all the fields into a single struct
and move the onus to the end user to correctly define the fields that
are required. A nice side effect of this is that the fields aren't
repeated for each configuration and reduces the toil if/when a field is
updated.

### DIFF
--- a/access_identity_provider.go
+++ b/access_identity_provider.go
@@ -9,192 +9,36 @@ import (
 
 // AccessIdentityProvider is the structure of the provider object.
 type AccessIdentityProvider struct {
-	ID     string      `json:"id,omitemtpy"`
-	Name   string      `json:"name"`
-	Type   string      `json:"type"`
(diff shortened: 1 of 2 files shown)


jacobbednarz:
We'll let this sit while I do some POC integrations before we commit to this and find out this introduces other issues.

jacobbednarz:
@patryk confirmed the API and we're ready for this to land 🛳️

patryk:
Awesome, thanks!

PR #404, opened 2020-01-08
Support WAF overrides

Evesy:
Adds support for [WAF Overrides](https://api.cloudflare.com/#waf-overrides-properties)

## Description

Cloudflare recently introduced the `firewall/waf/overrides` API endpoint to create per URL overrides for WAF rule/group configurations

Needed for https://github.com/terraform-providers/terraform-provider-cloudflare/issues/554

- Tests have been added
- Manual testing performed (Listing, creating, updating & deleting a WAF Override

(`rewrite_actions` are not currently included in any test cases for creating/updating as there is currently a bug in which the API object returned does not include any `rewrite_actions`, even if they were specified at creation. Ticket raised with Cloudflare support.)

## Types of changes

## Checklist:


Evesy:
`rewrite_actions` is down to some inconsistencies between the docs and actual API fields.

Docs: `rewrite_actions`
Actual: `rewrite_action`

Instead of a 400 when specifying the incorrect plural notation the CF API silently drops that data. Will cleanup shortly

jacobbednarz:
Could we split this into it’s own file instead (if it’s not entangled with the generic WAF stuff)? I know you were just following the convention here but this file is getting a bit large.

Evesy:
No worries, I've split it out into its own file and also updated to fix `rewrite_action`

waf_overrides.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	var r WAFOverridesResponse
+	err = json.Unmarshal(res, &r)
+	if err != nil {
+		return []WAFOverride{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	if !r.Success {
+		// TODO: Provide an actual error message instead of always returning nil
```
jacobbednarz:
```suggestion
```

I think we're good to remove this. It's consistent with other methods as it is.

waf_overrides.go
```diff
@@ -0,0 +1,136 @@
...
+
+	for ri := range r.Result {
+		overrides = append(overrides, r.Result[ri])
+	}
+	return overrides, nil
+}
+
+// WAFOverride returns a WAF override from the given override ID.
+//
+// API Reference: https://api.cloudflare.com/#waf-overrides-update-uri-controlled-waf-configuration
```
jacobbednarz:
```suggestion
// API Reference: https://api.cloudflare.com/#waf-overrides-uri-controlled-waf-configuration-details
```

waf_overrides.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return WAFOverride{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return r.Result, nil
+}
+
+// DeleteWAFOverride deletes a WAF override for a zone.
+//
+// API reference: https://api.cloudflare.com/#waf-overrides-delete-lockdown-rule
```
jacobbednarz: I've raised a ticket with Cloudflare about fixing this URL.

jacobbednarz:
Couple of very minor nitpicks but happy overall. Thanks @Evesy !

waf_overrides.go
patryk: Let's keep the TODO for now, because we do actually silently drop the error message here. We should wrap messages returned in `r.Errors` here.

Evesy:
Updated API link (except for `Delete` until the docs are updated)

Left the TODO in place, but happy to do either way, just let me know

patryk:
Thanks @Evesy !

PR #406, opened 2020-01-10
Read correct AccountMember 2FA attribute

calderonth:
The value for the 2FA attribute wasn't correctly read and seemed to always return false, I assume that it is because it wasn't processed from the JSON output.
This should fix it.


patryk:
Thanks @calderonth !

PR #415, opened 2020-01-23
auditlogs: fix GetOrganizationAuditLogs()

sean-:
Move to using stdlib strategies for encoding URL parameters.

Since this endpoint was broken and not usable.  Introduce a breaking change by removing `AuditLogFilter.String()` and rename to `ToQuery()`.  The leading slash is now no longer provided by `ToQuery()`.

`GetOrganizationAuditLogs()` was assuming it needed to base64 decode the response, but the response was a JSON payload and did not need to be decoded.

## Types of changes

### DIFF
--- a/auditlogs.go
+++ b/auditlogs.go
@@ -1,9 +1,10 @@
 package cloudflare
 
 import (
-	"encoding/base64"
 	"encoding/json"
-	"fmt"
+	"net/url"
+	"path"
+	"strconv"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


auditlogs.go
```diff
@@ -106,17 +109,16 @@ func (a AuditLogFilter) String() string {
 //
 // API Reference: https://api.cloudflare.com/#audit-logs-list-organization-audit-logs
 func (api *API) GetOrganizationAuditLogs(organizationID string, a AuditLogFilter) (AuditLogResponse, error) {
-	uri := "/organizations/" + organizationID + "/audit_logs" + fmt.Sprintf("%s", a)
-
-	res, err := api.makeRequest("GET", uri, nil)
-	if err != nil {
-		return AuditLogResponse{}, err
+	uri := url.URL{
+		Path:       path.Join("/organizations", organizationID, "audit_logs"),
```
jacobbednarz: Organization endpoints are deprecated and `/organization` should be replaced with `/accounts` instead. We should get the API documentation updated to reflect this a well.

sean-: Offf, good to know.  I'll switch to the preferred API.  How can I get audit logs via `/accounts`?  It looks like I'm missing something in the API.

Regardless, until then, this at least fixes something that was abjectly broken.

jacobbednarz: the `s/organization/account` swap is a 1:1 mapping from my understanding. Same structure, payloads, etc. give it a whirl and confirm it works as other parts of the API have been updated to reflect this (example https://api.cloudflare.com/#organizations-properties)

i'm not certain on this one but if it works, we can get in touch with the API team to update the documentation.

sean-: I have no idea how I was supposed to know that, but you're right!, it works!  PR updated.

Thank you!

jacobbednarz: unfortunately it is a little buried but that is what code review is for 😄

patryk:
Thanks!

PR #418, opened 2020-02-11
feat: add ability to create/list/update/delete User Tokens

UrosSimovic:


jacobbednarz:
Thanks for the PR @UrosSimovic! It's appreciated!

We generally don't add functionality that isn't publicly documented at api.cloudflare.com or developers.cloudflare.com to ensure we are building against stable endpoints. Are you able to please raise a ticket with Cloudflare Support to get this work looked at so we can accept these changes?

UrosSimovic:
Of course. Will do that.

garrettgalow:
Creating API tokens via API key are not supported. While you may be able to reverse engineer it, that behavior will be changed in the future. We are working on supporting creating tokens with tokens via API

jacobbednarz:
Thanks for the confirmation @garrettgalow 🙇‍♂️

In that case @UrosSimovic, I"m going to have to decline the PR as it's not going to provide a stable, maintained experience and it is using an unsupported API. Appreciate your effort here and we can revisit when the API supports this as a first class citizen.

patryk:
Thank you for your effort, @UrosSimovic, but as @garrettgalow explained, the API is not stable and currently unsupported. We'll return to it when it's ready.

UrosSimovic:
Of course. Not a problem. 
Thanks for cooperation.

PR #419, opened 2020-02-13
access_group/access_policy: Add support for non-identity clients

jacobbednarz:
Updates the two places where we can provide a non-identity configuration
(access group and access policies) to support using them.


patryk:
Looks good, thanks!

PR #432, opened 2020-03-25
Remove duplicated Access Policy structs

jacobbednarz:
When I was first putting Access support together, I made a boo-boo and
duplicated the Access Group structs into the Access Policy itself. This
shouldn't have been the case as they are the same structs, but they
should be tied to the groups and only _referenced_ in the policies
instead.

### DIFF
--- a/access_policy.go
+++ b/access_policy.go
@@ -33,58 +33,6 @@ type AccessPolicy struct {
 	Require []interface{} `json:"require"`
 }
 
-// AccessPolicyEmail is used for managing access based on the email.
-// For example, restrict access to users with the email addresses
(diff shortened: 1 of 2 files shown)


access_policy.go
```diff
@@ -33,58 +33,6 @@ type AccessPolicy struct {
 	Require []interface{} `json:"require"`
 }
 
-// AccessPolicyEmail is used for managing access based on the email.
```
jacobbednarz: See `access_groups.go` for all of these plus more.

patryk:
Thanks, there is some inconsistency with our struct naming scheme, but possibly nothing we can clean up for now.

PR #442, opened 2020-04-02
Migrate to GitHub actions

jacobbednarz:
Migrates CI to using GitHub Actions as it is faster and more reliable than Travis CI and remains free.


patryk:
Bye bye, indeed! :)

PR #446, opened 2020-04-24
Adds IP Address Management APIs for managing IP prefixes and their BGP status

cehrig:
This PR introduces interfaces for using IP Address Management APIs.
- List Prefixes
- Prefix Details
- Update Prefix Description
- Get Advertisement Status
- Update Prefix Dynamic Advertisement Status

This is the prerequisite for implementation in the Terraform Provider.
Tests have been implemented.

```
=== RUN   TestListIPPrefix
--- PASS: TestListIPPrefix (0.00s)
=== RUN   TestGetIPPrefix
--- PASS: TestGetIPPrefix (0.00s)
=== RUN   TestUpdatePrefixDescription
--- PASS: TestUpdatePrefixDescription (0.00s)
=== RUN   TestGetAdvertisementStatus
--- PASS: TestGetAdvertisementStatus (0.00s)
=== RUN   TestUpdateAdvertisementStatus
--- PASS: TestUpdateAdvertisementStatus (0.00s)
```

### DIFF
--- a/ip_address_management.go
+++ b/ip_address_management.go
@@ -1,0 +1,149 @@
+package cloudflare
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+// IPPrefix contains information about an IP prefix
+type IPPrefix struct {
+	ID                   string     `json:"id"`
+	CreatedAt            *time.Time `json:"created_at"`
+	ModifiedAt           *time.Time `json:"modified_at"`
+	CIDR                 string     `json:"cidr"`
+	AccountID            string     `json:"account_id"`
+	Description          string     `json:"description"`
+	Approved             string     `json:"approved"`
+	OnDemandEnabled      bool       `json:"on_demand_enabled"`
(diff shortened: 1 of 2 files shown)


tarnfeld:
Hey @jacobbednarz @patryk perhaps you can take a quick look when you get a moment? I wouldn't want to merge it without your input. :-)

patryk:
Great contribution, thanks @cehrig !

jacobbednarz:
The code itself is 👌 but we probably could have used more specific names for the methods and the structs as right now, the functionality is pretty ambiguous and doesn't lend itself to explaining it's intention or use. For example, `IPPrefix` is a broad struct name but used here has a very specific use case. The only suggestion I have here would be to include something along the lines of IP addressing to somewhat separate these.

- `IPPrefix` => `IPAddressingIPPrefix`
- `GetPrefix` => `IPAddressGetPrefix`

Not a big deal but something worth considering as Cloudflare have an ever growing portfolio of products and sometimes will overlap with naming (I'm looking at you Access Rules and Access the product) which can make using the library confusing if you don't dig into the method themselves.

PR #461, opened 2020-05-20
Fixes #453 updating CustomHostnames

### LINKED ISSUE #453: UpdateCustomHostnameSSL passes incorrect JSON to API endpoint
## Expected behavior

This function should update the CustomHostnames SSL settings or the entire CustomHostname

## Current behavior

The Library sends a PATCH with a CustomHostnameSSL Struct instead of a CustomHostname struct.

The API then returns a 202 but nothing actually gets updated. (So far as I can tell).
 
## Possible solution

Either update the UpdateCustomHostnameSSL to send a CustomHostnameSSL Struct wrapped so that the JSON is posted as a ssl: object  or create a new function that supports updating the whole customhostname struct.

## Steps to reproduce (for bugs)

Provide a link to a live example, or an unambiguous set of steps to
reproduce this bug. Include code to reproduce, if relevant.

1.
2.
3.
4.

## Context

Cannot update CustomHostname settings via the library

## Your Environment

the bug in:

* Version used: master 
* Environment name and version (e.g. Go 1.11, MacOS 10.14.1): go1.13.10 MacOS 10.15.3

cdloh:
## Description

This fixes #453  updating custom hostnames. The API expects a full CustomHostname NOT a CustomHostnameSSL struct as per here - https://api.cloudflare.com/#custom-hostname-for-a-zone-edit-custom-hostname

I've updated UpdateCustomHostnameSSL to send a CustomHostname struct whilst keeping the availability to just pass a CustomHostnameSSL struct.

Along with that I've added a UpdateCustomHostname function which lets you update the whole CustomHostname.

Could probably remove the UpdateCustomHostnameSSL function altogether however to keep it backward compatible I've left it in place.

I've updated the tests for both functions.

## Types of changes

## Checklist:

- [ ] I have updated the documentation accordingly. (I've updated the notes? I presume that auto updates the documentation?)

### DIFF
--- a/custom_hostname.go
+++ b/custom_hostname.go
@@ -97,7 +97,29 @@ type CustomHostnameFallbackOriginResponse struct {
 // API reference: https://api.cloudflare.com/#custom-hostname-for-a-zone-update-custom-hostname-configuration
 func (api *API) UpdateCustomHostnameSSL(zoneID string, customHostnameID string, ssl CustomHostnameSSL) (*CustomHostnameResponse, error) {
 	uri := "/zones/" + zoneID + "/custom_hostnames/" + customHostnameID
-	res, err := api.makeRequest("PATCH", uri, ssl)
+	ch := CustomHostname{
+		SSL: ssl,
+	}
+	res, err := api.makeRequest("PATCH", uri, ch)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


patryk:
Thank you!

PR #478, opened 2020-06-11
Expand `WithZoneFilter` to allow more customisation

jacobbednarz:
Updates the `WithZoneFilter` method to be renamed to `WithZoneFilters`
and allow further customisation of the query response and perform
sorting on the server side.

This will be used upstream in Terraform to replace the custom zone
filtering in favour of doing it via the API request.

I did consider making this _more_ reusable by accepting any `url.Values`
parameter but decided against it to avoid having to battle type issues.

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -138,7 +138,7 @@ func (api *API) SetAuthType(authType int) {
 // ZoneIDByName retrieves a zone's ID from the name.
 func (api *API) ZoneIDByName(zoneName string) (string, error) {
 	zoneName = normalizeZoneName(zoneName)
-	res, err := api.ListZonesContext(context.TODO(), WithZoneFilter(zoneName))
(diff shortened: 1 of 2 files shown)


cloudflare.go
```diff
@@ -420,10 +420,20 @@ type reqOption struct {
 	params url.Values
 }
 
-// WithZoneFilter applies a filter based on zone name.
-func WithZoneFilter(zone string) ReqOption {
+// WithZoneFilters applies a filter based on zone properties.
+func WithZoneFilters(zoneName, accountID, status string) ReqOption {
```
jacobbednarz, line 424: Looking through the documentation, `name`, `account.id` and `status` should provide enough flexibility here that we don't need to expand this again. If we do, it might be worth while replacing all of the `With*Filters` with the more generic method that accepts any `url.Values` instead and build the query from there instead.

patryk:
Thanks!

jacobbednarz:
🙌

Issue #479, opened 2020-06-12
CreateSSL geo_restrictions is not omitted when empty


trjstewart:
When using the China Network Enterprise feature and creating a custom SSL Certificate the Geo Key Manager feature is not supported.

## Expected behavior
According to the [API documentation](https://api.cloudflare.com/#custom-ssl-for-a-zone-create-ssl-configuration) the `geo_restrictions` property on the custom_certificates endpoint should be optional when creating a certificate via a POST request. I would expect that the below code has the output also shown below;
``` go
var opts cloudflare.ZoneCustomSSLOptions
opts.Certificate = "<certificate>"
opts.PrivateKey = "<private key>"
res, _ := json.Marshal(&opts)
log.Print(string(res))

// {
//   "certificate": "<certificate>",
//   "private_key": "<private key>",
// }
```

## Current behavior
When not providing the `geo_restrictions` property the output of the object marshal includes it as shown below;
``` go
var opts cloudflare.ZoneCustomSSLOptions
opts.Certificate = "<certificate>"
opts.PrivateKey = "<private key>"
res, _ := json.Marshal(&opts)
log.Print(string(res))

// {
//   "certificate": "<certificate>",
//   "private_key": "<private key>",
//   "geo_restrictions": { "label": ""}
// }
```
This happens because the `ZoneCustomSSLGeoRestrictions` property of the `ZoneCustomSSLOptions` struct is itself a struct and therefore never actually empty, causing the `omitempty` flag in the marshaler to never act as intended.

## Possible solution
I believe there are two solutions to this issue, however the more correct solution is a breaking change so I'm going to propose (and also submit PRs for) both.

1. change the `ZoneCustomSSLGeoRestrictions` to be a pointer within the `ZoneCustomSSLOptions` struct. This would cause a breaking change, however is the much cleaner and more forwards compatible way to implement the fix.
2. implement a custom marshaler for the `ZoneCustomSSLOptions` struct to handle omiting the `ZoneCustomSSLGeoRestrictions` property if it's not set. This would be a non-breaking change but also require the maintenance of this custom marshaler if the properties of `ZoneCustomSSLGeoRestrictions` change.

## Steps to reproduce (for bugs)
1. Setup a Cloudflare Zone with the China Network enabled.
2. Fill in the placeholders and run the below code snippet;
``` go
func main() {
  apiKey := "<api key>"
  apiEmail := "<api email>"
  api, err := cloudflare.New(apiKey, apiEmail)
[...]

PR #480, opened 2020-06-12
changes the ZoneCustomSSLGeoRestrictions to be a pointer

trjstewart:
This change fixes issue #479 with the first proposed solution; change the `ZoneCustomSSLGeoRestrictions` to be a pointer within the `ZoneCustomSSLOptions` struct. This is a breaking change as it requires the `ZoneCustomSSLGeoRestrictions` to now be passed by reference.

## Description
This is described in detail in the related Issue #479. Changing `ZoneCustomSSLGeoRestrictions` to be a pointer is the cleanest and more correct way to implement the fix, however I am mindful that this is a breaking change and may not be desirable.

I have updated the existing tests to reflect the change and I don't believe it's necessary to add any additional test cases as this is simply enforcing the already established expected behavior of `ZoneCustomSSLGeoRestrictions` being omitted when empty.

## Types of changes

## Checklist:


patryk:
Thanks for PR. I agree changing to the pointer is better solution, and we don't need to be that much vary of making breaking changes provided lib is still version 0.x, and the modifications themselves in other projects are minimal. Thanks for research!

trjstewart:
Thanks @patryk. Any chance of getting this cut into a 0.11.8 release so we can update the dependency over on https://github.com/terraform-providers/terraform-provider-cloudflare?

patryk:
Hey @trjstewart, sorry for late reply, didn't notice. We usually cut the release on a monthly cadence, or "when needed". It's not a fixed schedule.

PR #482, opened 2020-06-15
Return more revealing error message when creating an access application

Justin-Holmes:
This should hopefully help users who are getting opaque error messages.

### DIFF
--- a/access_application_test.go
+++ b/access_application_test.go
@@ -45,7 +45,7 @@ func TestAccessApplications(t *testing.T) {
 	createdAt, _ := time.Parse(time.RFC3339, "2014-01-01T05:20:00.12345Z")
 	updatedAt, _ := time.Parse(time.RFC3339, "2014-01-01T05:20:00.12345Z")
 
-	want := []AccessApplication{AccessApplication{
+	want := []AccessApplication{{
 		ID:              "480f4f69-1a28-4fdd-9240-1ed29f0ac1db",
 		CreatedAt:       &createdAt,
 		UpdatedAt:       &updatedAt,
--- a/access_application_test.go
+++ b/access_application_test.go
@@ -47,7 +47,7 @@ func TestAccessApplications(t *testing.T) {
 	createdAt, _ := time.Parse(time.RFC3339, "2014-01-01T05:20:00.12345Z")
 	updatedAt, _ := time.Parse(time.RFC3339, "2014-01-01T05:20:00.12345Z")
 
-  want := []AccessApplication{AccessApplication{
+	want := []AccessApplication{AccessApplication{
 		ID:                     "480f4f69-1a28-4fdd-9240-1ed29f0ac1db",
 		CreatedAt:              &createdAt,
 		UpdatedAt:              &updatedAt,
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -259,9 +259,9 @@ func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, u
 	switch {
 	case resp.StatusCode >= http.StatusOK && resp.StatusCode < http.StatusMultipleChoices:
 	case resp.StatusCode == http.StatusUnauthorized:
-		return nil, errors.Errorf("HTTP status %d: invalid credentials", resp.StatusCode)
+		return nil, errorFromResponse(resp.StatusCode, respBody)
 	case resp.StatusCode == http.StatusForbidden:
-		return nil, errors.Errorf("HTTP status %d: insufficient permissions", resp.StatusCode)
+		return nil, errorFromResponse(resp.StatusCode, respBody)
 	case resp.StatusCode == http.StatusServiceUnavailable,
 		resp.StatusCode == http.StatusBadGateway,
 		resp.StatusCode == http.StatusGatewayTimeout,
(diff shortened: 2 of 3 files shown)


errors.go
```diff
@@ -8,6 +8,7 @@ const (
 	errUnmarshalError       = "error unmarshalling the JSON response"
 	errRequestNotSuccessful = "error reported by API"
 	errMissingAccountID     = "account ID is empty and must be provided"
+	errAccessNotEnabled     = "access is not enabled for this account -- check that your zone is active, your domain and token are correct, and you've setup Access through the dashboard at cloudflare.com"
```
jacobbednarz, line 11: I wonder if we could reword this a little to be more specific to the entitlement missing? I don't think mentioning the zone/account/credentials check adds much here as that _should_ be picked up earlier in the HTTP back and forth.

```suggestion
	errAccessEntitlementNotEnabled     = "cloudflare access entitlement is not available for this account"
```

Thoughts?

Justin-Holmes: My main goal was to return a useful error message when a Terraform user hasn't enabled Access yet and receives a `403` when trying to create an application. After looking at this some more, I realized this PR would net us in the same spot since we don't have insight into the response status code and could return a non-helpful error message in the other direction (i.e. a bad request body).

I'm thinking it might be better to add this error message on the Terraform side and keep the error message here agnostic. Do you think that'd be a better approach?

jacobbednarz: Where possible, it makes sense to me to put these exceptions (and handling of them) into `cloudflare-go` and have as few differences as possible in the consumer clients. Example, `cloudflare-go` raises an exception based on provided logic and Terraform only needs to take care of displaying it to the end-user. This is how it currently works when recieving an exception for the `resourceCloudflareAccessApplicationCreate`

https://github.com/terraform-providers/terraform-provider-cloudflare/blob/c3e16f06f6e6daa504c276ac1e55dd81991204f8/cloudflare/resource_cloudflare_access_application.go#L61-L64

Understandably though, that may not always be possible and we may need to resort to a situation where we need both.

Looking at this again, I'm not sure if this is a concern of either the library or Terraform. If the entitlements for Access are not enabled on an account, the API should be responding as such. I know we've had issues in the past with Worker endpoints throwing perplexing issues mentioning authentication but in fact it was the entitlements not aligned. Is this the case here?

Justin-Holmes: Yep, exactly. It's an edge case that is more or less an issue with a misaligned entitlement.

I agree I think this should be handled in the API as well. Unfortunately, it looks like `403` codes have their error messages swallowed here: https://github.com/cloudflare/cloudflare-go/blob/master/cloudflare.go#L263-L264. In terms of the best way around this, I think we have a couple of options:

1. We'd return a different status code in the API to hit the default case (https://github.com/cloudflare/cloudflare-go/blob/master/cloudflare.go#L277-L283) which would render the response body. I'm not sure there is a better status code than the one we're already providing, but that could be debated (maybe a `409`?).

2. We could handle this as a special case, similar to what we're doing here: https://github.com/cloudflare/cloudflare-go/blob/master/cloudflare.go#L275-L276. Probably not ideal if we continue to have these one-offs, but sufficient in the short term.

3. We could create a custom error struct with an additional `StatusCode` field that would satisfy the built-in `error` interface. This would be more work but would allow for more flexibility in the future. 

I'm sure I'm overlooking other options, but I'm curious to get your opinion on what you think is the best way to move forward.

jacobbednarz: Ah, this is all coming back to me now. The issue why we don't pass the exact exception back was that the API responses didn't have a consistent location of where the error message was provided and we ended up in some situations where no error was returned to the end-user but things still failed. The assumptions were that a 403 would be authorisation related which, I guess if you squint, is technically the case here.

Since we've tripped up on this yet again, I think it's probably time to fix it properly to minimise the next encounter. My personal vote will lean more towards fixing the following lines:

https://github.com/cloudflare/cloudflare-go/blob/ddc9d81cec081faa805fb42362778559befaf154/cloudflare.go#L261-L264

Making them something more like this:

```go
// .. snip
case resp.StatusCode == http.StatusUnauthorized:
  return nil, errors.Errorf("HTTP status %d: %s", resp.StatusCode, respBody)
case resp.StatusCode == http.StatusForbidden:
  return nil, errors.Errorf("HTTP status %d: %s", resp.StatusCode, respBody)
```

Using this approach we don't make any assumptions about the payload, other than it exists, which should be safe enough. In an ideal world, I'd love to be able to do something like:

```go
// .. snip
case resp.StatusCode == http.StatusForbidden:
  if resp.Body.Error.Message != "" {
    return nil, errors.Errorf("HTTP status %d: %s", resp.StatusCode, respBodyError.Message)
  } else {
    return nil, errors.Errorf("HTTP status %d: %s", resp.StatusCode, respBody)
  }
```

Or similar detection where we nominate a field to hold the exceptions and if it's present, show it nicely otherwise dump the entire body value.

From your options, I also like the idea of #3 where we do a little more work now but can simplify this whole error handling section in the future to eliminate the ambiguity fully.

Justin-Holmes: @jacobbednarz I made some changes from your feedback that should give us what we need. Let me know what you think!

patryk:
Thanks!

PR #486, opened 2020-06-25
Convert /zones -> /accounts routes for access resources

Justin-Holmes:
## Description

The access team has deprecated the `/zones` routes in favor of `/accounts`. The public docs should be released over the next several days that reflect the changes in this PR. I'll be referencing a corresponding PR in the Terraform provider repository once that's complete.

## Types of changes

## Checklist:

### DIFF
--- a/access_application.go
+++ b/access_application.go
@@ -40,10 +40,10 @@ type AccessApplicationDetailResponse struct {
 	Result   AccessApplication `json:"result"`
 }
 
-// AccessApplications returns all applications within a zone.
+// AccessApplications returns all applications within an account.
 //
 // API reference: https://api.cloudflare.com/#access-applications-list-access-applications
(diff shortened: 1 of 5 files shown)


Justin-Holmes:
The deprecations page will be getting an update as well, but not exactly sure on when (definitely in the near future though). There won't be an "end of life" date since we don't have any plans of removing the `/zones` routes (AFAIK).

Justin-Holmes:
Do we have an ETA on when we think this will be released? I'm going to track this internally to make sure I don't let this fall through the cracks. 

Thanks a lot for the help!

jacobbednarz:
It's generally a fortnightly/monthly release cycle depending on the importance of what is in it. @patryk can cut a release for this and we can start using it upstream too.

patryk:
Thanks!

PR #487, opened 2020-06-25
add APIV4 interface and ability to autogenerate it

sunny-b:
## Description

We started using this package at my work and it's super handy. However, one thing that I found a bit inconvenient was that it didn't provide it's own interface that would make it easy to mock out and use in testing. This is something I've seen in other SDK's that we use and it's nice to have.

While we are creating our own interface for this package, I figured it might be a nice convenience to others using it to add an interface and the ability to automatically generate it so that the maintenance is low.

This PR adds a `generate.go` file that defines how the interface will get generated and a `Makefile` that provides the `make interface` convenience command. 

This could optionally also be used to generate a mock as well in case that something the owners of this library are interested in. It'd be as simple as adding `//go:generate mockgen -package cloudflare -source=interface.go -destination mock.go CloudflareAPI` at the end of the `generate.go` file.

I've tested it manually and it works well. The only issue was that `ifacemaker` would add an unused import to the `interface.go` file which would cause `go build` to error. That's why I added the `goimports -w .` afterwards to remove unused imports and format it.

## Types of changes

## Checklist:


sunny-b:
It looks like it's failing due to the linter not liking the name `CloudflareAPI`. However, since there's a struct that already uses the name `API`, calling the interface `API` as well wouldn't work.

go.sum
```diff
@@ -4,24 +4,10 @@ github.com/cpuguy83/go-md2man/v2 v2.0.0-20190314233015-f79a8a8ca69d/go.mod h1:ma
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c=
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
-github.com/mattn/go-runewidth v0.0.4 h1:2BvfKmzob6Bmd4YsL0zygOqfdFnK7GR4QL06Do4/p7Y=
```
sunny-b, line 7: FYI: All of this is the result of running `go mod tidy`, not from any code changes in this PR

sunny-b:
@patryk what do you think of this?

jacobbednarz:
I'm personally on the fence with this type of thing as it comes back to a point that you raised in your description (but counter to your intentions I think 😛).

> While we are creating our own interface for this package, I figured it might be a nice convenience to others using it ...

Other packages or libraries providing _me_ (as an end-user) with pre-defined mocks may or may not actually be what I want to use and in a way that I intend to use it. I can see the benefit of shipping one with the library however I'm not sure how much this matches the expectations of all consumers and whether it would be a hindrance or benefit.

cloudflare.go
```diff
@@ -48,6 +48,8 @@ type API struct {
 	logger            Logger
 }
 
+var _ APIV4 = &API{}
```
jacobbednarz, line 51: This on its own doesn't really provide much context as to the intention. Does `CloudflareAPIv4` work and meet your original goal of including "cloudflare" in the interface name? 

```suggestion
var _ CloudflareAPIv4 = &API{}
```

sunny-b: Yeah that works, though I suspect the linter may complain

sunny-b:
@jacobbednarz Yeah including a mock might not be as useful, but I still believe an interface would be a useful addition. You can still craft your own interface if you want to, but for those that find it to be a hassle, they can just use the one that comes with the library.

It's also something I've seen other SDKs do. Example: https://github.com/digitalocean/godo/blob/master/domains.go#L14

PR #488, opened 2020-06-28
Adds support for Access CA certificates

jacobbednarz:
This introduces support for managing Access CA certificates (often
referenced as short lived certificates) via the API.

I opted to call this "Access CA certificates" as that is what the functionality
technically is. Should this instead be "Access Short Lived certificates" to
line up with documentation?

Documentation: https://api.cloudflare.com/#access-short-lived-certificates-properties

Note: When I raised this PR, the create new certificate + list single
certificate details were missing the response payload in the
documentation. I instead made actual calls to confirm the correctness of
these tests and updated a Cloudflare support ticket to get this fixed
for others.

### DIFF
--- a/access_ca_certificate.go
+++ b/access_ca_certificate.go
@@ -1,0 +1,115 @@
+package cloudflare
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/pkg/errors"
+)
+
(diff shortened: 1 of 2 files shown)


patryk:
I think we can leave "CA" here if we clearly state that they mean "short-lived certificates" as a product.

PR #287, opened 2019-03-24
Add support for Cloudflare Registrar

jacobbednarz:
This introduces support for Cloudflare Registrar into `cloudflare-go`.

API documentation:

- List domains: https://api.cloudflare.com/#registrar-domains-list-domains
- Get a domain: https://api.cloudflare.com/#registrar-domains-get-domain
- Update a domain: https://api.cloudflare.com/#registrar-domains-update-domain
- Transfer domain: https://api.cloudflare.com/#registrar-domains-transfer-domain
- Cancel a transfer: https://api.cloudflare.com/#registrar-domains-cancel-transfer

PR #491, opened 2020-07-01
Fix Registrar Transfer

robbiet480:
## Description

As of today, `TransferRegistrarDomain` is totally broken. The call resides at a different URL and has different parameters and a different response. This PR begins to fix it. However, I noticed that the Cloudflare API docs for transferring are missing. I got this working by monitoring requests on dash.cloudflare.com as I transferred a domain. Therefore, I am submitting this PR as is without completing it because I need a Cloudflare employee to tell me where the API docs went and if these actions are even supported for public use. I see no deprecation notices about the transfer API so I'm hoping that the docs just went missing and Cloudflare didn't retroactively remove these APIs for public consumption without alerting customers.

Once I have these answers I will finish this PR by implementing remaining methods (like getting and updating contacts) and fixing tests.

If it turns out that Registrar Transfer support is no longer available in the API, then a Cloudflare employee should close this PR and open a new one to remove all trace of the API in another.

Thanks!

No

## Types of changes

## Checklist:

### DIFF
--- a/registrar.go
+++ b/registrar.go
@@ -53,9 +53,9 @@ type RegistrantContact struct {
 	Fax          string `json:"fax"`
 }
 
-// RegistrarDomainConfiguration is the structure for making updates to
-// and existing domain.
-type RegistrarDomainConfiguration struct {
+// UpdateRegistrarDomainConfiguration is the structure for making updates to
+// an existing domain.
+type UpdateRegistrarDomainConfiguration struct {
 	NameServers []string `json:"name_servers"`
 	Privacy     bool     `json:"privacy"`
 	Locked      bool     `json:"locked"`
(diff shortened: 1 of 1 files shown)


jacobbednarz:
> However, I noticed that the Cloudflare API docs for transferring are missing.

This is probably the biggest hurdle we need to address here. We aim to not have functionality in `cloudflare-go` that doesn't have a publicly documented endpoint. Unfortunately, watching the network traffic for the Dashboard isn't sufficient enough to deem it stable and we cannot make any guarantees of it's longevity unless it's documented. Are you able to open a support ticket and ask if this endpoint should be public and if so, ask for it to be documented?

Once we get an answer on this, we can review the PR with that context. As right now it is missing tests and failing other CI checks which need to be fixed.

robbiet480:
Hey @jacobbednarz, thanks for the quick response.

Totally understand about it can only be in the library if docs exist. The weird thing is if you look at the function comments, they all have links to docs that are now gone. Furthermore, I searched around on the forums and found a few references over the years to the endpoints, so I'm wondering if it was stable at one point (probably when you added it initially in #287) and they have since been removed. I'll go ahead and open a support ticket tonight asking for the status.

Once I get a (positive) answer, I'll go ahead and finish up fixing this PR to meet requirements, including tests.

Thanks again!

oliy:
We're planning a major overhaul of the APIs around this, so much of the functionality here is in flux.  Don't want to create confusion around the functionality in the meantime.  We will provide proper documentation around this once it is done and we have a solid API that we can maintain.

robbiet480:
@oliy Awesome, thanks for the update. I’ll close this PR for now then. Looking forward to the new APIs!

Issue #505, opened 2020-07-22
Add support for IP Lists


jacobbednarz:
@prdonahue announced support for [IP Lists today](https://blog.cloudflare.com/introducing-ip-lists/) however there isn't publicly documented API endpoints available. Once those are ready, we can add support to `cloudflare-go` and later Terraform.

prdonahue:
Thanks for creating, @jacobbednarz ! I've flagged this internally and as soon as docs are posted someone will drop a note here.

jacobbednarz:
🙌

cehrig:
I've opened a draft PR that might need some discussion around the naming of the component and how to handle asynchronous operations.

PR #507, opened 2020-07-26
Adds implementation for (IP) Rules Lists

cehrig:
This PR adds some new methods for working with IP Lists (https://api.cloudflare.com/#rules-lists-properties).
- List IP Lists
- Create IP List
- Get IP List
- Update IP List
- Delete IP List
- List IP List Items
- Create IP List Items (single and multi) (is asynchronous)
- Replace IP List Items (is asynchronous)
- Delete IP List Items (is asynchronous)
- Get IP List Item
- Get Bulk Operation Status

This feature has been introduced as IP Lists (https://blog.cloudflare.com/introducing-ip-lists/) although there will be Country lists or ASN lists in the future. The API calls it Rules Lists. I kept it more specific and re-used the term IP Lists here.

Todo:
- ~~Write Tests~~ 
- I'm on the fence if we shall keep the asynchronous behavior in the methods mentioned above and handle polling in Terraform. We could also handle polling in cloudflare-go directly. 

List IP List Items:
This endpoint works based on cursors, where each cursor returns 25 Items and there's no option to increase 'page size'. This will potentially cause some wait times when working with big IP Lists. 

#505

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -363,14 +363,21 @@ type Response struct {
 	Messages []ResponseInfo `json:"messages"`
 }
 
+// ResultInfoCursors contains information about cursors.
+type ResultInfoCursors struct {
+	Before string `json:"before"`
+	After  string `json:"after"`
+}
+
 // ResultInfo contains metadata about the Response.
 type ResultInfo struct {
-	Page       int    `json:"page"`
-	PerPage    int    `json:"per_page"`
-	TotalPages int    `json:"total_pages"`
-	Count      int    `json:"count"`
-	Total      int    `json:"total_count"`
-	Cursor     string `json:"cursor"`
+	Page       int               `json:"page"`
+	PerPage    int               `json:"per_page"`
+	TotalPages int               `json:"total_pages"`
+	Count      int               `json:"count"`
+	Total      int               `json:"total_count"`
+	Cursor     string            `json:"cursor"`
+	Cursors    ResultInfoCursors `json:"cursors"`
 }
 
 // RawResponse keeps the result as JSON form
--- a/ip_list.go
+++ b/ip_list.go
@@ -310,8 +310,6 @@ func (api *API) DeleteIPListItems(ctx context.Context, id string, items IPListIt
 	error) {
 	uri := fmt.Sprintf("/accounts/%s/rules/lists/%s/items", api.AccountID, id)
 	res, err := api.makeRequestContext(ctx, http.MethodDelete, uri, items)
-	test, _ := json.Marshal(items)
-	fmt.Printf("%+v\n", string(test))
 	if err != nil {
 		return IPListItemDeleteResponse{}, errors.Wrap(err, errMakeRequestError)
 	}
--- a/ip_list.go
+++ b/ip_list.go
@@ -11,7 +11,8 @@ import (
 )
 
 const (
-	IP_LIST_TYPE_IP = "ip"
+	// IPListTypeIP specifies a list containing IP addresses
+	IPListTypeIP = "ip"
 )
 
 // BulkOperation contains information about a Bulk Operation
--- a/ip_list.go
+++ b/ip_list.go
@@ -1,0 +1,358 @@
+package cloudflare
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+const (
+	IP_LIST_TYPE_IP = "ip"
+)
+
+// BulkOperation contains information about a Bulk Operation
+type BulkOperation struct {
+	ID        string     `json:"id"`
+	Status    string     `json:"status"`
+	Error     string     `json:"error"`
+	Completed *time.Time `json:"completed"`
+}
+
+// IPList contains information about an IP List
+type IPList struct {
+	ID                    string     `json:"id"`
+	Name                  string     `json:"name"`
+	Description           string     `json:"description"`
+	Kind                  string     `json:"kind"`
+	NumItems              int        `json:"num_items"`
+	NumReferencingFilters int        `json:"num_referencing_filters"`
+	CreatedOn             *time.Time `json:"created_on"`
+	ModifiedOn            *time.Time `json:"modified_on"`
+}
+
+// IPListItem contains information about a single IP List Item
+type IPListItem struct {
+	ID         string     `json:"id"`
+	IP         string     `json:"ip"`
+	Comment    string     `json:"comment"`
+	CreatedOn  *time.Time `json:"created_on"`
+	ModifiedOn *time.Time `json:"modified_on"`
+}
+
+// IPListCreateRequest contains data for a new IP List
+type IPListCreateRequest struct {
+	Name        string `json:"name"`
+	Description string `json:"description"`
+	Kind        string `json:"kind"`
+}
+
+// IPListItemCreateRequest contains data for a new IP List Item
+type IPListItemCreateRequest struct {
+	IP      string `json:"ip"`
+	Comment string `json:"comment"`
+}
+
+// IPListItemDeleteRequest wraps IP List Items that shall be deleted
+type IPListItemDeleteRequest struct {
+	Items []IPListItemDeleteItemRequest `json:"items"`
+}
+
+// IPListItemDeleteItemRequest contains single IP List Items that shall be deleted
+type IPListItemDeleteItemRequest struct {
+	ID string `json:"id"`
+}
+
+// IPListUpdateRequest contains data for an IP List update
+type IPListUpdateRequest struct {
+	Description string `json:"description"`
+}
+
+// IPListResponse contains a single IP Lists
+type IPListResponse struct {
+	Response
+	Result IPList `json:"result"`
+}
+
+// IPListItemResponse contains information about the creation of an IP List Item
+type IPListItemResponse struct {
+	Response
+	Result struct {
+		OperationID string `json:"operation_id"`
+	} `json:"result"`
+}
+
+// IPListListResponse contains a slice of IP Lists
+type IPListListResponse struct {
+	Response
+	Result []IPList `json:"result"`
+}
+
+// BulkOperationResponse contains a slice of IP Lists
+type BulkOperationResponse struct {
+	Response
+	Result BulkOperation `json:"result"`
+}
+
+// IPListDeleteResponse contains information about the deletion of an IP List
+type IPListDeleteResponse struct {
+	Response
+	Result struct {
+		ID string `json:"id"`
+	} `json:"result"`
+}
+
+// IPListItemsListResponse contains information about IP List Items
+type IPListItemsListResponse struct {
+	Response
+	ResultInfo `json:"result_info"`
+	Result     []IPListItem `json:"result"`
+}
+
+// IPListItemDeleteResponse contains information about the deletion of an IP List Item
+type IPListItemDeleteResponse struct {
+	Response
+	Result struct {
+		OperationID string `json:"operation_id"`
+	} `json:"result"`
+}
+
+// IPListItemsGetResponse contains information about a single IP List Item
+type IPListItemsGetResponse struct {
+	Response
+	Result IPListItem `json:"result"`
+}
+
+// ListIPLists lists all IP Lists
+// API reference: https://api.cloudflare.com/#rules-lists-list-lists
+func (api *API) ListIPLists(ctx context.Context) ([]IPList, error) {
+	uri := fmt.Sprintf("/accounts/%s/rules/lists", api.AccountID)
+	res, err := api.makeRequestContext(ctx, http.MethodGet, uri, nil)
+	if err != nil {
+		return []IPList{}, errors.Wrap(err, errMakeRequestError)
+	}
+
+	result := IPListListResponse{}
+	if err := json.Unmarshal(res, &result); err != nil {
+		return []IPList{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return result.Result, nil
+}
+
+// CreateIPList creates a new IP List
+// API reference: https://api.cloudflare.com/#rules-lists-create-list
+func (api *API) CreateIPList(ctx context.Context, name string, description string, kind string) (IPListResponse,
+	error) {
+	uri := fmt.Sprintf("/accounts/%s/rules/lists", api.AccountID)
+	res, err := api.makeRequestContext(ctx, http.MethodPost, uri,
+		IPListCreateRequest{Name: name, Description: description, Kind: kind})
+	if err != nil {
+		return IPListResponse{}, errors.Wrap(err, errMakeRequestError)
+	}
+
+	result := IPListResponse{}
+	if err := json.Unmarshal(res, &result); err != nil {
+		return IPListResponse{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return result, nil
+}
+
+// GetIPList returns a single IP List
+// API reference: https://api.cloudflare.com/#rules-lists-get-list
+func (api *API) GetIPList(ctx context.Context, id string) (IPList, error) {
+	uri := fmt.Sprintf("/accounts/%s/rules/lists/%s", api.AccountID, id)
+	res, err := api.makeRequestContext(ctx, http.MethodGet, uri, nil)
+	if err != nil {
+		return IPList{}, errors.Wrap(err, errMakeRequestError)
+	}
+
+	result := IPListResponse{}
+	if err := json.Unmarshal(res, &result); err != nil {
+		return IPList{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return result.Result, nil
+}
+
+// UpdateIPList updates the description of an existing IP List
+// API reference: https://api.cloudflare.com/#rules-lists-update-list
+func (api *API) UpdateIPList(ctx context.Context, id string, description string) (IPListResponse, error) {
+	uri := fmt.Sprintf("/accounts/%s/rules/lists/%s", api.AccountID, id)
+	res, err := api.makeRequestContext(ctx, http.MethodPut, uri, IPListUpdateRequest{Description: description})
+	if err != nil {
+		return IPListResponse{}, errors.Wrap(err, errMakeRequestError)
+	}
+
+	result := IPListResponse{}
+	if err := json.Unmarshal(res, &result); err != nil {
+		return IPListResponse{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return result, nil
+}
+
+// DeleteIPList deletes an IP List
+// API reference: https://api.cloudflare.com/#rules-lists-delete-list
+func (api *API) DeleteIPList(ctx context.Context, id string) (IPListDeleteResponse, error) {
+	uri := fmt.Sprintf("/accounts/%s/rules/lists/%s", api.AccountID, id)
+	res, err := api.makeRequestContext(ctx, http.MethodDelete, uri, nil)
+	if err != nil {
+		return IPListDeleteResponse{}, errors.Wrap(err, errMakeRequestError)
+	}
+
+	result := IPListDeleteResponse{}
+	if err := json.Unmarshal(res, &result); err != nil {
+		return IPListDeleteResponse{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return result, nil
+}
+
+// ListIPListItems returns a list with all items in an IP List
+// API reference: https://api.cloudflare.com/#rules-lists-list-list-items
+func (api *API) ListIPListItems(ctx context.Context, id string) ([]IPListItem, error) {
+	var list []IPListItem
+	var cursor string
+	var cursorQuery string
+
+	for {
+		if len(cursor) > 0 {
+			cursorQuery = fmt.Sprintf("?cursor=%s", cursor)
+		}
+		uri := fmt.Sprintf("/accounts/%s/rules/lists/%s/items%s", api.AccountID, id, cursorQuery)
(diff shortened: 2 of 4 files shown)


cehrig:
Acceptance Tests have been added
```
=== RUN   TestListIPLists
--- PASS: TestListIPLists (0.00s)
=== RUN   TestCreateIPList
--- PASS: TestCreateIPList (0.00s)
=== RUN   TestGetIPList
--- PASS: TestGetIPList (0.00s)
=== RUN   TestUpdateIPList
--- PASS: TestUpdateIPList (0.00s)
=== RUN   TestDeleteIPList
--- PASS: TestDeleteIPList (0.00s)
=== RUN   TestListIPListsItems
--- PASS: TestListIPListsItems (0.00s)
=== RUN   TestCreateIPListItems
--- PASS: TestCreateIPListItems (0.00s)
=== RUN   TestReplaceIPListItems
--- PASS: TestReplaceIPListItems (0.00s)
=== RUN   TestDeleteIPListItems
--- PASS: TestDeleteIPListItems (0.00s)
=== RUN   TestGetIPListItem
--- PASS: TestGetIPListItem (0.00s)
```

cloudflare.go
```diff
@@ -363,14 +363,21 @@ type Response struct {
...
-	Count      int    `json:"count"`
-	Total      int    `json:"total_count"`
-	Cursor     string `json:"cursor"`
+	Page       int               `json:"page"`
+	PerPage    int               `json:"per_page"`
+	TotalPages int               `json:"total_pages"`
+	Count      int               `json:"count"`
+	Total      int               `json:"total_count"`
+	Cursor     string            `json:"cursor"`
+	Cursors    ResultInfoCursors `json:"cursors"`
```
jacobbednarz, line 380: Nice! This would be great to see on other API endpoints 😄

jacobbednarz:
I see what you mean about the method names and blanketing the functionality as "IP Lists". For the future planned improvements, will they still be IPs when managing? Or will it be managed as say, the country code or ASN?

I don't like varying from the Cloudflare terminology for features as it creates confusion however I'm also wary of the potential issue we could be introducing by calling this an "IP List" if in fact, it is not a list of IPs. Does something like Firewall Lists or Firewall Rule Lists make more sense as to what this does?

cehrig:
Why actually decide between one or the other if we can have both. Create, Replace and Delete IP List Items now have 
- an asynchronous method, returning just an `operation_id` and 
- a synchronous method, that polls the status and upon completion return the current IP List. 
The algorithm for polling is rather naive and looks like this:
```
Wait for 1 second
Wait for 1 second
Wait for 2 seconds
Wait for 2 seconds
Wait for 4 seconds
Wait for 4 seconds 
...
```
Up to a total of 510 seconds.

```
var i uint8
for i = 0; i < 16; i++ {
	time.Sleep(0x1 << uint8(math.Ceil(float64(i/2))) * time.Second)
	... do stuff ...
}
```

Regarding the naming, I think continuing with IP Lists is a good approach. We will stick to the announcement from the Blog post. If there will be support for Countries or ASNs in the future, this is data that will be derived from IP Addresses in the end. @prdonahue might have other proposals but I think it's a good compromise and much clearer than something like Firewall Rules Lists.

jacobbednarz:
Having both is a great middle ground option; leave it up to the implementation to work out which suits the use case.

I'm 👍 to leaving the feature naming as is for now and we can manage the upcoming changes when they land.

cehrig:
Sounds good. I'm opening this for reviews.

ip_list.go
```diff
@@ -0,0 +1,445 @@
...
+}
+
+// IPListItemsGetResponse contains information about a single IP List Item
+type IPListItemsGetResponse struct {
+	Response
+	Result IPListItem `json:"result"`
+}
+
+// ListIPLists lists all IP Lists
+// API reference: https://api.cloudflare.com/#rules-lists-list-lists
```
jacobbednarz, line 132: In the interest of making the code comments for methods all consistent, can we add a newline between the description and API reference lines?

```suggestion
// ListIPLists lists all IP Lists
//
// API reference: https://api.cloudflare.com/#rules-lists-list-lists
```

ip_list.go
```diff
@@ -0,0 +1,445 @@
...
+	result := IPListListResponse{}
+	if err := json.Unmarshal(res, &result); err != nil {
+		return []IPList{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return result.Result, nil
+}
+
+// CreateIPList creates a new IP List
+// API reference: https://api.cloudflare.com/#rules-lists-create-list
```
jacobbednarz, line 149:
```suggestion
// CreateIPList creates a new IP List
//
// API reference: https://api.cloudflare.com/#rules-lists-create-list
```

ip_list.go
```diff
@@ -0,0 +1,445 @@
...
+	result := IPListResponse{}
+	if err := json.Unmarshal(res, &result); err != nil {
+		return IPList{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return result.Result, nil
+}
+
+// GetIPList returns a single IP List
+// API reference: https://api.cloudflare.com/#rules-lists-get-list
```
jacobbednarz, line 168:
```suggestion
// GetIPList returns a single IP List
//
// API reference: https://api.cloudflare.com/#rules-lists-get-list
```

ip_list.go
```diff
@@ -0,0 +1,445 @@
...
+	result := IPListResponse{}
+	if err := json.Unmarshal(res, &result); err != nil {
+		return IPList{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return result.Result, nil
+}
+
+// UpdateIPList updates the description of an existing IP List
+// API reference: https://api.cloudflare.com/#rules-lists-update-list
```
jacobbednarz, line 185:
```suggestion
// UpdateIPList updates the description of an existing IP List
//
// API reference: https://api.cloudflare.com/#rules-lists-update-list
```

ip_list.go
```diff
@@ -0,0 +1,445 @@
...
+	result := IPListResponse{}
+	if err := json.Unmarshal(res, &result); err != nil {
+		return IPList{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return result.Result, nil
+}
+
+// DeleteIPList deletes an IP List
+// API reference: https://api.cloudflare.com/#rules-lists-delete-list
```
jacobbednarz, line 202:
```suggestion
// DeleteIPList deletes an IP List
//
// API reference: https://api.cloudflare.com/#rules-lists-delete-list
```

ip_list.go
```diff
@@ -0,0 +1,445 @@
...
+	result := IPListDeleteResponse{}
+	if err := json.Unmarshal(res, &result); err != nil {
+		return IPListDeleteResponse{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return result, nil
+}
+
+// ListIPListItems returns a list with all items in an IP List
+// API reference: https://api.cloudflare.com/#rules-lists-list-list-items
```
jacobbednarz, line 219:
```suggestion
// ListIPListItems returns a list with all items in an IP List
//
// API reference: https://api.cloudflare.com/#rules-lists-list-list-items
```

ip_list.go
```diff
@@ -0,0 +1,445 @@
...
+			break
+		}
+	}
+
+	return list, nil
+}
+
+// CreateIPListItemAsync creates a new IP List Item asynchronously, Users have to poll the operation status by
+// using the operation_id returned by this function.
+// API reference: https://api.cloudflare.com/#rules-lists-create-list-items
```
jacobbednarz, line 251:
```suggestion
// CreateIPListItemAsync creates a new IP List Item asynchronously. Users have to poll the operation status by
// using the operation_id returned by this function.
//
// API reference: https://api.cloudflare.com/#rules-lists-create-list-items
```

ip_list.go
```diff
@@ -0,0 +1,445 @@
...
+	if err != nil {
+		return []IPListItem{}, err
+	}
+
+	return api.ListIPListItems(ctx, id)
+}
+
+// CreateIPListItemsAsync bulk creates many IP List Items asynchronously. Users have to poll the operation status by
+// using the operation_id returned by this function.
+// API reference: https://api.cloudflare.com/#rules-lists-create-list-items
```
jacobbednarz, line 286:
```suggestion
// using the operation_id returned by this function.
//
// API reference: https://api.cloudflare.com/#rules-lists-create-list-items
```

ip_list.go
```diff
@@ -0,0 +1,445 @@
...
+	if err != nil {
+		return []IPListItem{}, err
+	}
+
+	return api.ListIPListItems(ctx, id)
+}
+
+// ReplaceIPListItemsAsync replaces all IP List Items asynchronously. Users have to poll the operation status by
+// using the operation_id returned by this function.
+// API reference: https://api.cloudflare.com/#rules-lists-replace-list-items
```
jacobbednarz, line 321:
```suggestion
// using the operation_id returned by this function.
//
// API reference: https://api.cloudflare.com/#rules-lists-replace-list-items
```

ip_list.go
```diff
@@ -0,0 +1,445 @@
...
+	if err != nil {
+		return []IPListItem{}, err
+	}
+
+	return api.ListIPListItems(ctx, id)
+}
+
+// DeleteIPListItemsAsync removes specific Items of an IP List by their ID asynchronously. Users have to poll the
+// operation status by using the operation_id returned by this function.
+// API reference: https://api.cloudflare.com/#rules-lists-delete-list-items
```
jacobbednarz, line 356:
```suggestion
// operation status by using the operation_id returned by this function.
//
// API reference: https://api.cloudflare.com/#rules-lists-delete-list-items
```

ip_list.go
```diff
@@ -0,0 +1,445 @@
...
+	err = api.pollIPListBulkOperation(ctx, result.Result.OperationID)
+	if err != nil {
+		return []IPListItem{}, err
+	}
+
+	return api.ListIPListItems(ctx, id)
+}
+
+// GetIPListItem returns a single IP List Item
+// API reference: https://api.cloudflare.com/#rules-lists-get-list-item
```
jacobbednarz, line 391:
```suggestion
// GetIPListItem returns a single IP List Item
//
// API reference: https://api.cloudflare.com/#rules-lists-get-list-item
```

ip_list.go
```diff
@@ -0,0 +1,445 @@
...
+	result := IPListItemsGetResponse{}
+	if err := json.Unmarshal(res, &result); err != nil {
+		return IPListItem{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return result.Result, nil
+}
+
+// GetIPListBulkOperation returns the status of a bulk operation
+// API reference: https://api.cloudflare.com/#rules-lists-get-bulk-operation
```
jacobbednarz, line 408:
```suggestion
// GetIPListBulkOperation returns the status of a bulk operation
//
// API reference: https://api.cloudflare.com/#rules-lists-get-bulk-operation
```

ip_list.go
```diff
@@ -0,0 +1,445 @@
...
+	result := IPListBulkOperationResponse{}
+	if err := json.Unmarshal(res, &result); err != nil {
+		return IPListBulkOperation{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return result.Result, nil
+}
+
+// this is a helper method to implement synchronous behavior for some asynchronous endpoints.
+// bulk-operation status can be either pending, running, failed or completed
```
jacobbednarz, line 425: In the interest of matching GoDoc conventions, the comments of a method should be preceded by the element which it describes. I'm not sure why `go vet` didn't already pick this up but I'll look into it.

```suggestion
// pollIPListBulkOperation implements synchronous behaviour for some asynchronous endpoints.
// bulk-operation status can be either pending, running, failed or completed
```

ip_list.go
```diff
@@ -0,0 +1,445 @@
...
+	var i uint8
+	for i = 0; i < 16; i++ {
+		time.Sleep(0x1 << uint8(math.Ceil(float64(i/2))) * time.Second)
+
+		bulkResult, err := api.GetIPListBulkOperation(ctx, id)
+		if err != nil {
+			return err
+		}
+
+		switch bulkResult.Status {
```
jacobbednarz, line 436: I'm not familiar enough with the internals of this functionality so feel free to educate but do we want a `default` case on this switch to account for any unknown exceptions or statuses that are returned? I'm not too tied to it, but it is a convention that has saved me a number of times in the past for unexpected scenarios.

ip_list_test.go
```diff
@@ -0,0 +1,464 @@
...
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, r.Method, "GET", "Expected method 'GET', got %s", r.Method)
+		w.Header().Set("content-type", "application/json")
+
+		if len(r.URL.Query().Get("cursor")) > 0 && r.URL.Query().Get("cursor") == "yyy" {
+			fmt.Fprint(w, `{
+			"result": [
+    			{
+      				"id": "2c0fc9fa937b11eaa1b71c4d701ab86e",
+      				"ip": "10.0.0.2",
```
jacobbednarz, line 246: Are we able to swap this IP out for [a reserved IP designated for documentation/examples](https://tools.ietf.org/html/rfc5737)?

```suggestion
      				"ip": "192.0.2.2",
```

ip_list_test.go
```diff
@@ -0,0 +1,464 @@
...
+			"success": true,
+			"errors": [],
+			"messages": []
+		}`)
+		} else {
+			fmt.Fprint(w, `{
+			"result": [
+    			{
+      				"id": "2c0fc9fa937b11eaa1b71c4d701ab86e",
+      				"ip": "10.0.0.1",
```
jacobbednarz, line 266:
```suggestion
      				"ip": "192.0.2.1",
```

ip_list_test.go
```diff
@@ -0,0 +1,464 @@
...
+
+	mux.HandleFunc("/accounts/foo/rules/lists/2c0fc9fa937b11eaa1b71c4d701ab86e/items", handler)
+
+	createdOn, _ := time.Parse(time.RFC3339, "2020-01-01T08:00:00Z")
+	modifiedOn, _ := time.Parse(time.RFC3339, "2020-01-10T14:00:00Z")
+
+	want := []IPListItem{
+		{
+			ID:         "2c0fc9fa937b11eaa1b71c4d701ab86e",
+			IP:         "10.0.0.1",
```
jacobbednarz, line 294:
```suggestion
			IP:         "192.0.2.1",
```

ip_list_test.go
```diff
@@ -0,0 +1,464 @@
...
+			Comment:    "Private IP address",
+			CreatedOn:  &createdOn,
+			ModifiedOn: &modifiedOn,
+		},
+		{
+			ID:         "2c0fc9fa937b11eaa1b71c4d701ab86e",
+			IP:         "10.0.0.2",
```
jacobbednarz, line 301:
```suggestion
			IP:         "192.0.2.2",
```

ip_list_test.go
```diff
@@ -0,0 +1,464 @@
...
+
+	want := IPListItemCreateResponse{}
+	want.Success = true
+	want.Errors = []ResponseInfo{}
+	want.Messages = []ResponseInfo{}
+	want.Result.OperationID = "4da8780eeb215e6cb7f48dd981c4ea02"
+
+	actual, err := client.CreateIPListItemsAsync(context.Background(), "2c0fc9fa937b11eaa1b71c4d701ab86e",
+		[]IPListItemCreateRequest{{
+			IP:      "10.0.0.1",
```
jacobbednarz, line 341:
```suggestion
			IP:      "192.0.2.1",
```

ip_list_test.go
```diff
@@ -0,0 +1,464 @@
...
+			Comment: "Private IP",
+		}, {
+			IP:      "10.0.0.2",
```
jacobbednarz, line 344:
```suggestion
			IP:      "192.0.2.2",
```

ip_list_test.go
```diff
@@ -0,0 +1,464 @@
...
+
+	want := IPListItemCreateResponse{}
+	want.Success = true
+	want.Errors = []ResponseInfo{}
+	want.Messages = []ResponseInfo{}
+	want.Result.OperationID = "4da8780eeb215e6cb7f48dd981c4ea02"
+
+	actual, err := client.ReplaceIPListItemsAsync(context.Background(), "2c0fc9fa937b11eaa1b71c4d701ab86e",
+		[]IPListItemCreateRequest{{
+			IP:      "10.0.0.1",
```
jacobbednarz, line 379:
```suggestion
			IP:      "192.0.2.1",
```

ip_list_test.go
```diff
@@ -0,0 +1,464 @@
...
+			Comment: "Private IP",
+		}, {
+			IP:      "10.0.0.2",
```
jacobbednarz, line 382:
```suggestion
			IP:      "192.0.2.2",
```

ip_list_test.go
```diff
@@ -0,0 +1,464 @@
...
+	setup(UsingAccount("foo"))
+	defer teardown()
+
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, r.Method, "GET", "Expected method 'GET', got %s", r.Method)
+		w.Header().Set("content-type", "application/json")
+		fmt.Fprint(w, `{
+			"result": {
+				"id": "2c0fc9fa937b11eaa1b71c4d701ab86e",
+				"ip": "10.0.0.1",
```
jacobbednarz, line 434:
```suggestion
				"ip": "192.0.2.1",
```

ip_list_test.go
```diff
@@ -0,0 +1,464 @@
...
+
+	mux.HandleFunc("/accounts/foo/rules/lists/2c0fc9fa937b11eaa1b71c4d701ab86e/items/"+
+		"34b12448945f11eaa1b71c4d701ab86e", handler)
+
+	createdOn, _ := time.Parse(time.RFC3339, "2020-01-01T08:00:00Z")
+	modifiedOn, _ := time.Parse(time.RFC3339, "2020-01-10T14:00:00Z")
+
+	want := IPListItem{
+		ID:         "2c0fc9fa937b11eaa1b71c4d701ab86e",
+		IP:         "10.0.0.1",
```
jacobbednarz, line 453:
```suggestion
		IP:         "192.0.2.1",
```

ip_list.go
```diff
@@ -0,0 +1,445 @@
...
+	}
+
+	return list, nil
+}
+
+// CreateIPListItemAsync creates a new IP List Item asynchronously, Users have to poll the operation status by
+// using the operation_id returned by this function.
+// API reference: https://api.cloudflare.com/#rules-lists-create-list-items
+func (api *API) CreateIPListItemAsync(ctx context.Context, id string, ip string, comment string) (
+	IPListItemCreateResponse, error) {
```
jacobbednarz, line 253: As the last three arguments in this method are of the same type, do we want to consolidate it into a single definition?

```suggestion
func (api *API) CreateIPListItemAsync(ctx context.Context, id, ip, comment string) (IPListItemCreateResponse, error) {
```

I'm also wary of breaking this method signature onto a new line at the return types as I nearly missed it during this review and struggled to grok what we were expecting to return. How do you feel about leaving this on a single line for readability despite it violating your desired line length?

ip_list.go
```diff
@@ -0,0 +1,445 @@
...
+	result := IPListItemCreateResponse{}
+	if err := json.Unmarshal(res, &result); err != nil {
+		return IPListItemCreateResponse{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return result, nil
+}
+
+// CreateIPListItem creates a new IP List Item synchronously and returns the current set of IP List Items
+func (api *API) CreateIPListItem(ctx context.Context, id string, ip string, comment string) ([]IPListItem, error) {
```
jacobbednarz, line 269: Similar to the suggestion above, can we omit the extra types in this method signature since they are duplicated?

```suggestion
func (api *API) CreateIPListItem(ctx context.Context, id, ip, comment string) ([]IPListItem, error) {
```

ip_list.go
```diff
@@ -0,0 +1,445 @@
...
+	return api.ListIPListItems(ctx, id)
+}
+
+// CreateIPListItemsAsync bulk creates many IP List Items asynchronously. Users have to poll the operation status by
+// using the operation_id returned by this function.
+// API reference: https://api.cloudflare.com/#rules-lists-create-list-items
+func (api *API) CreateIPListItemsAsync(ctx context.Context, id string, items []IPListItemCreateRequest) (
+	IPListItemCreateResponse, error) {
+	uri := fmt.Sprintf("/accounts/%s/rules/lists/%s/items", api.AccountID, id)
+	res, err := api.makeRequestContext(ctx, http.MethodPost, uri, items)
```
jacobbednarz, line 290: I quite like `http.MethodPost` over `"POST"`, thanks for this! 🍭

ip_list.go
```diff
@@ -0,0 +1,445 @@
...
+	if err != nil {
+		return []IPListItem{}, err
+	}
+
+	return api.ListIPListItems(ctx, id)
+}
+
+// GetIPListItem returns a single IP List Item
+// API reference: https://api.cloudflare.com/#rules-lists-get-list-item
+func (api *API) GetIPListItem(ctx context.Context, listID string, id string) (IPListItem, error) {
```
jacobbednarz, line 392:
```suggestion
func (api *API) GetIPListItem(ctx context.Context, listID, id string) (IPListItem, error) {
```

ip_list.go
cehrig: The bulk operation should return pending, running, completed or failed. But it's a very good point handling a default case here if something unexpected happens. I've added this.

cehrig:
I've applied your suggestions. Thanks a bunch @jacobbednarz  for taking the time reviewing!

PR #520, opened 2020-08-24
diagnostics: Add support for traceroute

jacobbednarz:
Adds support for initiating a traceroute using the Cloudflare network to
the specified targets.

API documentation: https://api.cloudflare.com/#diagnostics-traceroute


patryk (approved):
Thanks!

PR #521, opened 2020-08-24
diagnostics: `s/latency/rtt`

jacobbednarz:
The documentation for this endpoint mentions "latency" in place of "rtt"
however the actual responses use the latter.

I've raised 1961060 to get the public documentation updated to reflect
the actual responses.


patryk (approved):

patryk:
Thank you!

PR #526, opened 2020-09-09
fix: Set context for requests properly

sebnow:
## Description

`WithContext` does not mutate the request, and the returned `Request` value is ignored, so context for requests is never actually being set.

A regression test was added that ensure the context is properly passed. It failed prior to changes to the implementation.

## Types of changes

## Checklist:


patryk:
Thanks!

PR #533, opened 2020-09-25
Add the ability to call for a single page and get result info metadata for DNS Records

cmalec:
Add a function that returns in a paged fashion and also gives back the result info metadata. 

## Description
The SDK has no way to currently get the metadata for a zone like record count. If you want just the metadata you should not need to collect all records. Also an account with a large number of records would benefit from being able to page the results themselves. 

I have the changes done in my own application and have tested it. I do not see any DNS record tests for the other functions.

## Types of changes

## Checklist:

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -2,6 +2,7 @@ package cloudflare
 
 import (
 	"encoding/json"
+	"fmt"
 	"net/url"
 	"strconv"
 	"time"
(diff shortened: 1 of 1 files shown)


jacobbednarz:
Some of the things you've mentioned have been intentionally designed that way. For example, DNS records, we automatically handle pagination and request all pages while returning an array of the records. The metadata is of no use as we handle it internally without needing to expose the user to it. Could you elaborate on some of your use cases this code is intended to help with so we can better understand it?

cmalec:
@jacobbednarz well my personal example is I have a need for tracking the record count of my zones. we have a lot of zones with a lot of records so it doesn't make sense to get them all as its a waste of resources and time(would probably timeout). However if this is by design thats fair as my case is probably not a common use case, I can continue to just hit the api directly and not go through the sdk. Feel free to close this PR if thats the case.

dns.go
```diff
@@ -109,6 +87,41 @@ func (api *API) DNSRecords(zoneID string, rr DNSRecord) ([]DNSRecord, error) {
 	return records, nil
 }
 
+// DNSRecordsWithResultInfo returns the full API response a slice of DNS records and Result info metadata for the given zone identifier.
+//
+// This takes a DNSRecord to allow filtering of the results returned.
+//
+// API reference: https://api.cloudflare.com/#dns-records-for-a-zone-list-dns-records
+func (api *API) DNSRecordsWithResultInfo(zoneID string, rr DNSRecord, perPage int, page int) (DNSListResponse, error) {
+	// Construct a query string
```
jacobbednarz, line 96:
```suggestion
```

dns.go
```diff
@@ -109,6 +87,41 @@ func (api *API) DNSRecords(zoneID string, rr DNSRecord) ([]DNSRecord, error) {
...
+		v.Set("type", rr.Type)
+	}
+	if rr.Content != "" {
+		v.Set("content", rr.Content)
+	}
+
+	var query string
+	v.Set("page", strconv.Itoa(page))
+	query = "?" + fmt.Sprintf("per_page=%d", perPage) + v.Encode()
+	uri := "/zones/" + zoneID + "/dns_records" + query
```
jacobbednarz, line 111: You don't need to initialise a new variable here; we can continue using `v.Set` and have `v.Encode()` handle it on the `uri` when it's called.

```suggestion
	v.Set("page", strconv.Itoa(page))
	v.Set("per_page", strconv.Itoa(perPage))
	uri := "/zones/" + zoneID + "/dns_records?" + v.Encode()
```

jacobbednarz:
Gotcha, thanks for expanding on your use case. After thinking about this more, I think we don't really need a method to do this when you could instead perform a `len()` on `DNSRecords()` and achieve the same outcome. 

```go
package main

import (
	"fmt"
	"log"
	"os"

	cloudflare "github.com/cloudflare/cloudflare-go"
)

func main() {
	api, _ := cloudflare.New(os.Getenv("CLOUDFLARE_API_KEY"), os.Getenv("CLOUDFLARE_EMAIL"))
	rr := cloudflare.DNSRecord{}
	records, _ := api.DNSRecords(os.Getenv("CLOUDFLARE_ZONE_ID"), rr)

	fmt.Println(len(records))
}
```

I just checked this on 3 zones that have ~ 2k records each and it took under 10s. Could you try the above and let me know if you experience any issues? We currently don't use any of Go's concurrency features here but if you're experiencing performance issues with large data sets, I'm happy to take a look at adding that to speed things up.

jacobbednarz:
Closing due to staleness. Feel free to let me know if you have feedback and we can continue discussions.

PR #534, opened 2020-09-27
zone: allow `ListZonesContext` to automatically handle pagination

jacobbednarz:
Updates the `ListZonesContext` method to automatically handle the
pagination provided by the API response to return all zones regardless
of the number of entries.

This will fix cloudflare/terraform-provider-cloudflare#790 which uses
this functionality for filtering the data resource.

Fixes cloudflare/terraform-provider-cloudflare#790

### DIFF
--- a/zone.go
+++ b/zone.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
+	"strconv"
 	"sync"
 	"time"
 
@@ -354,6 +355,7 @@ func (api *API) ZoneActivationCheck(zoneID string) (Response, error) {
 // API reference: https://api.cloudflare.com/#zone-list-zones
 func (api *API) ListZones(z ...string) ([]Zone, error) {
 	v := url.Values{}
+
 	var res []byte
 	var r ZonesResponse
 	var zones []Zone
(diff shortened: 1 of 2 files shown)


jacobbednarz:
Confirmed the functionality is working however the tests aren't happy with it. I'll dig into the tests and find out why.

jacobbednarz:
I've updated the stubbed response in 259a650 which was the cause of the failing test 🤦‍♂️

zone.go
```diff
@@ -444,6 +451,43 @@ func (api *API) ListZonesContext(ctx context.Context, opts ...ReqOption) (r Zone
 		return ZonesResponse{}, errors.Wrap(err, errUnmarshalError)
 	}
 
+	totalPageCount := r.TotalPages
+	var wg sync.WaitGroup
+	wg.Add(totalPageCount)
+	errc := make(chan error)
```
patryk, line 457: It doesn't seem we handle errors returned from channel in any way. We should return them at the end of the function, even a single one is enough. We might also want to cancel pending requests if any goroutine fails this way.

patryk (changes requested):

zone.go
jacobbednarz: Is [this](https://github.com/cloudflare/cloudflare-go/pull/534/files#diff-6d1d02aea42947cbb0ae78b9690828a2R477-R478) what you mean? Or did you have something else in mind for the error being returned? I'm not explicitly cancelling the routine but can drop it in if the channel error reporting isn't enough here.

jacobbednarz: Circling back on this one @patryk as we've still got an issue open in the Terraform provider depending on this one.

jacobbednarz:
As this is holding up a Terraform fix (and has sat for a while), let's merge this one in and when we want to fix the error handling you're mentioning we can raise a separate PR to address.

PR #540, opened 2020-10-11
fix: properly report messages from an "error" field instead of "errors".

fvbommel:
## Description

I was trying to dynamically update a DNS record pointing to my home IP address, but got an error:

> `Error creating DNS record: error from makeRequest: HTTP status 401: `

(and a similar message with s/creating/updating/ when trying to update an existing record)

Clearly, something was missing at the end there.

After a bit of debugging it turns out that the API does not always report errors in an `errors` field, in this case it reported one in an `error` field instead.

This PR adds an `Error string` field to the `Response` type, and if it's not empty the message in it is reported just like the messages in `Errors`.

I re-ran my command and got

> `Error creating DNS record: error from makeRequest: HTTP status 401: You cannot use this API for domains with a .cf, .ga, .gq, .ml, or .tk TLD (top-level domain). To configure the DNS settings for this domain, use the Cloudflare Dashboard.`

Which as an aside could have been better documented (maybe a warning when I was creating the API token specifically scoped to editing this zone?), but I guess I'll figure something else out.

I also added a unit test.

## Types of changes

## Checklist:

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -359,6 +359,7 @@ type ResponseInfo struct {
 type Response struct {
 	Success  bool           `json:"success"`
 	Errors   []ResponseInfo `json:"errors"`
+	Error    string         `json:"error"`
 	Messages []ResponseInfo `json:"messages"`
 }
 
@@ -465,5 +466,10 @@ func errorFromResponse(statusCode int, respBody []byte) error {
 	for _, v := range r.Errors {
 		errMsgs = append(errMsgs, v.Message)
 	}
+
+	if r.Error != "" {
+		errMsgs = append(errMsgs, r.Error)
+	}
+
 	return errors.Errorf("HTTP status %d: %s", statusCode, strings.Join(errMsgs, " "))
 }
--- a/cloudflare_test.go
+++ b/cloudflare_test.go
@@ -685,6 +685,35 @@ func TestErrorFromResponse(t *testing.T) {
 	assert.EqualError(t, err, "error from makeRequest: HTTP status 403: this is a test error")
 }
 
+func TestSingleErrorFromResponse(t *testing.T) {
+	setup()
+	defer teardown()
+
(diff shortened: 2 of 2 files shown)


jacobbednarz:
Cheers for pushing this @fvbommel. This was also raised in the Terraform provider (cloudflare/terraform-provider-cloudflare#705) and I've raised 1983882 with Cloudflare to get the response updated as the `error` field being produced by this endpoint doesn't match the [expected HTTP response outlined by the API documentation](https://api.cloudflare.com/#getting-started-responses).

The ideal here is that we don't merge this PR but instead get the API response fixed to conform as there are other parts of the code that rely on this specific structure being used.

fvbommel:
> The ideal here is that we don't merge this PR but instead get the API response fixed to conform as there are other parts of the code that rely on this specific structure being used.

That's fine too, I was just trying to prevent other people from running into this cryptic message, and solving it at the source would indeed be a better way to do that because it would apply to _all_ API users, not just the ones using this particular library.

I disagree about the ideal though, I think the *ideal* solution would be to get rid of this error entirely and just make it work for those TLDs as well. I understand that there's probably more abuse on them, given that they're free and thus "disposable" to bad actors. Still, ideally this API would work for the rest of their users...

jacobbednarz:
The latter part is something that is entirely out of our hands at this point, I'm afraid. They aren't able to be interacted with via the API due to the high levels of abuse and I can't see that changing any time soon. Instead, if people are using these domains, they must use the UI dashboard and not the APIs directly themselves. Hopefully we can get this response updated accordingly in the coming week or so.

zech564 (approved):

fvbommel:
I rebased on top of master to fix conflicts.

jacobbednarz:
This is no longer required as the API response is being fixed to match the standard response structure.

jacobbednarz:
API response has been updated.

PR #547, opened 2020-10-27
Make `DiagnosticsTracerouteConfiguration.Colos` `omitempty`

### LINKED ISSUE #546: `PerformTraceroute` should work without an explicit colocation list
`PerformTraceroute` should work without an explicit colocation list passed to it.

## Expected behavior

When not passing/passing empty `colos` to `PerformTraceroute`, the 'colos' field should be omitted from the API request, causing all colos should be queried. [docs](https://api.cloudflare.com/#diagnostics-traceroute)

## Current behavior

When not passing/passing empty `colos` to `PerformTraceroute` an empty array is sent to the API, causing an error

```
failed to perform traceroute: error from makeRequest: HTTP status 400: content "{\n  \"result\": null,\n  \"success\": false,\n  \"errors\": [\n    {\n      \"code\": 1005,\n      \"message\": \"Colo is not a valid colo\"\n    }\n  ],\n  \"messages\": []\n}\n"
```

## Possible solution

https://github.com/cloudflare/cloudflare-go/blob/master/diagnostics.go#L14

`Colos` should include `,omitempty`

## Steps to reproduce (for bugs)

1. Have an enterprise plan
2. Execute `PerformTraceroute` with colos being an empty slice
   - for example with `go run -a github.com/jacobbednarz/cf-traceroute -api-key=${CF_TOKEN} -email=${CF_EMAIL} -targets=1.1.1.1 -account-id=${CF_ACC_ID}`

## Context

`github.com/jacobbednarz/cf-traceroute` requires an explicit colo list because of this behavior. If this is fixed, that parameter can be omitted.  
Their code does not enforce the colo list but will error out as described above. With the adapted payload the API request should work just fine. (Verified via manual curl to the API)

## References

-

## Your Environment

the bug in:

* Version used (e.g. git sha or specific tag): ` master` 
* Environment name and version (e.g. Go 1.11, MacOS 10.14.1): ` go version go1.15.2 linux/amd64`

T4cC0re:
## Description

This will ensure that `PerformTraceroute` works without an explicit
colocation list passed to it as documented in the API docs.

Fixes #546

The previously failing condition was introduced as a test case

## Types of changes

## Checklist:

### DIFF
--- a/diagnostics.go
+++ b/diagnostics.go
@@ -11,7 +11,7 @@ import (
 // diagnostics traceroute requests.
 type DiagnosticsTracerouteConfiguration struct {
 	Targets []string                                  `json:"targets"`
-	Colos   []string                                  `json:"colos"`
+	Colos   []string                                  `json:"colos,omitempty"`
 	Options DiagnosticsTracerouteConfigurationOptions `json:"options,omitempty"`
 }
 
--- a/diagnostics_test.go
+++ b/diagnostics_test.go
@@ -1,6 +1,7 @@
 package cloudflare
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"testing"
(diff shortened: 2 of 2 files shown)


patryk (approved):
Thanks, @T4cC0re!

jacobbednarz:
Thank you @T4cC0re! This is great and I'll be sure to pull in this fix into `cf-traceroute` when it's available.

PR #553, opened 2020-11-04
feat(APIToken): make time.Time properties optional

UrosSimovic:
Now it is not possible to create non-expiring token or one would have to use a time far in the future.

Let's leave that decision to a user by making properties `NotBefore` and `ExpiresOn` pointers and thus nilable.

`ModifiedOn` and `IssuedOn` should also be nilable as they are set by API. To not sending json similar to the one below when creating new APIToken:
```
{
  ...
  "modified_on": "0001-01-01T00:00:00Z",
  "issued_on": "0001-01-01T00:00:00Z",
  ...
}
```

### DIFF
--- a/api_token.go
+++ b/api_token.go
@@ -13,8 +13,8 @@ type APIToken struct {
 	ID         string             `json:"id,omitempty"`
 	Name       string             `json:"name"`
 	Status     string             `json:"status,omitempty"`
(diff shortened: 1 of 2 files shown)


jacobbednarz (approved):

patryk (approved):
Thanks!

PR #572, opened 2021-01-06
argo_tunnel: make V4 API response compatible

jacobbednarz:
While diagnosing some issues implementing the `cloudflare_argo_tunnel`
resource in Terraform, I noticed that the API response was lacking the
outer response structure that the other V4 endpoints have.

The early implementation of Argo Tunnel endpoints didn't conform to the
V4 API standard response structure. This has been remedied going forward
however to support older clients this isn't yet the default. An explicit
`Accept` header is used to get the V4 compatible version.

Along with this change, is removing the use of the `context` package in
favour of existing request helpers. I opted not to reinvent a specific
case for this right now because:

- It's highly unlikely these requests will need to be cancelled; and
- This is currently a snowflake scenario and I didn't want to build a
  new interface after only needing it once. I'd prefer to wait for more
  use cases before expanding it.

As this change is internal, it won't impact clients and we can revert
this once the default response is the V4 compatible version without
interruption.

### DIFF
--- a/argo_tunnel.go
+++ b/argo_tunnel.go
@@ -47,7 +47,7 @@ type ArgoTunnelDetailResponse struct {
 func (api *API) ArgoTunnels(ctx context.Context, accountID string) ([]ArgoTunnel, error) {
 	uri := "/accounts/" + accountID + "/tunnels"
 
-	res, err := api.makeRequestContextWithHeaders(context.Background(), "GET", uri, nil, argoV1Header())
+	res, err := api.makeRequestContextWithHeaders(ctx, "GET", uri, nil, argoV1Header())
 	if err != nil {
 		return []ArgoTunnel{}, errors.Wrap(err, errMakeRequestError)
 	}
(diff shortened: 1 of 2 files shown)


patryk:
I'd prefer we kept context in these functions (the idea is to refactor all remaining functions in #324 by 0.14 release). Because we don't have `makeRequest...` with headers and context, we should add one.

jacobbednarz:
Sure thing. I didn't really want to copy paste as is due to the confusion around which method you should be using for what purposes. Alongside that, the method signatures are quite long which makes determining parameters a little more confusing again. 

I'll add a new method that follows the same patterns and we can always rejig it once we get a bit more usage under our belt.

jacobbednarz:
f90d9754aa1f14cd0bc24245f2789c41da784a35 has the changes I've mentioned above.

patryk:
No worries, they are internal to the library (lowercased name), we will likely clean them up during refactor by allowing optional arguments.

PR #584, opened 2021-02-04
Handle all 4xx responses.

jsha:
Previously, the API code only handled specific 4xx errors, like 403.
However, the HTTP spec says all 4xx series status should be treated as
client errors.

In particular, certain endpoints (like the firewall endpoints) return
405 in response to API Token auth. This causes a nil pointer
dereference.

https://tools.ietf.org/html/rfc7231#section-6

> HTTP clients are not required to
> understand the meaning of all registered status codes, though such
> understanding is obviously desirable.  However, a client MUST
> understand the class of any status code, as indicated by the first
> digit, and treat an unrecognized status code as being equivalent to
> the x00 status code of that class, with the exception that a
> recipient MUST NOT cache a response with an unrecognized status code.

Provide a general summary of your changes in the title above. You should
remove this overview, any sections and any section descriptions you
don't need below before submitting. There isn't a strict requirement to
use this template if you can structure your description and still cover
these points.

## Description

Describe your changes in detail through motivation and context. Why is
this change required? What problem does it solve? If it fixes an open
issue, link to the issue using GitHub's closing issues keywords[1].

required.

## Types of changes

## Checklist:

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -262,9 +262,7 @@ func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, u
 
 	switch {
 	case resp.StatusCode >= http.StatusOK && resp.StatusCode < http.StatusMultipleChoices:
-	case resp.StatusCode == http.StatusUnauthorized:
-		return nil, errorFromResponse(resp.StatusCode, respBody)
-	case resp.StatusCode == http.StatusForbidden:
+	case resp.StatusCode >= 400 && resp.StatusCode < 500:
 		return nil, errorFromResponse(resp.StatusCode, respBody)
 	case resp.StatusCode == http.StatusServiceUnavailable,
 		resp.StatusCode == http.StatusBadGateway,
--- a/go.sum
+++ b/go.sum
@@ -23,6 +23,7 @@ github.com/urfave/cli/v2 v2.3.0 h1:qph92Y649prgesehzOrQjdWyxFOp/QVM+6imKHad91M=
 github.com/urfave/cli/v2 v2.3.0/go.mod h1:LJmUH05zAU44vOAcrfzZQKsZbVcdbOG8rtL3/XcUArI=
 golang.org/x/net v0.0.0-20201224014010-6772e930b67b h1:iFwSg7t5GZmB/Q5TjiEAsdoLDrdJRC1RiF2WhuV29Qw=
 golang.org/x/net v0.0.0-20201224014010-6772e930b67b/go.mod h1:m0MpNAwzfU5UDzcl9v0D8zg8gWTRqZa9RBIspLL5mdg=
+golang.org/x/net v0.0.0-20210119194325-5f4716e94777 h1:003p0dJM77cxMSyCPFphvZf/Y5/NXf5fzg6ufd1/Oew=
 golang.org/x/net v0.0.0-20210119194325-5f4716e94777/go.mod h1:m0MpNAwzfU5UDzcl9v0D8zg8gWTRqZa9RBIspLL5mdg=
 golang.org/x/sys v0.0.0-20201119102817-f84b799fce68/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/term v0.0.0-20201126162022-7de9c90e9dd1/go.mod h1:bj7SfCRtBDWHUb9snDiAeCFNEtKQo2Wmx5Cou7ajbmo=


jacobbednarz (changes requested):
thanks for the PR @jsha however in it's current form, I'm going to have to decline as there are other considerations here which this change will break.

the biggest is that there is functionality further down which is a classified as a client error however needs to be handled separately to other 4xx errors.

to help me better understand your intentions here, what issues does this solve? to be clear here, unhandled 4xx already fall into the `default` statement of this block you've updated but it will not be parsed into JSON.

jsha:
Ah, thanks for the reminder! I should have included expected behavior, etc.

Steps to reproduce:
1. Get an "API Token" (not an "API Key") from Cloudflare, and export it as CF_API_TOKEN.
2. Run `flarectl firewall rules create --zone crud.net --mode=block --value='{"target":"ip","value","198.51.100.4"}'`

Actual result:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x7cf854]

goroutine 1 [running]:
main.firewallAccessRuleCreate(0xc000072f40, 0x6, 0xa)
        /home/jsha/go/src/github.com/cloudflare/cloudflare-go/cmd/flarectl/firewall.go:124 +0x4d4
github.com/urfave/cli/v2.(*Command).Run(0xc0001a6c60, 0xc000072d00, 0x0, 0x0)
        /home/jsha/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/command.go:163 +0x4ed
github.com/urfave/cli/v2.(*App).RunAsSubcommand(0xc00010b520, 0xc000072b40, 0x0, 0x0)
        /home/jsha/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/app.go:434 +0xa9f
github.com/urfave/cli/v2.(*Command).startApp(0xc0001a6a20, 0xc000072b40, 0x8bc53e, 0x7ffe8775f3a7)
        /home/jsha/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/command.go:278 +0x69b
github.com/urfave/cli/v2.(*Command).Run(0xc0001a6a20, 0xc000072b40, 0x0, 0x0)
        /home/jsha/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/command.go:94 +0x98d
github.com/urfave/cli/v2.(*App).RunAsSubcommand(0xc00010b380, 0xc000072a80, 0x0, 0x0)
        /home/jsha/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/app.go:434 +0xa9f
github.com/urfave/cli/v2.(*Command).startApp(0xc0001a6900, 0xc000072a80, 0x8bed16, 0x7ffe8775f39e)
        /home/jsha/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/command.go:278 +0x69b
github.com/urfave/cli/v2.(*Command).Run(0xc0001a6900, 0xc000072a80, 0x0, 0x0)
        /home/jsha/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/command.go:94 +0x98d
github.com/urfave/cli/v2.(*App).RunContext(0xc00010b1e0, 0x945c60, 0xc000016108, 0xc000012080, 0x8, 0x8, 0x0, 0x0)
        /home/jsha/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/app.go:313 +0x81f
github.com/urfave/cli/v2.(*App).Run(...)
        /home/jsha/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/app.go:224
main.main()
        /home/jsha/go/src/github.com/cloudflare/cloudflare-go/cmd/flarectl/flarectl.go:686 +0x4870
```

Expected result:

```
error from makeRequest: HTTP status 405: API Tokens are not supported by this API for now
```

jacobbednarz:
thanks, that's helpful to know. let me have a dig and see what i can come up with that doesn't involve changing all the request handlers at once.

jacobbednarz:
using your change locally doesn't fix this for me however the cause of the bug is that we aren't returning once we hit the error and `resp.Result` is nil.

looking through other parts of flarectl, everything else uses the following for error handling (which works)

```go
fmt.Fprintln(os.Stderr, "error message here", err)
return err
```

i think we update the error handling to match this convention to address the cause. i'm happy to push up a PR for you to try out.

jsha:
That would be great. Thanks for looking into it! FWIW, locally this change does fix the issue for me: it's what produced the "expected output above."

Issue #587, opened 2021-02-05
Surface JSON error code in Error type


jsha:
Feature request:

Right now the Error type looks like this:

```
type Error interface {
	error
	// Raised when user credentials or configuration is invalid.
	User() bool
	// Raised when a parsing error (e.g. JSON) occurs.
	Parse() bool
	// Raised when a network error occurs.
	Network() bool
}
```

I would like it if this also had a `Code() int` method that returned the numeric error code for API errors. For instance, right now when I submit a duplicate firewall rule I get this output:

```
error from makeRequest: HTTP status 400: content "{\n  \"result\": null,\n  \"success\": false,\n  \"errors\": [\n    {\n      \"code\": 10202,\n      \"message\": \"filter at index (0) conflicts with an existing filter\",\n      \"source\": {\n        \"pointer\": \"/0/filter\"\n      },\n      \"meta\": {\n        \"id\": \"xxx\"\n      }\n    }\n  ],\n  \"messages\": null\n}\n"
```

I'd like to be able to access that `"code": 10202` field and conclude that this is just an "already exists" error and suppress it.

jacobbednarz:
released in [v0.14.0](https://github.com/cloudflare/cloudflare-go/releases/tag/v0.14.0)

Issue #574, opened 2021-01-08
Error on duplicate firewall rules - no way of checking the error code


FireFart:
Currently we try to do automated blocking on cloudflare by creating IP access rules.
Example:
```golang
resp, err := c.client.CreateAccountAccessRule(accountID, cloudflare.AccessRule{
		Mode:  "block",
		Notes: comment,
		Configuration: cloudflare.AccessRuleConfiguration{
			Target: target,
			Value:  ip,
		},
	})
```

The problem is this rules takes some time to propagate and if we try to block the same ip again, the server returns a HTTP 400 with the following body

```json
{
  "result": null,
  "success": false,
  "errors": [
    {
      "code": 10009,
      "message": "firewallaccessrules.api.duplicate_of_existing"
    }
  ],
  "messages": []
}
```

## Expected behavior

I would expect that the `Response` object returned by the client contains `Success` = false and the error in the `Error` object

## Current behavior

An error is returned as the HTTP status is 400 with the following content:
```
error from makeRequest: HTTP status 400: content "{\n  \"result\": null,\n  \"success\": false,\n  \"errors\": [\n    {\n      \"code\": 10009,\n      \"message\": \"firewallaccessrules.api.duplicate_of_existing\"\n    }\n  ],\n  \"messages\": []\n}\n"
```

https://github.com/cloudflare/cloudflare-go/blob/a71461723e375f9fb4126b5d973b2619bb010f27/cloudflare.go#L281

This makes it really hard to check the returned error code. I think it would be best to either return a custom error with the error information that can be checked using `errors.Is` or `errors.As` or the response object is properly populated (but this might cause some errors on existing implementations when people rely on an returned error instead of checking the `Success` flag)

jacobbednarz:
Totally agree this is harder than it needs to be in some cases. It's definitely on my radar to address now that we have better error handling available in Go.

For now, I recommend you use something like `strings.Contains()` and check for a known bad attribute such as `\"success\": false` or the code to accomplish the outcome you need.

```go
if err != nil {
		// check for known error codes as strings; otherwise retry the request
		if strings.Contains(err.Error(), "\\\"code\\\": 10009") || strings.Contains(err.Error(), "firewallaccessrules.api.duplicate_of_existing") {
			// skip the operation or accept you've got a duplicate
		}
	}
```

jacobbednarz:
[...]

Issue #502, opened 2020-07-10
Unclear error message from using cloudflare API


Lilja:
I'm currently trying to set up letsencrypt certificate with cloudflare's dns through [traefik](https://github.com/containous/traefik).

traefik uses [go-acme/lego](https://github.com/go-acme/lego) [which apparently uses this SDK](https://github.com/go-acme/lego/issues/1213#issuecomment-656335307). 

I'm attempting to figure out what the error means. I think it's somewhat unclear which of the two errors are the actual one.

```
time="2020-07-09T10:46:19Z" level=error msg="Unable to obtain ACME certificate for domains \"dersand.net,*.dersand.net\" : unable to generate a certificate for the domains [dersand.net *.dersand.net]: error: one or more domains had a problem:\n[*.dersand.net] [*.dersand.net] acme: error presenting token: cloudflare: failed to find zone net.: ListZonesContext command failed: error from makeRequest: HTTP status 403: insufficient permissions\n[dersand.net] [dersand.net] acme: error presenting token: cloudflare: failed to find zone net.: ListZonesContext command failed: error from makeRequest: HTTP status 403: insufficient permissions\n" providerName=dnscloudflare.acme
```

To use the cloudflare provider with traefik, I use two environment variables: `CF_DNS_API_TOKEN`, and `CF_ZONE_API_TOKEN`.

According to the [configuration page of go-acme/lego](https://go-acme.github.io/lego/dns/cloudflare/#credentials), `CF_DNS_API_TOKEN` is supposed to have DNS:Edit permission and `CF_ZONE_API_TOKEN` is supposed to have Zone:Read permission.

Here are my API key permissions in cloudflare, taken from cloudflare login -> dashboard -> get your api token -> specific token(three dots button) -> View summary:

![permissions](https://i.imgur.com/9p3HR9x.png)

The permissions are very broad, I'd say it would probably work with what's given to it.

deuill:
Indeed the error comes from the Cloudflare API, though I believe this might be an issue with configuration on your side, based on the error message:

```
acme: error presenting token: cloudflare: failed to find zone net.
```

It seems that `lego` is trying to match against the `net.` FQDN? Tracing the code takes me here:

https://github.com/go-acme/lego/blob/6043c9347ca44682aa5d01b0f8711d1f2192cba9/providers/dns/cloudflare/cloudflare.go#L125-L133

And here:

[...]

Issue #414, opened 2020-01-20
api.DeleteDNSRecord cause error HTTP status 401: invalid credentials


cytown:
It's fine just few days before, but cause error today, dunno why, and other api such as api.ZoneIDByName and api.DNSRecords as well as api.CreateDNSRecord will works just fine, but only the api.DeleteDNSRecord will cause error:
> error from makeRequest: HTTP status 401: invalid credentials

## Expected behavior

Expect it will work without any error.

## Current behavior

Other api such as api.ZoneIDByName and api.DNSRecords as well as api.CreateDNSRecord will works just fine, but only the api.DeleteDNSRecord will cause error:
> error from makeRequest: HTTP status 401: invalid credentials

## Possible solution

Not yet.

## Steps to reproduce (for bugs)

create a new dns entry A record, and then use this api to delete that one, will reproduce.

## Your Environment

Include as many relevant details about the environment you experienced
the bug in:

* Tried 0.10.6 and update to 0.11.0 today, all fails
* Go 1.13.3, mac 10.14.6.

p.s. tried debug and found it will return EOF error, hope will helps

jacobbednarz:
I haven't been able to reproduce this using the following steps.

```go
package main

import (
	"fmt"
	"log"
	"os"

	"github.com/cloudflare/cloudflare-go"
)

func main() {
	api, err := cloudflare.New(os.Getenv("CF_API_KEY"), os.Getenv("CF_API_EMAIL"))
	if err != nil {
		log.Fatal(err)
	}

	zoneID := os.Getenv("CF_ZONE_ID")
	record := cloudflare.DNSRecord{
		Name:    "test",
		Content: "example.com",
		Type:    "CNAME",
	}

	r, cErr := api.CreateDNSRecord(zoneID, record)
	if cErr != nil {
		log.Fatal(cErr)
	}

	fmt.Println(fmt.Sprintf("%+v", r))

	dErr := api.DeleteDNSRecord(zoneID, r.Result.ID)
	if dErr != nil {
		log.Fatal(err)
	}

}
```

```
$ CF_ZONE_ID="redacted" CF_API_EMAIL="redacted" CF_API_KEY="redacted" go run my_file.go

[...]

PR #596, opened 2021-02-24
Better error handling for API errors

jacobbednarz:
This introduces a change to how we raise and interact with exceptions
raised by the API responses.

For quite some time, interacting with the API response errors has been
difficult due to them spitting out parsed JSON (as a string). This means
if you need to check the type of exception in order to recover or retry,
you need to use an unreliable `strings.Contains` check. This also
extends to internally when we want to show the user an error, we had a
`errorFromResponse` function which would return the HTTP status code and
whatever was in the `errors` array.

To improve the situation I'm proposing a revamp of our error handling by
exposing various methods on `APIRequestError` struct as helpers to get
the information clients may need to better handle errors. Here are some
examples:

### Return the whole error message

```go
fmt.Println(err.Error())

// => "HTTP status 400: Authentication error (10000)"

// with multiple errors returned

fmt.Println(err.Error())

// => "HTTP status 303: test (10000), test1 (10001)"
```

### Return only the HTTP status

```go
fmt.Println(err.(*cloudflare.APIRequestError).HTTPStatusCode())

// => 400
```

### Check if the response was rate limited

```go
fmt.Println(err.(*cloudflare.APIRequestError).ClientRateLimited())

// => false
```

### Fetch just the internal error codes 

```go
e := APIRequestError{
  StatusCode: 123,
  Errors: []ResponseInfo{
    {
      Message: "test",
      Code:    10000,
    },
    {
      Message: "test1",
      Code:    10001,
    },
  },
}

fmt.Printf("%#v", e.InternalErrorCodes())

// => []int{10000, 10001}

// or

fmt.Println(err.(*cloudflare.APIRequestError).InternalErrorCodes())

// => []int{10000, 10001}
```

In order to achieve this, we'll need to update almost every method that
makes an API call to reflect the following (example from
`createAccessRule` in firewall.go):

```diff
@@ -218,7 +218,7 @@ func (api *API) createAccessRule(prefix string, accessRule AccessRule) (*AccessR
        uri := prefix + "/firewall/access_rules/rules"
        res, err := api.makeRequest("POST", uri, accessRule)
        if err != nil {
-               return nil, errors.Wrap(err, errMakeRequestError)
+               return nil, err
        }

        response := &AccessRuleResponse{}
```

Personally, I think this is a better situation for the maintainers and
consumers of this library as we'll be better equipped to handle
differing error scenarios. The Terraform provider will also benefit from
these changes for things like retries.

Recent issues related to this (more upstream in Terraform):

- #587
- #574
- #502
- #414

I haven't gone ahead and fixed any tests yet but if we agree this is 
a good move, I'll take the usual steps with CI.


jacobbednarz:
@jsha @FireFart would you folks mind chucking your 2 cents in here as well whether this jives for your uses?

cloudflare.go
```diff
@@ -260,30 +260,26 @@ func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, u
...
+
+		errBody := &Response{}
+		err = json.Unmarshal(respBody, &errBody)
+		if err != nil {
+			return nil, errors.Wrap(err, errUnmarshalErrorBody)
+		}
+
+		return nil, &APIRequestError{
+			StatusCode: resp.StatusCode,
+			Message:    errBody.Errors[0].Message,
```
patryk, line 280: There are several API endpoints that output multiple error codes/messages in the single response. It would be great to expose them.

patryk (changes requested):

cloudflare.go
jacobbednarz: sure thing, i've updated the usage to account for multiples in 0535f9b8616dba2f0baae5f70e877e4ab2b8c6db.

patryk (approved):
Love the idea of concatenating multiple messages while stringifying.

jacobbednarz:
Awesome, let me add some test coverage and update the methods themselves and we can get this shipped.

errors.go
```diff
@@ -1,52 +1,96 @@
...
 }
 
-// Parse error.
-func (e *UserError) Parse() bool {
-	return true
+// ClientError returns a boolean whether or not the raised error was caused by
+// something client side.
+func (e *APIRequestError) ClientError() bool {
+	return e.StatusCode >= http.StatusBadRequest &&
+		e.StatusCode <= http.StatusInternalServerError &&
```
FireFart, line 88: shouldn't this be `<` instead of `<=`?

cloudflare.go
```diff
@@ -260,30 +260,25 @@ func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, u
...
-		return nil, errors.Errorf("%s", respBody)
-	default:
-		var s string
-		if respBody != nil {
-			s = string(respBody)
+	if resp.StatusCode >= http.StatusBadRequest {
+		if strings.HasSuffix(resp.Request.URL.Path, "/filters/validate-expr") {
+			return nil, errors.Errorf("%s", respBody)
+		}
+
```
FireFart, line 267: I wonder if we should still raise errors on all 4xx status codes (https://api.cloudflare.com/#getting-started-responses) but with the error object populated. We could define some public Error types for all those cases mentioned in the docs and return them so the user can check against using the "new" errors methods. For example return a ErrRateLimit in case of a rate limit.
This would also make this change more backwards compatible. With the current implemention I think it would be best to release a v2 so updates will not break existing implementations.

FireFart (commented):
Thanks for the quick implementation @jacobbednarz !
Left you some comments on the changes

errors.go
jacobbednarz: good catch, it is! I'll push a fix next week for this.

cloudflare.go
jacobbednarz: I'll have a think about the error types and see what I come up with. I think that will be useful for some but wrapping them all up in the same type with helper methods will be slightly more accessible than dealing with Go's types. 

As for compatibility, we haven't yet released 1.x so the API should still be considered in development with breaking changes. However, we're not using that as an excuse to break the world and instead we are going to do a minor version bump to 0.14 with release notes that we've introduced some breaking changes to keep people across it.

jacobbednarz: I've had a think about this and I can't quite see any benefit of either approach at the moment. Let's assume Go 1.13 for simplicity, here are the two comparable cases for rate limiting.

**checking err types**

```go
if errors.Is(err, cloudflare.ClientRatedLimited) {
  // perform some backoff operation
  return
}

if err != nil {
  // .. do something else
}
```

**checking with helper methods**
 
```go
// check with helper methods
if errors.Is(err, *cloudflare.APIRequestError) && err.(*cloudflare.APIRequestError).ClientRateLimited() {
  // perform some backoff operation
}
```

The other thing to take into account here is that even if we defined an err interface for all the error types to share, we'd also need to factor in returning multiple errors at once, together which seems like it's getting into murky waters for what improvements we're trying to implement here.

At this stage, I'm inclined to leave this as is unless we can find some benefits/examples to tip the scales in favour of it.

errors.go
jacobbednarz: updated in 0afe83f

jacobbednarz:
@FireFart for your specific use case raised in #574, I envision your code looking something like this:

```go
resp, err := c.client.CreateAccountAccessRule(accountID, cloudflare.AccessRule{
  Mode:  "block",
  Notes: comment,
  Configuration: cloudflare.AccessRuleConfiguration{
    Target: target,
    Value:  ip,
  },
})

// BYO contains function 
if err != nil {
  if contains(err.(*APIRequestError).InternalErrorCodes(), 10009) {
    // fine to ignore
  } else {
    // better look into the error
  }
}

// ...or just access the index position directly
if err != nil {
  if err.(*APIRequestError).InternalErrorCodes()[0] == 10009 {
    // fine to ignore
  } else {
    // better look into the error
  }
}
```

asasr (commented):
fhad1012@icloud.com

asasr (commented):
fhad1012@icloud.com

asasr:
fhad1012@icloud.com

Issue #599, opened 2021-03-02
Possible deadlock for getting ListZones


trilobit:
When we use method [zone.ListZones()](https://github.com/cloudflare/cloudflare-go/blob/master/zone.go#L356) without arguments we can get a deadlock. 

![Screenshot 2021-03-02 at 18 58 34](https://user-images.githubusercontent.com/1970047/109678174-6def4500-7b8b-11eb-9e18-f8d4d48c3a17.png)

## Expected behavior

I guess the method don't need to work with channel and return any value. And it should call wg.Done always.

## Current behavior

A method send GET request at least twice and if it receive error in [second call](https://github.com/cloudflare/cloudflare-go/blob/master/zone.go#L399), [wg.Done()](https://github.com/cloudflare/cloudflare-go/blob/master/zone.go#L417) will not be called. And wg.Wait() never stops.

## Possible solution

I'll link a pull request to fix this issue.

PR #600, opened 2021-03-02
fix for possible deadlock in zone.ListZones

trilobit:
When we use method zone.ListZones() without arguments we can get a deadlock.

## Description

A method send GET request at least twice and if it receive error in second call, wg.Done() will not be called. And wg.Wait() never stops. [issue](https://github.com/cloudflare/cloudflare-go/issues/599)

## Types of changes

## Checklist:


jacobbednarz:
🤔 I'm not sure we want to remove the error handling reporting back to the channel as otherwise we don't get error information returned should it encounter it.

trilobit:
The channel was used only inside the goroutine and nobody reads return value outside. Now, the error will be written into the variable defined outside the goroutine and can be checked after `wg.Wait()`.

PR #604, opened 2021-03-05
always marshal CustomHostnameSSL.Wildcard into api request body payload

nicktate:
## Description

When the `CustomHostnameSSL` field is converted to a json payload for the http request body, the `Wildcard` value is stripped when set to `false`. 

* I tested updating a custom hostname wildcard from `true` to `false` with my branch and the update succeeded
* The modified tests also succeeded via `go test .`

## Types of changes

> Previously omitting the `Wildcard` key from the `CustomHostname` struct would effectively execute a partial update and omit the update of the wildcard field. Custom hostname updates will now always set the wildcard to the value of the passed in struct (whether or not explicitly referenced).

## Checklist:

### DIFF
--- a/custom_hostname.go
+++ b/custom_hostname.go
@@ -52,7 +52,7 @@ type CustomHostnameSSL struct {
 	Type                 string                              `json:"type,omitempty"`
 	CnameTarget          string                              `json:"cname_target,omitempty"`
 	CnameName            string                              `json:"cname,omitempty"`
-	Wildcard             bool                                `json:"wildcard,omitempty"`
+	Wildcard             *bool                               `json:"wildcard,omitempty"`
 	CustomCertificate    string                              `json:"custom_certificate,omitempty"`
 	CustomKey            string                              `json:"custom_key,omitempty"`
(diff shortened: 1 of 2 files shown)


jacobbednarz:
I think we may need a similar approach to #595 whereby we use a boolean pointer and explicitly set the value when it is expected. IIRC, this puts tools like Terraform in a bad position to handle changes as is.

As this is breaking change (and we've just cut 0.14 which was also breaking) we will need to wait for another breaking release to handle this one.

nicktate:
> I think we may need a similar approach to #595 whereby we use a boolean pointer and explicitly set the value when it is expected. IIRC, this puts tools like Terraform in a bad position to handle changes as is.
> 
> As this is breaking change (and we've just cut 0.14 which was also breaking) we will need to wait for another breaking release to handle this one.

Updated for a `*bool` -- good callout on the terraform implications

jacobbednarz:
Thanks for this @nicktate. I suspect this will be going out in our next release as 0.15.

PR #614, opened 2021-04-07
Adds GetWorker route API (singular) endpoint

cdloh:
## Description

Adds GetWorkerRoute API. https://api.cloudflare.com/#worker-routes-get-route

Wasn't entirely sure on the naming of the function and kept it same as the API endpoint.

I've added testing code however I noted that the other tests have a normal zone vs Enterprise zone JSON blobs. I don't think this API returns any different data based on zone type but I was unable to test this.

## Types of changes

## Checklist:

### DIFF
--- a/workers.go
+++ b/workers.go
@@ -696,7 +696,7 @@ func (api *API) ListWorkerRoutes(ctx context.Context, zoneID string) (WorkerRout
 	return r, nil
 }
 
-// GetWorkerRoute returns a worker route
+// GetWorkerRoute returns a worker route.
 //
 // API reference: https://api.cloudflare.com/#worker-routes-get-route
 func (api *API) GetWorkerRoute(ctx context.Context, zoneID string, routeID string) (WorkerRouteResponse, error) {
--- a/workers.go
+++ b/workers.go
@@ -696,6 +696,23 @@ func (api *API) ListWorkerRoutes(ctx context.Context, zoneID string) (WorkerRout
 	return r, nil
 }
 
+// GetWorkerRoute returns a worker route
+//
+// API reference: https://api.cloudflare.com/#worker-routes-get-route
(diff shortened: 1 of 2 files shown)


workers_test.go
```diff
@@ -818,6 +828,28 @@ func TestWorkers_ListWorkerRoutesEnt(t *testing.T) {
 	}
 }
 
+func TestWorkers_GetWorkerRoute(t *testing.T) {
+	setup()
+	defer teardown()
+
+	mux.HandleFunc("/zones/foo/workers/routes/1234", func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, http.MethodGet, r.Method, "Expected method 'GET', got %s", r.Method)
+		w.Header().Set("content-type", "application-json")
```
jacobbednarz, line 837: this is an invalid content-type according to [rfc7231 § 3.1.1.1](https://tools.ietf.org/html/rfc7231#section-3.1.1.1) but it's somehow working 🤷‍♂️

jacobbednarz:
> I've added testing code however I noted that the other tests have a normal zone vs Enterprise zone JSON blobs. I don't think this API returns any different data based on zone type but I was unable to test this.

The purpose of these separate tests was that in the past, only Enterprise zones had the named scripts entitlement. That is no longer the case so I think we're 👌 to ignore that now.

workers.go
```diff
@@ -696,6 +696,23 @@ func (api *API) ListWorkerRoutes(ctx context.Context, zoneID string) (WorkerRout
 	return r, nil
 }
 
+// GetWorkerRoute returns a worker route
```
jacobbednarz, line 699:
```suggestion
// GetWorkerRoute returns a worker route.
```

jacobbednarz:
thanks for this one @cdloh! 🤜🤛

PR #623, opened 2021-05-04
Use WithContext instead of NewRequestWithContext for go 1.11 support

Gobd:
This change is needed to support building with older versions of go. NewRequestWithContext was added in go 1.13 and there is an easy to use alternative that works all the way back to 1.7.

I agree that it's crazy to still be using 1.11.

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -288,10 +288,12 @@ func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, u
 // *http.Response, or an error if one occurred. The caller is responsible for
 // closing the response body.
 func (api *API) request(ctx context.Context, method, uri string, reqBody io.Reader, authType int, headers http.Header) (*http.Response, error) {
-	req, err := http.NewRequestWithContext(ctx, method, api.BaseURL+uri, reqBody)
+	req, err := http.NewRequest(method, api.BaseURL+uri, reqBody)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


jacobbednarz:
@Gobd can you please update the description to include some context as to why this change is necessary and what problem this solves? In it's current state I'm hesitant to even consider as it mentions supporting Go 1.11 which we are well beyond.

jacobbednarz:
Thanks for the update and context. I can definitely sympathise with you on this being a long time supporter of older software but I'm unsure whether swapping this out is a good move considering the [documentation for `http.WithContext`](https://github.com/golang/go/blob/62a87f64b97d27e79d53a84e9006425aa234d7e6/src/net/http/request.go#L343-L363) recommends the current approach and I wouldn't want to go against this unless we had a really good reason (such as major breakages or regressions) not to.

```
// WithContext returns a shallow copy of r with its context changed
// to ctx. The provided ctx must be non-nil.
//
// For outgoing client request, the context controls the entire
// lifetime of a request and its response: obtaining a connection,
// sending the request, and reading the response headers and body.
//
// To create a new request with a context, use NewRequestWithContext.
// To change the context of a request, such as an incoming request you
// want to modify before sending back out, use Request.Clone. Between
// those two uses, it's rare to need WithContext.
```

If `http.WithContext` isn't the recommended approach, I wonder how long it would be until we are reverting this in favour of the preferred approach when it is deprecated and whether it is worth the effort in doing so.

PR #626, opened 2021-05-06
Make the URI formatting style consistent across the codebase

itsmatinx:
This PR makes the URI formatting style consistent across the codebase.

## Description

The current codebase uses multiple styles for formatting the URI. This PR refactors all of these approaches and migrates to a single consistent one.

The changes has been tested through the project's tests. However, some of the changes that aren't covered in these tests,  were manually tested with accuracy.

### DIFF
--- a/access_application.go
+++ b/access_application.go
@@ -84,7 +84,7 @@ func (api *API) accessApplications(ctx context.Context, id string, pageOpts Pagi
 
 	uri := fmt.Sprintf("/%s/%s/access/apps", routeRoot, id)
 	if len(v) > 0 {
-		uri = uri + "?" + v.Encode()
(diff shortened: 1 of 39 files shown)


jacobbednarz:
Thanks for this. I'm a little hesitant to merge this due to the scope of
it being so wide reaching but give me a few days to go through this with
a fine tooth comb and make sure we're not introducing a regression with
this mass change.

itsmatinx:
> Thanks for this. I'm a little hesitant to merge this due to the scope of it being so wide reaching but give me a few days to go through this with a fine tooth comb and make sure we're not introducing a regression with this mass change.

You're welcome. I understand it's a pretty huge change! So with that, take as much time as you need for review and let me know. Thank you in advance.

PR #630, opened 2021-05-11
Fix misuses of the `assert.Equal()` in HTTP methods tests

itsmatinx:
This PR fixes misuses of the `assert.Equal()` in HTTP methods tests.

## Description

According to the [testify documentation](https://pkg.go.dev/github.com/stretchr/testify), the "assert.Equal()" method for check the equality of the values accepts the "TestingT", "expected", "actual", and variadic of "msgAndArgs" arguments in the defined order. The problem is the project in some tests that are dealing with the HTTP methods, does not follow the correct argument order and uses "expected" and "actual" arguments in reverse order. This PR fixes this issue.

## Types of changes

## Checklist:

### DIFF
--- a/access_application_test.go
+++ b/access_application_test.go
@@ -15,7 +15,7 @@ func TestAccessApplications(t *testing.T) {
 	defer teardown()
 
 	handler := func(w http.ResponseWriter, r *http.Request) {
-		assert.Equal(t, r.Method, http.MethodGet, "Expected method 'GET', got %s", r.Method)
+		assert.Equal(t, http.MethodGet, r.Method, "Expected method 'GET', got %s", r.Method)
 		w.Header().Set("content-type", "application/json")
 		fmt.Fprintf(w, `{
 			"success": true,
@@ -87,7 +87,7 @@ func TestAccessApplication(t *testing.T) {
 	defer teardown()
 
 	handler := func(w http.ResponseWriter, r *http.Request) {
-		assert.Equal(t, r.Method, http.MethodGet, "Expected method 'GET', got %s", r.Method)
+		assert.Equal(t, http.MethodGet, r.Method, "Expected method 'GET', got %s", r.Method)
 		w.Header().Set("content-type", "application/json")
 		fmt.Fprintf(w, `{
 			"success": true,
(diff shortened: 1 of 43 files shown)


jacobbednarz:
Does this actually cause any issues with the tests? The comparison is still being performed on the two values and the third argument tells you what is not correct in the two in the event of a failure.

itsmatinx:
> Does this actually cause any issues with the tests? The comparison is still being performed on the two values and the third argument tells you what is not correct in the two in the event of a failure.

No, it's not causing any serious issues with the tests, and everything still fine without these changes. But regardless of supplying the third argument, testify still gonna show their output based on the "expected" and "actual" arguments. With that, this may end up making some wrong assumptions.

For example, without these changes, I intentionally made one of the tests end up failing. Take a look at the output. You will find out why this is cloud be a problem.

````
FAIL: TestCreateLoadBalancerPool (0.00s)
    load_balancing_test.go:19:
                Error Trace:    load_balancing_test.go:19
                                                        server.go:2069
                                                        server.go:2448
                                                        server.go:2887
                                                        server.go:1952
                                                        asm_amd64.s:1371
                Error:          Not equal:
                                expected: "GET"
                                actual  : "POST"
                          
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -GET
                                +POST
                Test:           TestCreateLoadBalancerPool
                Messages:       Expected method 'POST', got GET
FAIL
FAIL    github.com/cloudflare/cloudflare-go     0.257s
?       github.com/cloudflare/cloudflare-go/cmd/flarectl        [no test files]
FAIL
```

jacobbednarz:
Sure, I can see where the confusion comes in. I was just making sure this is prioritised accordingly. If this was covering up potential bugs, we could throw more people power at auditing the tests and assertions but if this is just a bit of a mismatch in the output, we can address that in our own time. 

To get CI green, you'll need to update from the upstream master branch as I've just fixed an issue with the CI runners as well.

itsmatinx:
> Sure, I can see where the confusion comes in. I was just making sure this is prioritised accordingly. If this was covering up potential bugs, we could throw more people power at auditing the tests and assertions but if this is just a bit of a mismatch in the output, we can address that in our own time.

Yeah, that's understandable. I did the same mistake in another project and after I took a look at the test files of this project, the first thing that caught my eyes was this issue. Generally, it's not a big deal and as I said everything still fine without these changes, just it's better to do that in the right way.

> To get CI green, you'll need to update from the upstream master branch as I've just fixed an issue with the CI runners as well.

Done.

jacobbednarz:
Have you pushed your fixed commits? CI still seems stuck on the issue I fixed on master.

itsmatinx:
> Have you pushed your fixed commits? CI still seems stuck on the issue I fixed on master.

Previously I updated from the upstream master in Github itself. This time I manually updated, and the CI appears to have passed.

jacobbednarz:
Sweet! Thanks!

SexApeal69:
I would like to ask the same question ❓ because everything seems to go just right until I get to the (https://)
Last question of the page and for some reason idk what to put can someone please help me

PR #653, opened 2021-06-05
Add new UnsuccessfulResponseError struct for unsuccessful API responses

jn-lp:
## Description

Returning nil error along with empty slice of expected struct is unconventional for Go code. We can construct an error message from slice of `message` returned in body with `status` false.

- New tests have been added for all methods of the `UnsuccessfulResponseError`
- `NoError` assertion is already in tests for `ListWAFPackages`, `ListWAFGroups`, `ListWAFRules` and `ListWAFOverrides`

## Types of changes

## Checklist:

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -119,5 +119,75 @@ func (e *APIRequestError) ErrorMessageContains(s string) bool {
 			return true
 		}
 	}
+
+	return false
+}
+
+// UnsuccessfulResponseError is an error type that is raised
+// when a server response contains a negative status in the body.
+type UnsuccessfulResponseError struct {
+	Messages []ResponseInfo
+}
+
+func (e UnsuccessfulResponseError) Error() string {
+	var errMessages []string
+
+	for _, err := range e.Messages {
+		var m string
+		if err.Message != "" {
+			m += fmt.Sprintf("%s", err.Message)
+		}
+
+		if err.Code != 0 {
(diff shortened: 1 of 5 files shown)


jacobbednarz:
Are you able to please provide an example how you got an empty error for an unsuccessful request? If `success` is false, the error exception should be constructed using the information from the `messages` API response. If `messages`, that is a service error and not something we should cover up in here.

As is, this change isn't something we want as it would cover up the more informative error messages provided by the API endpoint.

jn-lp:
Not sure how this can hide an error message from the API endpoint if, for now, all this code does in case of unsuccessful response is return a slice of empty structs.

But, with your comment, you inspired me to develop my original idea to create a new error message for unsuccessful API responses.

jacobbednarz:
Let's rewind a second; let's cover off how this scenario actually comes about and that we're solving an issue that needs solving or whether it can be deleted or handled different. 

For starters, your updated PR pretty much duplicates what we already added in #596. The interaction and intended API appears to already be covered of. 

The second point, I'm not sure _why_ we need this `!p.Success` at all. The API call itself should be handling if the `err` returned is not nil with the existing error handling.

jacobbednarz:
Closing this off as we can revisit the `!p.Success` bit in another PR but we already have this error handling in place.

PR #664, opened 2021-07-07
Simplify ZoneIDByName

favonia:
## Description

The filtering loop in `ZoneIDByName` seems to be entirely doable using the API itself. Currently, the loop checks zone names (which should always match) and account IDs after receiving the response, but both constraints were or can be enforced via `WithZoneFilters`. Filtering out unmatched account IDs through the API itself can potentially reduce the traffic as well.

Related issue: #234

The current testing framework is mocking the API response, and since this PR outsources all the filtering to the CloudFlare server, several test cases no longer make sense and were removed.

## Types of changes

The error message for ambiguous zone names has been updated, though no applications should rely on the actual messages. Happy to revert the change. (The change is isolated as a separate commit.)

## Checklist:

PS: I could not find a guideline on this, but my code was formatted with `go fmt`. Happy to update the PR.


cloudflare.go
```diff
@@ -138,28 +138,19 @@ func (api *API) SetAuthType(authType int) {
...
+	res, err := api.ListZonesContext(context.Background(), WithZoneFilters(zoneName, api.AccountID, ""))
 	if err != nil {
 		return "", errors.Wrap(err, "ListZonesContext command failed")
 	}
 
-	if len(res.Result) > 1 && api.AccountID == "" {
-		return "", errors.New("ambiguous zone name used without an account ID")
+	switch len(res.Result) {
+	case 0:
+		return "", errors.New("Zone could not be found")
```
jacobbednarz, line 148: Go error messages should be lowercased

```suggestion
		return "", errors.New("zone could not be found")
```

cloudflare.go
```diff
@@ -138,28 +138,19 @@ func (api *API) SetAuthType(authType int) {
...
+	case 1:
+		return res.Result[0].ID, nil
+	default:
+		return "", errors.New("Ambiguous zone name; an account ID might help")
```
jacobbednarz, line 152: Go error messages should be lowercased

```suggestion
		return "", errors.New("ambiguous zone name; an account ID might help")
```

cloudflare.go
favonia: Sure. (I did not change this message---it was capitalized.)

jacobbednarz:
thank you!

PR #666, opened 2021-07-08
add support for Custom Hostnames without SSL

nickysemenza:
This change updates the client to match the fact that the `ssl` is an optional (yet almost always used) attribute on the custom hostnames API. Breaking change since an existing struct field was modified.

One caveat/part I haven't implemented here is that to turn off SSL for a hostname, you have to PATCH with a payload of `{"hostname: "x", "ssl":"null"}`, quotes important, simply relying on `omitempty` (`{"hostname: "x"}`) is not enough. 

## Types of changes

## Checklist:

### DIFF
--- a/custom_hostname.go
+++ b/custom_hostname.go
@@ -74,7 +74,7 @@ type CustomHostname struct {
 	ID                        string                                  `json:"id,omitempty"`
 	Hostname                  string                                  `json:"hostname,omitempty"`
 	CustomOriginServer        string                                  `json:"custom_origin_server,omitempty"`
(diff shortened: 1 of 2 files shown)


mihirjham:
Looks good to me.
@jacobbednarz will you help us with reviews and merging?

jacobbednarz:
as we're not 1.0 yet it think this is fine as a minor release with some notice in the release notes to show that we've changed the type to a pointer. the terraform work will be the most annoying part of this change.

CryptoGodfatherVA4 (commented):
VAR5

PR #671, opened 2021-07-17
Respect context cancellation and deadlines.

favonia:
## Description

The backoff during retries should have checked the deadline/timeout set by context.

Two regression tests were added, though they are not very robust. The 2s and ~~.5s~~ 1s thresholds to detect whether the library is honoring the deadlines are in a sense arbitrary.

## Types of changes

## Checklist:

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -212,7 +212,12 @@ func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, u
 			}
 			// useful to do some simple logging here, maybe introduce levels later
 			api.logger.Printf("Sleeping %s before retry attempt number %d for request %s %s", sleepDuration.String(), i, method, uri)
-			time.Sleep(sleepDuration)
+
+			select {
+			case <-time.After(sleepDuration):
+			case <-ctx.Done():
+				return nil, errors.Wrap(ctx.Err(), "context deadline exceeded during retries")
+			}
 
 		}
 		err = api.rateLimiter.Wait(context.Background())
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -216,7 +216,7 @@ func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, u
 			select {
 			case <-time.After(sleepDuration):
 			case <-ctx.Done():
-				return nil, errors.Wrap(ctx.Err(), "context deadline exceeded during retries")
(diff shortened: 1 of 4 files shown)


favonia:
The linting error reported by golangci-lint was about
```go
0x1 << uint8(math.Ceil(float64(i/2)))
```
I did not write the expression and it's orthogonal to this PR. I would like to hear from the maintainers what to do before fixing it.

cloudflare.go
```diff
@@ -212,7 +212,12 @@ func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, u
 			}
 			// useful to do some simple logging here, maybe introduce levels later
 			api.logger.Printf("Sleeping %s before retry attempt number %d for request %s %s", sleepDuration.String(), i, method, uri)
-			time.Sleep(sleepDuration)
+
+			select {
+			case <-time.After(sleepDuration):
+			case <-ctx.Done():
+				return nil, errors.Wrap(ctx.Err(), "context deadline exceeded during retries")
```
ribasushi, line 219: @favonia `"context deadline exceeded during retries"` is a somewhat misleading message: the `ctx` might be `Done()` due to a cancel of a parent `context.WithCancel`. I suggest rephrasing ( or just not wrapping at all ).

favonia: Ah, good point. Any suggestion? I think it's still better to wrap external errors, but I am not a maintainer here, so I don't have a strong opinion.

favonia: Alright, I pushed something which hopefully addressed the issue.

jacobbednarz:
> The linting error reported by golangci-lint was about this expression:
> 
> 0x1 << uint8(math.Ceil(float64(i/2)))

golangci-lint would have picked this up as the line itself was edited in this PR. as the fix is just removing the `math.Ceil` function, i think we can get rid of that safely in this PR.

favonia:
> > The linting error reported by golangci-lint was about this expression:
> > 0x1 << uint8(math.Ceil(float64(i/2)))
> 
> golangci-lint would have picked this up as the line itself was edited in this PR. as the fix is just removing the `math.Ceil` function, i think we can get rid of that safely in this PR.

Got it. I tried to simplify the expression further.

jacobbednarz:
LGTM, thanks! 🥳

cloudflare.go
ribasushi: @favonia The wording is great thank you!

PR #672, opened 2021-07-19
Add new field json_response_enabled in waiting room

bpaquet:
A the `json_response_enabled` into the waiting room api


jacobbednarz:
thank you! this is perfect

PR #673, opened 2021-07-22
Don't fetch unneeded DNS records in UpdateDNSRecord.

favonia:
## Description

With this patch, `UpdateDNSRecord` only fetches the referred DNS record when `Name` or `Type` is empty. Previously, it always fetches the current record even if it is not used.

The original test `TestUpdateDNSRecord` was updated to reject `GET` queries. Two new tests `TestUpdateDNSRecordWithoutName` and `TestUpdateDNSRecordWithoutType` were added for the cases where a `GET` query is expected.

## Types of changes

## Checklist:

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -133,17 +133,20 @@ func (api *API) DNSRecord(ctx context.Context, zoneID, recordID string) (DNSReco
 //
 // API reference: https://api.cloudflare.com/#dns-records-for-a-zone-update-dns-record
 func (api *API) UpdateDNSRecord(ctx context.Context, zoneID, recordID string, rr DNSRecord) error {
-	rec, err := api.DNSRecord(ctx, zoneID, recordID)
-	if err != nil {
-		return err
-	}
 	// Populate the record name from the existing one if the update didn't
 	// specify it.
-	if rr.Name == "" {
-		rr.Name = rec.Name
-	}
-	if rr.Type == "" {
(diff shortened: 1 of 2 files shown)


dns.go
```diff
@@ -133,17 +133,20 @@ func (api *API) DNSRecord(ctx context.Context, zoneID, recordID string) (DNSReco
 //
 // API reference: https://api.cloudflare.com/#dns-records-for-a-zone-update-dns-record
 func (api *API) UpdateDNSRecord(ctx context.Context, zoneID, recordID string, rr DNSRecord) error {
-	rec, err := api.DNSRecord(ctx, zoneID, recordID)
```
jacobbednarz, line 136: thanks, this looks like a hangover from a80236a02adfc1fc95be394d3bc5ddf1501caa34 where we swapped from HTTP PUT => PATCH but still wanted the option for end users to only supply a single attribute.

favonia: Got it. Is there something else I should change in this PR?

jacobbednarz: Not at this stage. I'm in the process of reviewing and checking this as it's pretty critical.

jacobbednarz:
i've had a play with this and i think we're 👌 to proceed with this. the client code i've found already passes in the full record so this shouldn't be an issue.

thanks for you contribution! 🥇

PR #674, opened 2021-07-23
fix: multiple bugs in ListZones and ListZonesContext

favonia:
## Description

Current implementation of `ListZonesContext` has several issues (some shared by `ListZones`): 
1. **Deadlock:** if all workers failed, no one is receiving the error from the unbuffered `errc`.
2. **Deadlock:** `wg.Done()` might not be called when there's an error, and `wg.Wait()` will block forever in that case.
3. **Racing**: the built-in function `append` is not (obviously) thread-safe.
4. **Inefficiency**: The first page is fetched twice.
5. **(Lack of) error-reporting**: The errors from workers seem to be dropped.
6. **Code duplication**: Code seems to be duplicated in `ListZonesContext` and `ListZones`.

This PR draft attempts to fix all of the above by implementing a robust and correct `ListZonesContext` and calling `ListZonesContext` in `ListZones` when no zone names are specified.

Yes. Two test cases were added to test concurrent fetching. Another two were added to test unexpected pagination options.

## Types of changes

## Checklist:

### DIFF
--- a/zone.go
+++ b/zone.go
@@ -454,7 +454,7 @@ func (api *API) ListZonesContext(ctx context.Context, opts ...ReqOption) (r Zone
 			defer wg.Done()
 
 			opt.params.Set("page", strconv.Itoa(pageNumber))
-			res, err = api.makeRequestContext(ctx, http.MethodGet, "/zones?"+opt.params.Encode(), nil)
+			res, err := api.makeRequestContext(ctx, http.MethodGet, "/zones?"+opt.params.Encode(), nil)
 			if err != nil {
 				recordError(err)
 				return
--- a/zone.go
+++ b/zone.go
@@ -422,14 +422,6 @@ func (api *API) ListZonesContext(ctx context.Context, opts ...ReqOption) (r Zone
 		totalCount     = r.Total
 	)
 
-	// These checks should be redundant when CloudFlare works as advertised.
-	// However, this function critically relies on these conditions.
-	if len(r.Result) != pageSize || // first page does not have exactly 50 zones
-		pageSize*totalPageCount < totalCount || // too few pages
-		totalCount < pageSize*(totalPageCount-1) { // too many pages
-		return ZonesResponse{}, errors.New(errPagination)
-	}
-
 	var (
 		// The size of the last page (which would be <= 50).
 		lastPageSize = totalCount - pageSize*(totalPageCount-1)
--- a/zone.go
+++ b/zone.go
@@ -469,7 +469,7 @@ func (api *API) ListZonesContext(ctx context.Context, opts ...ReqOption) (r Zone
 
 			if (pageNumber < totalPageCount && len(r.Result) != pageSize) ||
 				(pageNumber == totalPageCount && len(r.Result) != lastPageSize) {
-				recordError(err)
+				recordError(errors.New(errPagination))
 				return
 			}
 
--- a/zone.go
+++ b/zone.go
@@ -355,13 +355,14 @@ func (api *API) ZoneActivationCheck(ctx context.Context, zoneID string) (Respons
 //
 // API reference: https://api.cloudflare.com/#zone-list-zones
 func (api *API) ListZones(ctx context.Context, z ...string) ([]Zone, error) {
-	v := url.Values{}
-
-	var res []byte
-	var r ZonesResponse
 	var zones []Zone
-	var err error
 	if len(z) > 0 {
+		var (
+			v   = url.Values{}
+			res []byte
+			r   ZonesResponse
+			err error
+		)
 		for _, zone := range z {
 			v.Set("name", normalizeZoneName(zone))
 			res, err = api.makeRequestContext(ctx, http.MethodGet, "/zones?"+v.Encode(), nil)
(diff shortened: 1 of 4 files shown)


favonia:
@jacobbednarz I think I need help in creating test cases. Some problems are very difficult to test but some seem manageable, if we have an easy way to generate 100+ zones. What we absolutely need are:

1. A test returning 50 initial zones but then failing on all subsequent requests.
2. A test returning 50 initial zones but then failing on one of the subsequent requests.
3. A test returning 50 initial zones and then the remaining 50+N (e.g., 60) zones.

I need help in generating 100+ realistic zones (and there are no existing tests for ListZones). In the meanwhile, I hope the code looks clean enough that it did not obviously introduce new bugs.

favonia:
An alternative to concurrent slice writing is to have a dedicated "collector" receiving zones from a thread-safe channel. The code will (appear) more reliable because we do not have to check exact page sizes to align all concurrent accesses. However, the order of zones would then be non-deterministic and the `order` parameter would not be respected. Let me know if you find this is more preferable.

jacobbednarz:
> Some issues are very difficult to test but some seem manageable, if we have an easy way to generate 100+ zones

i've gone ahead and sent you an invitation to my test organisation where i've created 60 test zones for you. feel free to use this account and zones for your testing. let me know if you need more added and i can spin up another handful of them for you.

> However, the order of zones would then be non-deterministic and the order parameter would not be respected.

i think we need to keep the order as it's a parameter the API accepts and people can configure to use. it would raise some questions if people provided it but the parameter wasn't respected.

zone.go
```diff
@@ -452,44 +412,88 @@ func (api *API) ListZonesContext(ctx context.Context, opts ...ReqOption) (r Zone
...
+		return r, nil
+	}
+
+	// parameters of pagination
+	var (
+		totalPageCount = r.TotalPages
+		totalCount     = r.Total
+	)
+
+	// These checks should be redundant when CloudFlare works as advertised.
```
jacobbednarz, line 425: i'm not sure i follow the comment here; what scenarios is this not the case?

favonia:
> > Some issues are very difficult to test but some seem manageable, if we have an easy way to generate 100+ zones
> 
> i've gone ahead and sent you an invitation to my test organisation where i've created 60 test zones for you. feel free to use this account and zones for your testing. let me know if you need more added and i can spin up another handful of them for you.

Thank you. I was thinking about adding some public tests to `zone_test.go` to simulate all kinds of strange and broken responses the server could give to see whether the code is robust enough. However, an interesting case would require at least 3 pages, which means we need at least 101 zones. I suppose we could programmatically generate hundreds of `wwwNNN.example.org` domains in tests, but then I decided to pause and give it some more thought.

jacobbednarz:
we can adjust the `per_page` for testing purposes and just generate a smaller number of zones? you can also dynamically create the API response used in the tests as well if you'd like.

zone.go
jacobbednarz: hmm, i think that is a little beyond the realm of this library. realistically, if Cloudflare responds with a 200, you should be safe to assume we're good for pagination parameters.

jacobbednarz:
50 per page was just a back of the napkin happy mid point for the internal pagination; it could be anything really.

i'm happy to create another 100 or so zones if you'd prefer that though?

zone.go
favonia: I like to be a bit more defensive so that the code is logically `panic`-free. On the other hand, after checking the code more carefully, it's actually the case even without these checks, so I just removed them.

favonia: The reason is that the following lines eliminate all possible conditions or memory access violations:

https://github.com/cloudflare/cloudflare-go/blob/7f2f0f88ecc2ea1f6c882428cb122007a0e88aa7/zone.go#L465-L472

jacobbednarz:
what do we have left here? only tests? the code itself looks fine so i'm 👍 to that and if we wrap this up in the next couple of days, we can put this into the next release.

favonia:
> what do we have left here? only tests? the code itself looks fine so i'm +1 to that and if we wrap this up in the next couple of days, we can put this into the next release.

Yes, only the tests. Sorry I didn't have time to add them yet.

jacobbednarz:
no worries! just checking I hadn't left you in limbo 🙂 let me know if you need anything further.

favonia:
It turns out my first attempt was also full of concurrent map writes due to capturing. I felt the potential benefits of inline functions are small (if any) so I just move the worker out to eliminate any implicit capturing. It is arguably overkilling but at least two bugs were fixed.

favonia:
> what do we have left here? only tests? the code itself looks fine so i'm +1 to that and if we wrap this up in the next couple of days, we can put this into the next release.

I am much more confident about the correctness and quality of this PR now. Feel free to merge it.

zone.go
favonia: > i'm not sure i follow the comment here; what scenarios is this not the case?

I realized at least three test cases in `cloudflare_test.go` violated these checks.

zone_test.go
```diff
@@ -2,15 +2,155 @@ package cloudflare
...
 	"net/url"
+	"strconv"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/assert"
 )
 
+// mockID returns a hex string of length 32, suitable for all kinds of IDs
+// used in the CloudFlare API.
```
jacobbednarz, line 19:
```suggestion
// used in the Cloudflare API.
```

zone_test.go
```diff
@@ -2,15 +2,155 @@ package cloudflare
...
+func mustParseTime(s string) time.Time {
+	t, err := time.Parse(time.RFC3339Nano, s)
+	if err != nil {
+		panic(err)
+	}
+	return t
+}
+
+func mockZone(i int) *Zone {
+	zoneName := fmt.Sprintf("%d.example.org", i)
```
jacobbednarz, line 34:
```suggestion
	zoneName := fmt.Sprintf("%d.example.com", i)
```

zone_test.go
```diff
@@ -2,15 +2,155 @@ package cloudflare
...
+
+func mockResultInfo(total, page, start, count int) *ResultInfo {
+	return &ResultInfo{
+		Page:       page,
+		PerPage:    50,
+		TotalPages: (total + 49) / 50,
+		Count:      count,
+		Total:      total,
+	}
+}
```
favonia, line 135: Ah! I should have avoided hardcoding the number `50` in it unless all zone-related API uses 50 as the maximum number of items per page. :confused: Let me make one final change.

favonia:
Done now!

jacobbednarz:
I think we're good here! Thanks for getting this over the line @favonia! It's great.

PR #682, opened 2021-07-31
WithPagination should ignore non-positive numbers.

favonia:
Make `WithPagination(PaginationOptions{})` a no-op.

## Description

`With...` options should ignore zero values. This has been the case for various options, but not for `WithPagination`.

Existing tests passed.

## Types of changes

## Checklist:

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -442,7 +442,12 @@ func WithZoneFilters(zoneName, accountID, status string) ReqOption {
 // WithPagination configures the pagination for a response.
 func WithPagination(opts PaginationOptions) ReqOption {
 	return func(opt *reqOption) {
-		opt.params.Set("page", strconv.Itoa(opts.Page))
-		opt.params.Set("per_page", strconv.Itoa(opts.PerPage))
+		if opts.Page != 0 {
+			opt.params.Set("page", strconv.Itoa(opts.Page))
+		}
+
+		if opts.PerPage != 0 {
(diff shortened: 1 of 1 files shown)


cloudflare.go
```diff
@@ -442,7 +442,12 @@ func WithZoneFilters(zoneName, accountID, status string) ReqOption {
 // WithPagination configures the pagination for a response.
 func WithPagination(opts PaginationOptions) ReqOption {
 	return func(opt *reqOption) {
-		opt.params.Set("page", strconv.Itoa(opts.Page))
-		opt.params.Set("per_page", strconv.Itoa(opts.PerPage))
+		if opts.Page != 0 {
```
jacobbednarz, line 445: this is great but let's make it a little better by ensuring the value is also positive 

```suggestion
		if opts.Page > 0 {
```

cloudflare.go
```diff
@@ -442,7 +442,12 @@ func WithZoneFilters(zoneName, accountID, status string) ReqOption {
...
+			opt.params.Set("page", strconv.Itoa(opts.Page))
+		}
+
+		if opts.PerPage != 0 {
```
jacobbednarz, line 449:
```suggestion
		if opts.PerPage > 0 {
```

jacobbednarz (commented):
thanks for finding this! if we're touching this code, let's take it that one step further and ensure we're only passing positive values and not allowing this library to request negative values

Issue #688, opened 2021-08-06
Normalization of DNS record names


favonia:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the library and it is still present.

### cloudflare-go version

0.20.0

### Go environment

<details><summary><code>go env</code></summary>

```
GO111MODULE=""

GOARCH="amd64"

GOBIN=""

GOCACHE="/home/***/.cache/go-build"

GOENV="/home/***/.config/go/env"

GOEXE=""

GOFLAGS=""

GOHOSTARCH="amd64"

GOHOSTOS="linux"

GOINSECURE=""

GOMODCACHE="/home/***/go/pkg/mod"

GONOPROXY=""

GONOSUMDB=""

GOOS="linux"

GOPATH="/home/***/go"

GOPRIVATE=""

GOPROXY="https://proxy.golang.org,direct"

GOROOT="/usr/lib/go"

GOSUMDB="sum.golang.org"

GOTMPDIR=""

GOTOOLDIR="/usr/lib/go/pkg/tool/linux_amd64"

GOVCS=""

GOVERSION="go1.16.6"

GCCGO="gccgo"

AR="ar"

CC="gcc"

CXX="g++"

CGO_ENABLED="1"

GOMOD="/dev/null"

CGO_CFLAGS="-g -O2"

CGO_CPPFLAGS=""

CGO_CXXFLAGS="-g -O2"

CGO_FFLAGS="-g -O2"

CGO_LDFLAGS="-g -O2"

PKG_CONFIG="pkg-config"

GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build***=/tmp/go-build -gno-record-gcc-switches"
```
</details>

### Expected output

Showing the DNS record that was just added, because the same internationalized domain name was used.

### Actual output

No record:
```
[]cloudflare.DNSRecord(nil)
```

### Code demonstrating the issue

```go
func main() {
	const UNICODE_DOMAIN = "😺.meow.org"

	handle, _ := cloudflare.NewWithAPIToken(CF_API_TOKEN)

	handle.CreateDNSRecord(context.Background(), ZONE_ID,
		cloudflare.DNSRecord{
			Name:    UNICODE_DOMAIN,
			Type:    "A",
			Content: "10.10.10.10",
		})

	rs, _ := handle.DNSRecords(context.Background(), ZONE_ID,
[...]

PR #690, opened 2021-08-06
feat(dns): normalize DNS record names

favonia:
This is a **proof of concept** of the proposed workaround in #688. We may or may not want this, depending on the discussion in #688. (I did not use the "Close" keyword because, in my opinion, a proper solution is to fix the API server instead of working around it.)

## Description

See #688 and cloudflare/cloudflare-docs#1991. The current API server has trouble handling Unicode forms of IDNs.

`CreateDNSRecord`, `DNSRecords`, and `UpdateDNSRecord` now call `normalizeDomainName` before accessing the API.

## Types of changes

## Checklist:

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -58,9 +58,6 @@ var nontransitionalLookup = idna.New(
 // from Unicode form to Punycode, using non-transitional process specified
 // in UTS 46.
 //
-// Confusinigly, this is in the opposite direction of normalizeZoneName
-// to work around the current quirks of the Cloudflare servers.
-//
 // Note: conversion errors are silently discarded and partial conversion
 // results are used.
 func normalizeDomainName(name string) string {
--- a/dns.go
+++ b/dns.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"golang.org/x/net/idna"
 )
 
 // DNSRecord represents a DNS record in a zone.
@@ -45,10 +46,34 @@ type DNSListResponse struct {
 	ResultInfo `json:"result_info"`
 }
 
+// nontransitionalLookup implements the nontransitional processing as specified in
+// Unicode Technical Standard 46 with almost all checkings off to maximize user freedom.
+var nontransitionalLookup = idna.New(
+	idna.MapForLookup(),
+	idna.StrictDomainName(false),
+	idna.ValidateLabels(false),
+)
+
+// normalizeDomainName tries to convert IDNs (international domain names)
+// from Unicode form to Punycode, using non-transitional process specified
+// in UTS 46.
+//
+// Confusinigly, this is in the opposite direction of normalizeZoneName
+// to work around the current quirks of the Cloudflare servers.
+//
+// Note: conversion errors are silently discarded and partial conversion
+// results are used.
+func normalizeDomainName(name string) string {
+	name, _ = nontransitionalLookup.ToASCII(name)
+	return name
+}
+
 // CreateDNSRecord creates a DNS record for the zone identifier.
 //
 // API reference: https://api.cloudflare.com/#dns-records-for-a-zone-create-dns-record
 func (api *API) CreateDNSRecord(ctx context.Context, zoneID string, rr DNSRecord) (*DNSRecordResponse, error) {
+	rr.Name = normalizeDomainName(rr.Name)
+
 	uri := fmt.Sprintf("/zones/%s/dns_records", zoneID)
 	res, err := api.makeRequestContext(ctx, http.MethodPost, uri, rr)
 	if err != nil {
@@ -75,7 +100,7 @@ func (api *API) DNSRecords(ctx context.Context, zoneID string, rr DNSRecord) ([]
 	// Request as many records as possible per page - API max is 100
 	v.Set("per_page", "100")
 	if rr.Name != "" {
-		v.Set("name", rr.Name)
+		v.Set("name", normalizeDomainName(rr.Name))
 	}
 	if rr.Type != "" {
 		v.Set("type", rr.Type)
@@ -133,6 +158,8 @@ func (api *API) DNSRecord(ctx context.Context, zoneID, recordID string) (DNSReco
 //
 // API reference: https://api.cloudflare.com/#dns-records-for-a-zone-update-dns-record
 func (api *API) UpdateDNSRecord(ctx context.Context, zoneID, recordID string, rr DNSRecord) error {
+	rr.Name = normalizeDomainName(rr.Name)
+
 	// Populate the record name from the existing one if the update didn't
 	// specify it.
 	if rr.Name == "" || rr.Type == "" {
--- a/dns.go
+++ b/dns.go
@@ -54,13 +54,13 @@ var nontransitionalLookup = idna.New(
 	idna.ValidateLabels(false),
 )
 
-// normalizeDomainName tries to convert IDNs (international domain names)
+// toUTS46ASCII tries to convert IDNs (international domain names)
 // from Unicode form to Punycode, using non-transitional process specified
 // in UTS 46.
 //
 // Note: conversion errors are silently discarded and partial conversion
 // results are used.
-func normalizeDomainName(name string) string {
+func toUTS46ASCII(name string) string {
 	name, _ = nontransitionalLookup.ToASCII(name)
 	return name
 }
@@ -69,7 +69,7 @@ func normalizeDomainName(name string) string {
 //
 // API reference: https://api.cloudflare.com/#dns-records-for-a-zone-create-dns-record
 func (api *API) CreateDNSRecord(ctx context.Context, zoneID string, rr DNSRecord) (*DNSRecordResponse, error) {
-	rr.Name = normalizeDomainName(rr.Name)
+	rr.Name = toUTS46ASCII(rr.Name)
 
 	uri := fmt.Sprintf("/zones/%s/dns_records", zoneID)
 	res, err := api.makeRequestContext(ctx, http.MethodPost, uri, rr)
@@ -97,7 +97,7 @@ func (api *API) DNSRecords(ctx context.Context, zoneID string, rr DNSRecord) ([]
 	// Request as many records as possible per page - API max is 100
 	v.Set("per_page", "100")
 	if rr.Name != "" {
-		v.Set("name", normalizeDomainName(rr.Name))
+		v.Set("name", toUTS46ASCII(rr.Name))
 	}
 	if rr.Type != "" {
 		v.Set("type", rr.Type)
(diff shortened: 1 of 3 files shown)


jacobbednarz:
@aaranmcguire do you mind eyeballing this and weighing in? I'm not familiar enough with our DNS API enough to pick up any potential shortcomings of it.

dns.go
```diff
@@ -45,10 +46,34 @@ type DNSListResponse struct {
...
+// normalizeDomainName tries to convert IDNs (international domain names)
+// from Unicode form to Punycode, using non-transitional process specified
+// in UTS 46.
+//
+// Confusinigly, this is in the opposite direction of normalizeZoneName
+// to work around the current quirks of the Cloudflare servers.
+//
+// Note: conversion errors are silently discarded and partial conversion
+// results are used.
+func normalizeDomainName(name string) string {
```
jacobbednarz, line 66: this seems like a poorly named method (`normalizeZoneName` isn't any better IMO). can we instead rename this to cover what the method does? i.e. `stringToUTS46ASCII()` or similar? ideally this would it's own type that we could pass a string into and perform the comparison independent of the DNS structs.

dns.go
```diff
@@ -45,10 +46,34 @@ type DNSListResponse struct {
...
+	idna.StrictDomainName(false),
+	idna.ValidateLabels(false),
+)
+
+// normalizeDomainName tries to convert IDNs (international domain names)
+// from Unicode form to Punycode, using non-transitional process specified
+// in UTS 46.
+//
+// Confusinigly, this is in the opposite direction of normalizeZoneName
+// to work around the current quirks of the Cloudflare servers.
```
jacobbednarz, line 62: this line isn't quite accurate; it's nothing to do with _servers_ but the Cloudflare DNS API service itself. if we're going to keep this line (which my vote is to remove this and instead use a better type/method structure) we need to reflect reality.

jacobbednarz: i.e. implement the [stringer interface](https://pkg.go.dev/fmt#Stringer) for these types

dns.go
jacobbednarz: i.e. implement the [stringer interface](https://pkg.go.dev/fmt#Stringer) for these types

zone.go
```diff
@@ -849,6 +849,15 @@ func (api *API) UpdateFallbackOrigin(ctx context.Context, zoneID string, fbo Fal
...
 // is returned unchanged.
 //
+// Because all the zone name comparison is currently done on the server side
+// (except for comparison with the empty string), theoretically, we could
+// remove this function and let the server decide equivalence between zone
+// names. However, there should be no harm calling this function other than
+// potential performance penality.
+//
+// Confusinigly, this is in the opposite direction of normalizeDomainName
+// to work around the current quirks of the Cloudflare servers.
```
jacobbednarz, line 859: similar comments to https://github.com/cloudflare/cloudflare-go/pull/690/files#r689215432

dns.go
favonia: > this seems like a poorly named method (`normalizeZoneName` isn't any better IMO). can we instead rename this to cover what the method does? i.e. `stringToUTS46ASCII()` or similar? ideally this would it's own type that we could pass a string into and perform the comparison independent of the DNS structs.

Will change the name. Ideally, we would not need any of these functions once the API is fixed. Should we rename or even remove `normalizeZoneName` (in a separate PR)? As far as I know, it's not needed anymore.

favonia: > i.e. implement the [stringer interface](https://pkg.go.dev/fmt#Stringer) for these types -- https://play.golang.org/p/vp00zNkQkMS

I see. How about something like this?
```go
type DNSRecordName string
```
More details: https://play.golang.org/p/GmhyEQwEFAf The benefit is that it can then be used in all different data structures.

dns.go
favonia: I will update or remove the comments accordingly. Currently, "server" is being used in `cloudflare.go` (and maybe other files) to refer to "API service" but I will leave the updates to another PR.

dns.go
favonia: > I see. How about something like this?
> 
> ```go
> type DNSRecordName string
> ```
> 
> More details: https://play.golang.org/p/GmhyEQwEFAf The benefit is that it can then be used in all different data structures.

On the second thought, maybe I should keep `string` as `string` so that the API remains the same.

favonia: So I changed `normalizeDomainName` to `toUTS46ASCII`, hopefully addressing the issue. I did not implement the `toASCII` method for those structures because I am not sure for what structures I should implement it. Also, when other domain-like data are present (e.g., emails), I am not sure if the method should apply the conversion to them as well.

favonia:
@jacobbednarz Is there something I should do now? I will soon be very occupied and unable to maintain this PR.

jacobbednarz:
Nothing from you at this point. I'll poke our internal DNS team again to have a look and confirm there won't be any side effects here.

favonia:
Sounds good! As I mentioned earlier, the most ideal solution is to fix the API itself so that it's more or less consistent with the web interface. (The web interface will do the conversion/normalization automatically.)

jacobbednarz:
Thanks, I had some eyeballing internally and it all makes sense so let's
get this merged and out with the next release.

Appreciate it!

aaranmcguire:
@favonia Would you mind sharing the API call you're doing where we don't convert the record correctly? 

I'm one of the people responsible for the API you mentioned, and we do have tests for inputting emojis, and them being converted correctly.

If you have valid input we would be happy to fix it, and then it's up to you if you want to revert or clean up this client code.

Here's an example test we have:
**Request:**
```json
{
    "zone_name": "example.com",
    "name": "😀",
    "type": "A",
    "content": "198.51.100.1",
    "proxied": true,
    "ttl": 1
}
```
**Response:**
```json
{
    "zone_name": "example.com",
    "name": "xn--e28h.example.com",
    "type": "A",
    "content": "198.51.100.1",
    "proxiable": true,
    "proxied": true,
    "ttl": 1,
    "locked": false
}
```

favonia:
> @favonia Would you mind sharing the API call you're doing where we don't convert the record correctly?

See https://github.com/cloudflare/cloudflare-docs/issues/1991#issue-962925116. In sum, I discovered at least these issues:

1. Zone API in general returns Unicode forms, while DNS record API in general returns ASCII forms. They should be more consistent in my opinion.
2. The `name` parameter of `GET zones/:zone_identifier/dns_records` does not seem to accept Unicode forms.
3. The conversion done by the DNS record API is not any standard one. It does accept `😀.example.com`, but it did not accept `ÖBB.example.com`. It still does some conversion so that `HELLO.example.com` is converted to `hello.example.com`. I am not aware of any standard describing this style of partial conversion.
4. The documentation of zone creation is problematic: the description of `pattern` needs to be updated.
5. It is unclear how all of these affect name servers, emails, and other fields contain domain-like data.

aaranmcguire:
@favonia Thanks for providing this feedback. The DNS Records API will work differently to the Zones API as they run on different languages/versions, but I will investigate consistency issues within the DNS Records API.

> The `name` parameter of `GET zones/:zone_identifier/dns_records` does not seem to accept Unicode forms.

This is a good feature request, and something we've not thought of, so I will create an internal ticket for this.

> The conversion done by the DNS record API is not any standard one. It does accept `😀.example.com`, but it did not accept `ÖBB.example.com`. It still does some conversion so that `HELLO.example.com` is converted to `hello.example.com`. I am not aware of any standard describing this style of partial conversion.

We use the following:

```go
parser := idna.New(
	idna.ValidateLabels(true),
	idna.CheckHyphens(false),
	idna.VerifyDNSLength(true),
	idna.StrictDomainName(true),
)
host, _ := parser.ToASCII(hostname)
```

So `ÖBB` likely fails somewhere in there. I can dig deeper later to confirm.

> It is unclear how all of these affect name servers, emails, and other fields containing domain-like data.

Not sure what you mean here, but the zone_name doesn't affect the name servers we provide.

favonia:
Some additional info:

> > The conversion done by the DNS record API is not any standard one. It does accept `😀.example.com`, but it did not accept `ÖBB.example.com`. It still does some conversion so that `HELLO.example.com` is converted to `hello.example.com`. I am not aware of any standard describing this style of partial conversion.
> 
> We use the following:
> 
> ```go
> parser := idna.New(
> 	idna.ValidateLabels(true),
> 	idna.CheckHyphens(false),
> 	idna.VerifyDNSLength(true),
> 	idna.StrictDomainName(true),
> )
> host, _ := parser.ToASCII(hostname)
> ```
> 
> So `ÖBB` likely fails somewhere in there. I can dig deeper later to confirm.

I actually managed to reconstruct this part of the code, and that's only part of the story. Currently, my reverse-engineering indicated that the conversion is done in two stages:

1. Doing the above conversion from Unicode to ASCII forms. (You showed this step.)
2. Converting labels into their lowercase forms. (`HELLO` -> `hello`)

The second step should have been done before the first one so that ÖBB is turned into öbb before its Punycode is computed. `idna.MapForLookup` can enable that.  Check out https://util.unicode.org/UnicodeJsps/idna.jsp for more tricky cases.

> > It is unclear how all of these affect name servers, emails, and other fields containing domain-like data.
> 
> Not sure what you mean here, but the zone_name doesn't affect the name servers we provide.

My point is that the documentation is lacking, and the issue is not limited to zones and DNS records we have been focusing on. IDNs can appear in name servers, domains of email addresses, etc. They enjoy and suffer from the same IDN issues we have to deal with for zones and DNS records. If, say, my email address uses an IDN, which form should I use for `CF_EMAIL` or any API wanting an email address? Currently there's no mentioning of IDNs on https://api.cloudflare.com/ and as an API user I can only guess and experiment to figure out which conversion algorithm (if any) is in use. Personally, I suggest adding a new section on https//api.cloudflare.com/ to describe the supposed conversion algorithm for IDNs and fix all the inconsistent or unexpected behaviors of each individual endpoint.

favonia:
@aaranmcguire I should have added that now I understand my last point would require collaboration between perhaps multiple teams in charge of different parts of the API, but such collaboration is needed. I hope that you could consider pushing for such an internal collaboration to bring forth consistent and complete support of IDNs. It would really make a difference for users like me who creates tools for your service. Thanks.

aaranmcguire:
@favonia Thanks for pointing out this bug. This should now be fixed now. I'll also push for documentation on how IDN formatting should work, but this may take some time as this would be a cross team effort.

If you find an more bugs on the DNS Records API, please feel free to ping me.

jacobbednarz:
@favonia can you confirm if this is working for you using the API directly? if so, we can revert this change.

favonia:
> @favonia can you confirm if this is working for you using the API directly? if so, we can revert this change.

Sorry I would not have time to work on this until the end of the year. Some quick thoughts:

1. It might not make sense to revert the entire PR because the new test Unicode cases should be modified to signify the normalization done by the API service instead of being completely removed. Only `toUTS46ASCII` should be purged.
2. I made sure all my tools will work correctly regardless what would happen to this PR. Given that no one else complained, I suppose we are relatively safe no matter what we do.
3. That said, the efficiency penalty seems to be very small, so this is probably not very urgent.
4. We should probably purge `normalizeZoneName` as well.

Anyways, I would not have time to touch this PR now.

favonia:
> > The `name` parameter of `GET zones/:zone_identifier/dns_records` does not seem to accept Unicode forms.
> 
> This is a good feature request, and something we've not thought of, so I will create an internal ticket for this.
> 
@aaranmcguire I tried it yesterday and it seems this feature (using IDNs in the `name` parameter) has not been implemented. So, we probably should not revert this PR yet. @jacobbednarz

PR #693, opened 2021-08-10
GATE-1703 team rules + accounts + config api

rexscaria:
The pr has newly publicized Teams api for rules + account config
## Description

This is the new teams rules api.

required.

## Types of changes

Mew features- Teams rules + accounts api

### DIFF
--- a/teams_accounts.go
+++ b/teams_accounts.go
@@ -4,9 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/pkg/errors"
 	"net/http"
 	"time"
+
+	"github.com/pkg/errors"
 )
 
 type TeamsAccount struct {
@@ -15,13 +16,15 @@ type TeamsAccount struct {
 	ID           string `json:"id"`            // cloudflare account ID
 }
 
-// TeamsAccountResponse is the API response, containing information on teams account
+// TeamsAccountResponse is the API response, containing information on teams
+// account.
 type TeamsAccountResponse struct {
 	Response
 	Result TeamsAccount `json:"result"`
 }
 
-// TeamsConfigResponse is the API response, containing information on teams account config
+// TeamsConfigResponse is the API response, containing information on teams
+// account config.
 type TeamsConfigResponse struct {
 	Response
 	Result TeamsConfiguration `json:"result"`
@@ -64,7 +67,9 @@ type TeamsBlockPage struct {
 	Name            string `json:"name,omitempty"`
 }
 
-// TeamsAccount returns teams account information with internal and external id
+// TeamsAccount returns teams account information with internal and external ID.
+//
+// API reference: TBA
 func (api *API) TeamsAccount(ctx context.Context, accountID string) (TeamsAccount, error) {
 	uri := fmt.Sprintf("/accounts/%s/gateway", accountID)
 
@@ -82,7 +87,9 @@ func (api *API) TeamsAccount(ctx context.Context, accountID string) (TeamsAccoun
 	return teamsAccountResponse.Result, nil
 }
 
-// TeamsAccountConfiguration returns teams account configuration
+// TeamsAccountConfiguration returns teams account configuration.
+//
+// API reference: TBA
 func (api *API) TeamsAccountConfiguration(ctx context.Context, accountID string) (TeamsConfiguration, error) {
 	uri := fmt.Sprintf("/accounts/%s/gateway/configuration", accountID)
 
@@ -100,7 +107,9 @@ func (api *API) TeamsAccountConfiguration(ctx context.Context, accountID string)
 	return teamsConfigResponse.Result, nil
 }
 
-// TeamsAccountUpdateConfiguration upserts teams account configuration
+// TeamsAccountUpdateConfiguration updates a teams account configuration.
+//
+// API reference: TBA
 func (api *API) TeamsAccountUpdateConfiguration(ctx context.Context, accountID string, config TeamsConfiguration) (TeamsConfiguration, error) {
 	uri := fmt.Sprintf("/accounts/%s/gateway/configuration", accountID)
 
--- a/teams_accounts.go
+++ b/teams_accounts.go
@@ -1,0 +1,119 @@
+package cloudflare
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"github.com/pkg/errors"
+	"net/http"
+	"time"
+)
+
+type TeamsAccount struct {
+	GatewayTag   string `json:"gateway_tag"`   // Internal teams ID
+	ProviderName string `json:"provider_name"` // Auth provider
+	ID           string `json:"id"`            // cloudflare account ID
+}
+
+// TeamsAccountResponse is the API response, containing information on teams account
+type TeamsAccountResponse struct {
+	Response
+	Result TeamsAccount `json:"result"`
+}
+
+// TeamsConfigResponse is the API response, containing information on teams account config
+type TeamsConfigResponse struct {
+	Response
+	Result TeamsConfiguration `json:"result"`
+}
+
+// TeamsConfiguration data model.
+type TeamsConfiguration struct {
+	Settings  TeamsAccountSettings `json:"settings"`
+	CreatedAt time.Time            `json:"created_at,omitempty"`
+	UpdatedAt time.Time            `json:"updated_at,omitempty"`
+}
+
+type TeamsAccountSettings struct {
+	Antivirus   *TeamsAntivirus   `json:"antivirus,omitempty"`
+	TLSDecrypt  *TeamsTLSDecrypt  `json:"tls_decrypt,omitempty"`
+	ActivityLog *TeamsActivityLog `json:"activity_log,omitempty"`
+	BlockPage   *TeamsBlockPage   `json:"block_page,omitempty"`
+}
+
+type TeamsAntivirus struct {
+	EnabledDownloadPhase bool `json:"enabled_download_phase,omitempty"`
+	EnabledUploadPhase   bool `json:"enabled_upload_phase,omitempty"`
+	FailClosed           bool `json:"fail_closed,omitempty"`
+}
+
+type TeamsTLSDecrypt struct {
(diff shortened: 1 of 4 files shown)


teams_accounts.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+type TeamsAccount struct {
+	GatewayID    string `json:"gateway_tag"`   // Internal teams ID
+	ProviderName string `json:"provider_name"` // Auth provider
+	CloudflareID string `json:"id"`            // cloudflare account ID
+}
+
+// TeamsAccountResponse is the API response, containing information on teams account
+type TeamsAccountResponse struct {
```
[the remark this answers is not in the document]
rexscaria: That struct can not hold the result we have. That's why I used a new struct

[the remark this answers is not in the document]
Michael9127: you can embed it:
`type TeamsAccountResponse struct {
    Response
    Result      TeamsAccount `json:"result"`
}`

teams_rules.go
```diff
@@ -0,0 +1,246 @@
...
+	Name        string             `json:"name"`
+	Description string             `json:"description"`
+	Precedence  uint64             `json:"precedence"`
+	Enabled     bool               `json:"enabled"`
+	Action      TeamsGatewayAction `json:"action"`
+	Filters     []TeamsFilterType  `json:"filters"`
+	Traffic     string             `json:"traffic"`
+	Identity    string             `json:"identity"`
+	Version     uint64             `json:"version"`
+	Settings    TeamsRuleSettings  `db:"settings" json:"rule_settings,omitempty"`
```
Michael9127, line 84: I dont think you need that db:settings tag

teams_accounts.go
```diff
@@ -0,0 +1,131 @@
...
+	"fmt"
+	"github.com/pkg/errors"
+	"net/http"
+	"time"
+)
+
+type TeamsAccount struct {
+	GatewayID    string `json:"gateway_tag"`   // Internal teams ID
+	ProviderName string `json:"provider_name"` // Auth provider
+	CloudflareID string `json:"id"`            // cloudflare account ID
```
jacobbednarz, line 15: is there a reason the struct field is different to the JSON tag? generally we've used a 1:1 for simplicity and reduce confusion between the API and the SDK counterpart.

teams_accounts.go
```diff
@@ -0,0 +1,131 @@
...
+	FooterText      string `json:"footer_text,omitempty"`
+	HeaderText      string `json:"header_text,omitempty"`
+	LogoPath        string `json:"logo_path,omitempty"`
+	BackgroundColor string `json:"background_color,omitempty"`
+	Name            string `json:"name,omitempty"`
+}
+
+// TeamsAccount returns teams account information with internal and external id
+func (api *API) TeamsAccount(ctx context.Context, accountID string) (TeamsAccount, error) {
+	return api.teamsAccount(ctx, accountID)
```
jacobbednarz, line 69: if we're only offering this at the Account level, we can probably do away with these intermediatry methods. in the past the single endpoint has been used to configure Account/Zone level Access features.

teams_accounts.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// TeamsAccountConfiguration returns teams account configuration
+func (api *API) TeamsAccountConfiguration(ctx context.Context, accountID string) (TeamsConfiguration, error) {
+	return api.teamsAccountConfiguration(ctx, accountID)
+}
+
+// TeamsAccountUpsertConfiguration upserts teams account configuration
+func (api *API) TeamsAccountUpsertConfiguration(ctx context.Context, accountID string, config TeamsConfiguration) (TeamsConfiguration, error) {
```
jacobbednarz, line 78: `s/Upsert/Update`? the term upsert definitely leaves some ambiguity as to what is happening internally. if we only offer `HttpPatch`, perhaps Update is more appropriate here.

teams_accounts.go
jacobbednarz, line 15:
```suggestion
	GatewayTag    string `json:"gateway_tag"`   // Internal teams ID
	ProviderName string `json:"provider_name"` // Auth provider
	ID string `json:"id"`            // cloudflare account ID
```

is there a reason the struct field is different to the JSON tag? generally we've used a 1:1 for simplicity and reduce confusion between the API and the SDK counterpart.

teams_rules.go
```diff
@@ -0,0 +1,246 @@
...
+	// settings for l4(network) level overrides
+	L4Override *TeamsL4OverrideSettings `json:"l4override"`
+
+	// settings for browser isolation actions
+	BISOAdminControls *TeamsBISOAdminControlSettings `json:"biso_admin_controls"`
+}
+
+// TeamsL4OverrideSettings : only used in l4 filter type rule with action set to override
+type TeamsL4OverrideSettings struct {
+	Ip   string `json:"ip,omitempty"`
```
jacobbednarz, line 35:
```suggestion
	IP   string `json:"ip,omitempty"`
```

teams_rules.go
```diff
@@ -0,0 +1,246 @@
...
+	OverrideHost string `json:"override_host"`
+
+	// settings for l4(network) level overrides
+	L4Override *TeamsL4OverrideSettings `json:"l4override"`
+
+	// settings for browser isolation actions
+	BISOAdminControls *TeamsBISOAdminControlSettings `json:"biso_admin_controls"`
+}
+
+// TeamsL4OverrideSettings : only used in l4 filter type rule with action set to override
```
jacobbednarz, line 33:
```suggestion
// TeamsL4OverrideSettings used in l4 filter type rule with action set to override.
```

teams_rules.go
```diff
@@ -0,0 +1,246 @@
...
+	Precedence  uint64             `json:"precedence"`
+	Enabled     bool               `json:"enabled"`
+	Action      TeamsGatewayAction `json:"action"`
+	Settings    TeamsRuleSettings  `json:"rule_settings,omitempty"`
+}
+
+// TeamsRules returns all rules within an account.
+//
+// API reference: https://api.cloudflare.com/#teams-rules-properties
+func (api *API) TeamsRules(ctx context.Context, accountID string) ([]TeamsRule, error) {
```
jacobbednarz, line 113: simialr to https://github.com/cloudflare/cloudflare-go/pull/693/files#r689148220, if this is only at the Account level, we don't really need this intermediate methods.

teams_accounts.go
rexscaria: Yeah. I changed the variable names to give a little more readability to the user. The original struct json tags we use in the API are confusing.

teams_rules.go
rexscaria: resolved

teams_rules.go
rexscaria: removed all intermediates

teams_accounts.go
jacobbednarz: i agree it can be confusing here, but i think we need to go one way or another as right now it's in limbo and hard to grok if you don't know the reasoning. i think we should rename the fields to match the JSON to avoid further confusion especially seeing how the API docs reference the JSON tags.

teams_rules.go
```diff
@@ -0,0 +1,222 @@
...
+
+// TeamsRulePatchRequest is used to patch an existing rule
+type TeamsRulePatchRequest struct {
+	UUID        string             `json:"id"`
+	Name        string             `json:"name"`
+	Description string             `json:"description"`
+	Precedence  uint64             `json:"precedence"`
+	Enabled     bool               `json:"enabled"`
+	Action      TeamsGatewayAction `json:"action"`
+	Settings    TeamsRuleSettings  `json:"rule_settings,omitempty"`
```
jacobbednarz, line 107:
```suggestion
	ID        string             `json:"id"`
	Name        string             `json:"name"`
	Description string             `json:"description"`
	Precedence  uint64             `json:"precedence"`
	Enabled     bool               `json:"enabled"`
	Action      TeamsGatewayAction `json:"action"`
	RuleSettings    TeamsRuleSettings  `json:"rule_settings,omitempty"`
```

these should also match the JSON tag names

teams_rules.go
```diff
@@ -0,0 +1,222 @@
...
+	Name        string             `json:"name"`
+	Description string             `json:"description"`
+	Precedence  uint64             `json:"precedence"`
+	Enabled     bool               `json:"enabled"`
+	Action      TeamsGatewayAction `json:"action"`
+	Filters     []TeamsFilterType  `json:"filters"`
+	Traffic     string             `json:"traffic"`
+	Identity    string             `json:"identity"`
+	Version     uint64             `json:"version"`
+	Settings    TeamsRuleSettings  `json:"rule_settings,omitempty"`
```
jacobbednarz, line 84:
```suggestion
	RuleSettings    TeamsRuleSettings  `json:"rule_settings,omitempty"`
```

teams_rules.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	return teamsRulesResponse.Result, nil
+}
+
+// TeamsRule returns the rule with rule id in the url (uuid without dashes)
+//
+// API reference: https://api.cloudflare.com/#teams-rules-properties
+func (api *API) TeamsRule(ctx context.Context, accountID string, ruleId string, ) (TeamsRule, error) {
+	uri := fmt.Sprintf("/accounts/%s/gateway/rules/%s", accountID, ruleId)
```
jacobbednarz, line 134: looking between the API and this PR, we've got a difference in how the rule ID is stored and retrieved.

in the API (the RESTful route), we want a hash style value, however on the resource itself it is a UUIDv4 value. this is very confusing and will make handling this in Terraform/cf-terraforming very painful as the identifier is used as an anchor for subsequent API calls. ideally, we decide on one and continue with it to avoid extra work going forward.

rexscaria: Turns we have new change in api, that accepts uuid with dashes. I have updated the comment. Resolving this.

rexscaria:
I have made all the json tags match variable names.

jacobbednarz:
thanks @rexscaria, this is awesome! when the public documentation links are available we can get those updated.

PR #697, opened 2021-08-20
GATE-1848: Updates omitempty fields to allow disabling account features

Michael9127:
## Description
Removes omitempty tag from boolean fields in teams account config. This allows disabling account features.

## Types of changes

## Checklist:


jacobbednarz:
I think you probably want these to be pointers to explicitly control the value but omit it when not provided. Otherwise, the endpoint will always get a value even if the end user doesn't specify one.

Michael9127:
The payloads should always be called with all the fields present, but ill update just in case im wrong.

cotaboss (commented):
Fix please

Issue #619, opened 2021-04-11
Make all List* methods automatically paginate results


jacobbednarz:
### Current cloudflare-go version

0.15.0

### Description

When fetching resources, all resources should be returned automatically and the end user shouldn't need to explicitly do it.

cdloh:
@jacobbednarz are you happy to accept PRs for this? I'm looking for this for the CustomHostnames endpoint.

jacobbednarz:
Sure thing. See the zones endpoint for example.

cdloh:
Gotcha. Will give it a shot @jacobbednarz !

Some of the endpoints use for loops (DNSRecords for eg) over sync.WaitGroup. Do you want to use sync.WaitGroup for any going forward?

jacobbednarz:
Yeah, I think that's fine. Unless we have good reason not to?

favonia:
I want to point out that the current `ListZones` and `ListZonesContext` might be flawed; see my pull request #674.

jacobbednarz:
this is no longer required as we will be taking a different approach with the experimental client (for v2) with different, yet consistent interfaces.

see for initial musings https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md

PR #709, opened 2021-09-17
WDAPI-532 - adding warp split tunnels

kjhandy:
Adding warp split tunnel functionality.

Tested with Terraform provider.

## Types of changes

## Checklist:

### DIFF
--- a/split_tunnel.go
+++ b/split_tunnel.go
@@ -9,7 +9,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-// SplitTunnelIncludeGetResponse represents the response from the get split
+// SplitTunnelGetResponse represents the response from the get split
 // tunnel endpoints.
 type SplitTunnelGetResponse struct {
 	Response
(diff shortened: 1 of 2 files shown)


aplunk (approved):

split_tunnel.go
```diff
@@ -0,0 +1,105 @@
...
+	Address     string `json:"address,omitempty"`
+	Host        string `json:"host,omitempty"`
+	Description string `json:"description,omitempty"`
+}
+
+
+// SplitTunnelInclude returns all split tunnel include rules within an account.
+//
+// API reference: https://api.cloudflare.com/#device-policy-get-split-tunnel-include-list
+func (api *API) SplitTunnelInclude(ctx context.Context, accountID string) ([]SplitTunnel, error) {
```
jacobbednarz, line 30: I was having a look at the Terraform implementation and wondered if we would be better off passing in the `include`/`exclude` into this method as a parameter and constructing the URL using the value (similar to what we do with the other parameters already)?

We should also prefix this method with `List` seeing how we are pulling them all back and it will line up with #619 when we get to it.

split_tunnel_test.go
```diff
@@ -0,0 +1,256 @@
...
+			"success": true,
+			"errors": [],
+			"messages": [],
+			"result": [
+				{
+					"address": "10.0.0.0/8",
+					"description": "default address"
+			 },
+			 {
+				  "address": "100.64.0.0/10",
```
jacobbednarz, line 99: can we update this to use the IPs from [rfc5737](https://datatracker.ietf.org/doc/html/rfc5737#section-3)?

jacobbednarz:
thanks for raising this one @kjhandy! looks like a solid start 👏 i've included a few questions inline for when you have a chance.

split_tunnel.go
kjhandy: That makes sense. I was originally thinking that, but then thought if they ever diverged that it would be better to have two separate implementations, but we can cross that bridge if we get to it. I also combined the update functions since those are the exact same as well and can just be constructed using an `include/exclude` parameter. Let me know if that was a bad idea.

jacobbednarz:
thanks for this @kjhandy, you rock 🤘

PR #713, opened 2021-09-22
Boolean can't have omitempty or it can never be false

kjhandy:
Need to strip `omitempty` from these `bool` fields or they will never be returned as `false`.

## Types of changes

## Checklist:

### DIFF
--- a/device_posture_rule.go
+++ b/device_posture_rule.go
@@ -29,12 +29,12 @@ type DevicePostureRuleMatch struct {
 type DevicePostureRuleInput struct {
 	ID         string `json:"id,omitempty"`
 	Path       string `json:"path,omitempty"`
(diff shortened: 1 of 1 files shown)


jacobbednarz:
Another option is to make these fields a `*bool` and only explicitly set them when we want them. Here are some examples of it in action in cloudflare-go

https://github.com/cloudflare/cloudflare-go/blob/b8935550e028d7c46a7adb20ab8c48f8e849bd06/dns.go#L23
https://github.com/cloudflare/cloudflare-go/blob/b8935550e028d7c46a7adb20ab8c48f8e849bd06/dns_test.go#L77

jacobbednarz:
closing to becoming stale. let me know if we still end up needing this.

PR #735, opened 2021-11-02
AS-7353: Add cloudflare-go integration for alert history API

priyakvs:
Adding cloudflare-go integration for alert history API to notifications.go

### DIFF
--- a/notifications.go
+++ b/notifications.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/url"
+	"strconv"
 	"time"
 )
 
@@ -146,6 +148,27 @@ type NotificationAlertWithDescription struct {
 type NotificationAvailableAlertsResponse struct {
 	Response
 	Result NotificationsGroupedByProduct
+}
+
+// NotificationHistory describes the history
+// of notifications sent for an account.
+type NotificationHistory struct {
+	ID            string    `json:"id"`
+	Name          string    `json:"name"`
+	Description   string    `json:"description"`
+	AlertBody     string    `json:"alert_body"`
+	AlertType     string    `json:"alert_type"`
+	Mechanism     string    `json:"mechanism"`
+	MechanismType string    `json:"mechanism_type"`
+	Sent          time.Time `json:"sent"`
+}
+
+// NotificationHistoryResponse describes the notification history
+// response for an account for a specific time period.
+type NotificationHistoryResponse struct {
+	Response
+	ResultInfo `json:"result_info"`
+	Result     []NotificationHistory
 }
 
 // ListNotificationPolicies will return the notification policies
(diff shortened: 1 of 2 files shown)


notifications.go
```diff
@@ -148,6 +148,27 @@ type NotificationAvailableAlertsResponse struct {
 	Result NotificationsGroupedByProduct
 }
 
+// NotificationHistory describes the history
+// of notifications sent for an account.
+type NotificationHistory struct {
+	Id            string    `json:"id"`
```
jacobbednarz, line 154:
```suggestion
	ID            string    `json:"id"`
```

notifications.go
```diff
@@ -148,6 +148,27 @@ type NotificationAvailableAlertsResponse struct {
...
+	Mechanism     string    `json:"mechanism"`
+	MechanismType string    `json:"mechanism_type"`
+	Sent          time.Time `json:"sent"`
+}
+
+// NotificationHistoryResponse describes the notification history
+// response for an account for a specific time period.
+type NotificationHistoryResponse struct {
+	Response
+	ResultInfo
```
jacobbednarz, line 168: it doesn't look like we're using `ResultInfo` anywhere; did you intend to?

jacobbednarz (commented):
couple of inline comments but looks great. would you mind also adding a test for this new endpoint please?

notifications.go
priyakvs: My understanding is if the API returns pagination, it is marshalled to this as well. Please correct me if I am wrong.

jacobbednarz: yep, that's all good. my question was more pointed at the fact the previous return type was only ever returning the `Result` and `error` 🙂 what do you think of having this method return 3 parameters and the result info is in it's own parameter as opposed to putting it all together? The SDK is a little bit of a mishmash of approaches but this does allow people to disregard the result info if they don't want to explicitly paginate.

priyakvs: I see what you are saying. I have now changed it to return three parameters separately and also have included an pagination argument to paginate if the users want to.

notifications.go
```diff
@@ -384,6 +407,38 @@ func (api *API) GetAvailableNotificationTypes(ctx context.Context, accountID str
 	return r, nil
 }
 
+// ListNotificationHistory will return the history of alerts sent for
+// a given account. The time period varies based on zone plan.
+// Free, Biz, Pro = 30 days
+// Ent = 90 days
+//
+// API Reference: https://api.cloudflare.com/#notification-history-list-history
+func (api *API) ListNotificationHistory(ctx context.Context, pageOpts PaginationOptions, accountID string) ([]NotificationHistory, ResultInfo, error) {
```
jacobbednarz, line 416:
```suggestion
func (api *API) ListNotificationHistory(ctx context.Context, accountID string, pageOpts PaginationOptions) ([]NotificationHistory, ResultInfo, error) {
```

notifications_test.go
```diff
@@ -504,3 +504,65 @@ func TestDeleteNotificationWebhooks(t *testing.T) {
...
+			string(resultInfo),
+			string(alertHistory))
+		if err != nil {
+			return
+		}
+	}
+
+	mux.HandleFunc("/accounts/"+testAccountID+"/alerting/v3/history", handler)
+
+	actualResult, actualResultInfo, err := client.ListNotificationHistory(context.Background(), pageOptions, testAccountID)
```
jacobbednarz, line 563:
```suggestion
	actualResult, actualResultInfo, err := client.ListNotificationHistory(context.Background(), testAccountID, pageOptions)
```

jacobbednarz:
thank you @priyakvs, this is awesome 🍭

PR #753, opened 2021-12-13
Add support for CLOUDFLARE_API_HOSTNAME

manatarms:
Allows for talking to hostnames other than api.cloudflare.com

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -19,7 +19,7 @@ import (
 	"golang.org/x/time/rate"
 )
 
-const apiURL = "https://api.cloudflare.com/client/v4"
+var apiURL = GetAPIUrl()
 
 const (
 	originCARootCertEccURL = "https://developers.cloudflare.com/ssl/0d2cd0f374da0fb6dbf53128b60bbbf7/origin_ca_ecc_root.pem"
--- a/util.go
+++ b/util.go
@@ -1,0 +1,12 @@
+package cloudflare
+
+import "os"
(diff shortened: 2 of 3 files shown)


jacobbednarz:
👋👋👋👋

within the library, we already have this configuration option, unfortunately named `BaseURL` - https://github.com/cloudflare/cloudflare-go/blob/8f84ec39c0f8f61930232fc2e227b9fcd70ed5d4/options.go#L87

here is an example of it in action within the Terraform provider - https://github.com/cloudflare/terraform-provider-cloudflare/blob/f948788352200cba83d07e107e775b3645d68416/cloudflare/provider.go#L201

both of these are available in released versions too so that should cover everything. 

hope this helps!

jacobbednarz:
the library example can be seen in the initial PR (#606) too as the options can be a little confusing to use.

manatarms:
I'm mostly doing this so that cf-terraforming uses the same hostname ENV as terraform. I can probably do this in cf-terraforming itself.
I'll add it to my current PR https://github.com/cloudflare/cf-terraforming/pull/331

jacobbednarz:
that’s probably the best way. the other two places (Go library and Terraform) have support for this so cf-terraforming just needs it.

manatarms:
Updated https://github.com/cloudflare/cf-terraforming/pull/332/commits/70f5f1f075672fa9cd921e4ea405d601579a94f1

PR #764, opened 2021-12-30
Cloudflare Images

matthewpi:
## Description

Adds support for [Cloudflare Images](https://developers.cloudflare.com/images/cloudflare-images)

https://api.cloudflare.com/#cloudflare-images-properties

- Unit tests
- Against production API, uploading over 40,000 images.

## Types of changes

## Checklist:

### DIFF
--- a/images.go
+++ b/images.go
@@ -75,10 +75,27 @@ func (b ImageUploadRequest) write(mpw *multipart.Writer) error {
 	return nil
 }
 
-// ImageUploadResponse is the API response for an image upload.
-type ImageUploadResponse struct {
-	Result Image `json:"result"`
+// ImageUpdateRequest is the data required for an UpdateImage request.
+type ImageUpdateRequest struct {
+	RequireSignedURLs bool                   `json:"requireSignedURLs"`
+	Metadata          map[string]interface{} `json:"metadata,omitempty"`
+}
+
+// ImageDirectUploadURLRequest is the data required for a CreateImageDirectUploadURL request.
+type ImageDirectUploadURLRequest struct {
+	Expiry time.Time `json:"expiry"`
+}
+
+// ImageDirectUploadURLResponse is the API response for a direct image upload url.
+type ImageDirectUploadURLResponse struct {
+	Result ImageDirectUploadURL `json:"result"`
 	Response
+}
+
+// ImageDirectUploadURL .
+type ImageDirectUploadURL struct {
+	ID        string `json:"id"`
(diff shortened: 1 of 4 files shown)


images.go
```diff
@@ -0,0 +1,190 @@
...
+	if err != nil {
+		return Image{}, errors.Wrap(err, errUnmarshalError)
+	}
+	return imageUploadResponse.Result, nil
+}
+
+// Images lists all images.
+//
+// API Reference: https://api.cloudflare.com/#cloudflare-images-list-images
+func (api *API) Images(ctx context.Context, accountID string, pageOpts PaginationOptions) ([]Image, error) {
```
jacobbednarz, line 146: we should prefix this with `List` as we are fetching all resources.

```suggestion
func (api *API) ListImages(ctx context.Context, accountID string, pageOpts PaginationOptions) ([]Image, error) {
```

jacobbednarz:
> Endpoints are publicly documented, but I am unsure if they are considered stable.

providing they are documented publicly, we can consider the latter part true :)

matthewpi:
> > Endpoints are publicly documented, but I am unsure if they are considered stable.
> 
> providing they are documented publicly, we can consider the latter part true :)

I mentioned that because I was unsure.  I wrote this code back when Images were still in beta and the API documentation at the time had some incorrect responses, but everything seems to be good now.

After quickly looking at the documentation I found I missed some endpoints, so I'm going to see about quickly adding those in as well.

images_test.go
```diff
@@ -0,0 +1,395 @@
...
+	Variants: []string{
+		"https://imagedelivery.net/MTt4OTd0b0w5aj/ZxR0pLaXRldlBtaFhhO2FiZGVnaA/hero",
+		"https://imagedelivery.net/MTt4OTd0b0w5aj/ZxR0pLaXRldlBtaFhhO2FiZGVnaA/original",
+		"https://imagedelivery.net/MTt4OTd0b0w5aj/ZxR0pLaXRldlBtaFhhO2FiZGVnaA/thumbnail",
+	},
+	Uploaded: timeMustParse(time.RFC3339, "2014-01-02T02:20:00Z"),
+}
+
+func TestUploadImage(t *testing.T) {
+	setup(UsingAccount("foo"))
```
jacobbednarz, line 42: let's use the `testAccountID` value here instead of "foo".

```suggestion
	setup(UsingAccount(testAccountID))
```

jacobbednarz:
thank you @matthewpi, this is awesome!

PR #766, opened 2022-01-02
add 'omitempty' to APIToken 'name' and 'policies' fields

kevinschoonover:
## Description
The PR adds 'omitempty' to name and policies so that when the APIToken is marshaled during UpdateToken it only includes the provided fields to prevent errors when not specifying name and policies.

Lets say you wanted to update the expiration date of a API Token like so:
```go
  // Adding a small buffer since the TTL will be calculated again after this
  // call to ensure the credential do not expire before the lease
  expirationDate := time.Now().UTC().Truncate(time.Second)
  if ttl > 0 {
      expirationDate = expirationDate.Add(ttl).Add(time.Minute * 1)
  }
  updatedAPIToken := cloudflare.APIToken{ExpiresOn: &expirationDate}
  
  _, err = c.UpdateAPIToken(ctx, id.(string), updatedAPIToken)
  if err != nil {
      return logical.ErrorResponse("failed to update token with new expiration date. err: %s", err), nil
  }
```

when you run this, you'll get the following error
```
HTTP status 400: 
- These rules must pass for { \"name\": \"\", \"expires_on\": \"2022-01-02T05:08:49Z\", \"policies\": null } 
- name must have a length between 1 and 120\n  - policies must be of the type array (400)"}, Redirect:"", Warnings:[]string(nil)
```

because when it is serialized to JSON it isn't omitting the empty values for 'name' and 'policies' according to the json tags in the struct
https://github.com/cloudflare/cloudflare-go/blob/bd861b7ac39b0bb86abc5f9b367a42897d290866/api_token.go#L14-L25

according to the [documentation](https://api.cloudflare.com/#user-api-tokens-update-token) all fields are optional so this should be a safe change from the update side. I'm not sure if this has any widespread effects.

This has been manually tested. Happy to add a unit test if it would be helpful.

## Types of changes

## Checklist:

### DIFF
--- a/api_token.go
+++ b/api_token.go
@@ -13,13 +13,13 @@ import (
 // APIToken is the full API token.
 type APIToken struct {
 	ID         string             `json:"id,omitempty"`
-	Name       string             `json:"name"`
+	Name       string             `json:"name,omitempty"`
 	Status     string             `json:"status,omitempty"`
 	IssuedOn   *time.Time         `json:"issued_on,omitempty"`
 	ModifiedOn *time.Time         `json:"modified_on,omitempty"`
 	NotBefore  *time.Time         `json:"not_before,omitempty"`
 	ExpiresOn  *time.Time         `json:"expires_on,omitempty"`
-	Policies   []APITokenPolicies `json:"policies"`
+	Policies   []APITokenPolicies `json:"policies,omitempty"`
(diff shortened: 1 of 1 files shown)


jacobbednarz:
thank you @kevinschoonover 👏

PR #770, opened 2022-01-08
Propagate Context in Workers KV methods

maxwellpeterson:
Update the `WriteWorkersKV`, `WriteWorkersKVBulk`, and `DeleteWorkersKVBulk` methods to pass the provided Context to the helper method that makes the HTTP request. Before this change, the Context provided as an argument to these methods was ignored, and a new, empty context was created and used for the HTTP request.

## Description

Replace calls to `makeRequestWithHeaders` with calls to `makeRequestContextWithHeaders` and include the provided Context. Before this change, `makeRequestWithHeaders` would create a new, empty context that would be used for the HTTP request. This is unexpected because all other methods for Workers KV use the provided Context for the HTTP request (by common convention), and in these three cases the caller is unable to control the HTTP request using the provided Context.

Yes, I wrote some unit tests to verify the changes, but did not include them in the PR since this behavior is not tested in other methods that use Context. In these tests, I created a Cloudflare API client that used an HTTP client that used a transport implementation that recorded the Context of the outbound request, and checked that a customized Context provided to these methods made it through to the actual HTTP request.

## Types of changes

## Checklist:

### DIFF
--- a/workers_kv.go
+++ b/workers_kv.go
@@ -171,8 +171,8 @@ func (api *API) UpdateWorkersKVNamespace(ctx context.Context, namespaceID string
 func (api *API) WriteWorkersKV(ctx context.Context, namespaceID, key string, value []byte) (Response, error) {
 	key = url.PathEscape(key)
 	uri := fmt.Sprintf("/accounts/%s/storage/kv/namespaces/%s/values/%s", api.AccountID, namespaceID, key)
-	res, err := api.makeRequestWithHeaders(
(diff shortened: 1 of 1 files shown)


jacobbednarz:
thanks for this @maxwellpeterson! i actually just added a lint rule for this in #772 and inadvertently fixed this (along with a couple of others) so i'll close this one in favour of that to keep the changes together but thanks for the effort here!

PR #773, opened 2022-01-10
Replace hard coded per_page parameter on DNS API with default

Fattouche:
## Description
The API has increased its limit to 5000 therefore its easier for both client
and server to grab as many records as possible per query. This may change in the future
and therefore we should just use default.

https://api.cloudflare.com/#dns-records-for-a-zone-list-dns-records

Yes, we queried the API with this variable removed and verified that it indeed used the default value.


janik-cloudflare (approved):

Chreo (approved):

jacobbednarz (approved):
seems reasonable 👍 we should probably add pagination options to this method at some point too.

PR #786, opened 2022-01-20
GATE-2386: Adds support for missing team rules settings

jgpaiva:
## Description

This PR adds support for creating rules with session management
settings (check_session), adding HTTP headers (add_headers), and other
missing browser isolation options.

Automated tests + manual tests against a live account.

## Types of changes

## Checklist:

### DIFF
--- a/teams_rules.go
+++ b/teams_rules.go
@@ -49,6 +49,7 @@ type TeamsBISOAdminControlSettings struct {
 	DisableUpload    bool `json:"du"`
 	DisableKeyboard  bool `json:"dk"`
 }
+
 type TeamsCheckSessionSettings struct {
 	Enforce  bool     `json:"enforce"`
 	Duration Duration `json:"duration"`
--- a/teams_rules.go
+++ b/teams_rules.go
@@ -26,8 +26,14 @@ type TeamsRuleSettings struct {
 	// settings for l4(network) level overrides
 	L4Override *TeamsL4OverrideSettings `json:"l4override"`
 
+	// settings for adding headers to http requests
+	AddHeaders http.Header `json:"add_headers"`
+
 	// settings for browser isolation actions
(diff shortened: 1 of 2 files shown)


teams_rules.go
```diff
@@ -39,6 +45,13 @@ type TeamsL4OverrideSettings struct {
 type TeamsBISOAdminControlSettings struct {
 	DisablePrinting  bool `json:"dp"`
 	DisableCopyPaste bool `json:"dcp"`
+	DisableDownload  bool `json:"dd"`
+	DisableUpload    bool `json:"du"`
+	DisableKeyboard  bool `json:"dk"`
+}
+type TeamsCheckSessionSettings struct {
```
jacobbednarz, line 52:
```suggestion

type TeamsCheckSessionSettings struct {
```

teams_rules_test.go
```diff
@@ -224,6 +246,23 @@ func TestTeamsCreateRule(t *testing.T) {
 
 	handler := func(w http.ResponseWriter, r *http.Request) {
 		assert.Equal(t, http.MethodPost, r.Method, "Expected method 'POST', got %s", r.Method)
+		type PartialTeamsCheckSessionSettings struct {
```
jacobbednarz, line 249: why are we recreating structs here? should you be using the ones you created above?

jgpaiva: I didn't want to reuse the ones above because the main thing I wanted to test was the `duration` is being correctly serialized into that specific string. If I reused the structs above, I would only be testing that we can deserialize what we serialize, not that it gets serialized to this format.

jgpaiva: Maybe I'm just being too conservative, as we already testing the deserialization from `15m0s` into `Duration` is correct, and there's no reason the serialization wouldn't be also. If you think so too, I don't mind removing this block of serialization validations I added on the current PR.

jacobbednarz: that seems fine to me 👍

PR #801, opened 2022-02-06
Add support for custom environment variables when creating Pages Project

rhaidiz:
## Description
While working with pages project I noticed that the go implementation forces you to have only one environment variable when creating a new pages project, specifically only "BUILD_VERSION".

However, the API accepts any arbitrary variables as long as the format is:
```
{ 
  "VAR_NAME": {
     "value" : "THE_VALUE"
   }
} 
```

At first I thought this might have been an undocumented behavior, however the  [API documentation](https://api.cloudflare.com/#pages-project-create-project) seems generally off too. It seems to suggest that all fields for creating a pages project are optional when they are not, it seems to suggest that there are a bunch of "read only" fields, but they are not all "read only", and finally from the dashboard UI it is possible to create environment variables when creating a new pages project, so I guess this is not an undocumented behavior after all.

I have tested it using the tests already included for pages project and extended them to cover the case of custom environment variables.

## Types of changes

## Checklist:

### DIFF
--- a/pages_project.go
+++ b/pages_project.go
@@ -58,16 +58,11 @@ type PagesProjectDeploymentConfigs struct {
 
 // PagesProjectDeploymentConfigEnvironment represents the configuration for preview or production deploys.
 type PagesProjectDeploymentConfigEnvironment struct {
-	EnvVars PagesProjectDeploymentConfigEnvVars `json:"env_vars"`
-}
-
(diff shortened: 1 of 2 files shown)


jacobbednarz:
thank you @rhaidiz; this is awesome work!

PR #805, opened 2022-02-09
Add functions to manage cache settings (i.e. cache variants)

maleblond:
## Description

The [Zone Cache Settings API endpoints](https://api.cloudflare.com/#zone-cache-settings-properties) have been added recently by Cloudflare. We'd like to be able to configure cache variants via terraform, but the cloudflare terraform provider doesn't support that yet. The first step to add this feature to the terraform provider is to allow managing cache settings through `cloudflare-go`.

I followed the same approach that's been taken to read and update zone settings (`ZoneSingleSetting` & `UpdateZoneSingleSetting`). I chose not to go with a cache variant specific approach (i.e. `UpdateZoneCacheVariant`) to follow the same pattern (i.e. other cache settings could be added in the future, and it would already be supported)

Added unit tests. Also used the following script to test manually:

```
package main

import (
	"context"
	"fmt"
	"log"
	"os"

	"github.com/cloudflare/cloudflare-go"
)

func main() {
	// Construct a new API object using a global API key
	// api, err := cloudflare.New(os.Getenv("CLOUDFLARE_API_KEY"), os.Getenv("CLOUDFLARE_API_EMAIL"))
	// alternatively, you can use a scoped API token
	api, err := cloudflare.NewWithAPIToken(os.Getenv("CLOUDFLARE_API_KEY"))
	if err != nil {
		log.Print(err)
	}

	// Most API calls require a Context
	ctx := context.Background()

	variants, err := api.ZoneCacheVariants(ctx, os.Getenv("CLOUDFLARE_ZONE_ID"))
	if err != nil {
		log.Print(err)
	}
	fmt.Println(variants)

	variants, err = api.UpdateZoneCacheVariants(ctx, os.Getenv("CLOUDFLARE_ZONE_ID"), cloudflare.ZoneCacheVariantsValues{Gif: []string{"image/avif", "image/gif"}})

	if err != nil {
		log.Fatal(err)
	}
	fmt.Println(variants)

	// err = api.DeleteZoneCacheVariants(ctx, os.Getenv("CLOUDFLARE_ZONE_ID"))

	// if err != nil {
	// 	log.Fatal(err)
	// }
}

```

## Types of changes

## Checklist:


zone_cache_settings.go
```diff
@@ -0,0 +1,71 @@
...
+	"net/http"
+
+	"github.com/pkg/errors"
+)
+
+// ZoneSetting contains settings for a zone.
+type ZoneCacheSetting struct {
+	ID         string      `json:"id"`
+	ModifiedOn string      `json:"modified_on,omitempty"`
+	Value      interface{} `json:"value"`
```
jacobbednarz, line 16: this type is too loose i'm afraid. we'll need to match the request/response body a bit closer - https://api.cloudflare.com/#zone-cache-settings-change-variants-setting

we should be able to add each of the image formats to a `[]string`

jacobbednarz (changes requested):
thanks for taking the time to add this 🎉 in order to fully utilise types, we need to be stricter with the `ZoneCacheSetting.Value` to be a specific set of attributes instead of `interface{}` (the catch all Go type). something like this might work

```go
type ZoneCacheSettingValues struct {
	Avif []string `json:"avif"`
	Bmp  []string `json:"bmp"`
	Gif  []string `json:"gif"`
	Jpeg []string `json:"jpeg"`
	Jpg  []string `json:"jpg"`
	Jpg2 []string `json:"jpg2"`
	Jp2  []string `json:"jp2"`
	Png  []string `json:"png"`
	Tiff []string `json:"tiff"`
	Tif  []string `json:"tif"`
	Webp []string `json:"webp"`
}

type ZoneCacheSetting struct {
	ID         string                  `json:"id"`
	ModifiedOn time.Time               `json:"modified_on"`
	Value      ZoneCacheSettingValues  `json:"value"`
}
```

zone_cache_settings.go
maleblond: @jacobbednarz I could do that, and make the new functions specific to cache variants (i.e. `ZoneCacheVariants` instead of `ZoneCacheSettings`), but it would not be consistent with the approach taken with `ZoneSingleSetting` / `UpdateZoneSingleSetting`. I think the upside is that if new cache settings are added in the future, they would get automatically supported. The downside is less clarity around what's expected to be passed as the `Value` here / less type validation.

The cache setting endpoints are documented about the same as the zone setting endpoints, so I thought it'd make sense to use the same approach

maleblond: To be clear, I don't have a strong preference. I'm just trying to get clarity around if you'd like to really expose cache variants as a first-class resource, or consider it as a sub-type of a "Zone Cache Setting".

I guess this would also influence how we should expose this through the terraform provider (i.e. should we have an heavily typed `cloudflare_zone_cache_variants` resource, or rather a `cloudflare_zone_cache_settings`, with a `variants` property embedded in it?)

jacobbednarz: please please please don't use zone settings a good example to work from. it suffers from a few problems (loose types, old endpoints) which make maintenance really difficult and hard to extend. if you're after a good recent, example check out `rulesets.go`.

maleblond: all good, thanks for the clarifications 😄 

I'll add some typings and make the functions more specific to cache variants.

You suggested the following struct:

```
type ZoneCacheSettingValues struct {
	Avif []string `json:"avif"`
	Bmp  []string `json:"bmp"`
	Gif  []string `json:"gif"`
	Jpeg []string `json:"jpeg"`
	Jpg  []string `json:"jpg"`
	Jpg2 []string `json:"jpg2"`
	Jp2  []string `json:"jp2"`
	Png  []string `json:"png"`
	Tiff []string `json:"tiff"`
	Tif  []string `json:"tif"`
	Webp []string `json:"webp"`
}

type ZoneCacheSetting struct {
	ID         string                  `json:"id"`
	ModifiedOn time.Time               `json:"modified_on"`
	Value      ZoneCacheSettingValues  `json:"value"`
}
```

A downside of this approach is that if a file extension is added, then it would also need to be added to both the terraform provider and cloudflare-go. An alternative would be to just use `map[string][]string`. WDYT?

jacobbednarz: yep, that is intentional behaviour. just because a field/value comes back in the API doesn't mean it is yet ready for public consumption. Cloudflare is API driven and until the service teams 👍 the field for public use, we don't want end users depending on it.

maleblond: All good! I pushed a few commits to address the feedback you provided. Ready for a re-review!

zone_cache_variants_test.go
```diff
@@ -0,0 +1,171 @@
...
+				  "image/jpeg"
+				],
+				"bmp": [
+				  "image/webp",
+				  "image/jpeg"
+				]
+			  }
+			}
+		  }`)
+	}
```
jacobbednarz, line 41:
```suggestion
		fmt.Fprintf(w, `
        {
          "success": true,
          "errors": [],
          "messages": [],
          "result": {
            "id": "variants",
            "modified_on": "2014-01-01T05:20:00.12345Z",
            "value": {
              "avif": [
                "image/webp",
                "image/jpeg"
              ],
              "bmp": [
                "image/webp",
                "image/jpeg"
              ]
            }
          }
        }`)
	}
```

zone_cache_variants_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+	testZoneId := "foo"
+	mux.HandleFunc("/zones/"+testZoneId+"/cache/variants", handler)
```
jacobbednarz, line 44:
```suggestion
	mux.HandleFunc("/zones/"+testZoneID+"/cache/variants", handler)
```

zone_cache_variants_test.go
```diff
@@ -0,0 +1,171 @@
...
+				"image/jpeg",
+			},
+			Bmp: []string{
+				"image/webp",
+				"image/jpeg",
+			},
+		},
+	}
+
+	actual, err := client.ZoneCacheVariants(context.Background(), testZoneId)
```
jacobbednarz, line 61:
```suggestion
	actual, err := client.ZoneCacheVariants(context.Background(), testZoneID)
```

zone_cache_variants_test.go
```diff
@@ -0,0 +1,171 @@
...
+				  "image/jpeg"
+				],
+				"bmp": [
+				  "image/webp",
+				  "image/jpeg"
+				]
+			  }
+			}
+		  }`)
+	}
```
jacobbednarz, line 103:
```suggestion
		fmt.Fprintf(w, `
        {
            "success": true,
            "errors": [],
            "messages": [],
            "result": {
                "id": "variants",
                "modified_on": "2014-01-01T05:20:00.12345Z",
                "value": {
                    "avif": [
                        "image/webp",
                        "image/jpeg"
                    ],
                    "bmp": [
                        "image/webp",
                        "image/jpeg"
                    ]
                }
            }
        }`)
	}
```

zone_cache_variants_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+	testZoneId := "foo"
+	mux.HandleFunc("/zones/"+testZoneId+"/cache/variants", handler)
```
jacobbednarz, line 106:
```suggestion
	mux.HandleFunc("/zones/"+testZoneID+"/cache/variants", handler)
```

zone_cache_variants_test.go
```diff
@@ -0,0 +1,171 @@
...
+		Avif: []string{
+			"image/webp",
+			"image/jpeg",
+		},
+		Bmp: []string{
+			"image/webp",
+			"image/jpeg",
+		}}
+
+	actual, err := client.UpdateZoneCacheVariants(context.Background(), testZoneId, zoneCacheVariants)
```
jacobbednarz, line 134:
```suggestion
	actual, err := client.UpdateZoneCacheVariants(context.Background(), testZoneID, zoneCacheVariants)
```

zone_cache_variants_test.go
```diff
@@ -0,0 +1,171 @@
...
+		_, _ = fmt.Fprintf(w, `{
+			"success": true,
+			"errors": [],
+			"messages": [],
+			"result": {
+			  "id": "variants",
+			  "modified_on": "2014-01-01T05:20:00.12345Z"
+			}
+		  }`)
+	}
```
jacobbednarz, line 162:
```suggestion
		fmt.Fprintf(w, `
        {
            "success": true,
            "errors": [],
            "messages": [],
            "result": {
                "id": "variants",
                "modified_on": "2014-01-01T05:20:00.12345Z"
            }
        }`)
	}
```

zone_cache_variants_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+	testZoneId := "foo"
+	mux.HandleFunc("/zones/"+testZoneId+"/cache/variants", handler)
+
+	err := client.DeleteZoneCacheVariants(context.Background(), testZoneId)
+
```
jacobbednarz, line 168:
```suggestion
	mux.HandleFunc("/zones/"+testZoneID+"/cache/variants", handler)

	err := client.DeleteZoneCacheVariants(context.Background(), testZoneID)

```

jacobbednarz (changes requested):
some minor formatting nitpicks inline.

- fixing whitespace in JSON responses to make it readable
- using `testZoneID` as it is a predefined constant instead of `testZoneId` (local var)

jacobbednarz:
i tried to fix this inline for you but it looks like the "allow maintainers to push changes" is disabled here.

jacobbednarz:
ignore my comments; i'll follow up in a PR as i'm planning on cutting a release today and will include this functionality.

PR #814, opened 2022-02-23
fix waiting room event JSON marshalling

broswen:
## Description
The value `disable_session_renewal` should be omitted when `nil` as the falsey value has meaning in this context.
The current code would never allow the event to override the waiting room with a `false` value as it would be omitted from json marshalling.

Yes, the tests have been modified in `waiting_room_test.go`
I changed the struct `waitingRoomEvent` to use `nil` instead of `false` and the json generates as `null`

## Types of changes

(the previous changes that introduce the bug haven't been released yet)
## Checklist:


jacobbednarz:
thank you!

PR #826, opened 2022-03-11
fix issue where the request body is empty when retrying requests

nickysemenza:
## Description 
Previously the body was built from an `io.Reader` once, and then this body was used up to three times. However, the `io.Reader` is empty if the first request successfully reads it, but ultimately fails. This means the subsequent requests have no body (since the reader is fully used up by that point) I updated the retry test to test an action that has a payload, and without this fix the updated test fails.

yes
## Types of changes

## Checklist:


jacobbednarz:
nice find @nickysemenza! thank you!

Issue #748, opened 2021-12-04
support tunnel route api endpoints


tjstansell:
### Current cloudflare-go version

latest

### Description

Please support all of the tunnel route API endpoints as documented at https://api.cloudflare.com/#tunnel-route-properties

* List Tunnel Routes (`GET accounts/:account_identifier/teamnet/routes`)
* Get Tunnel Route by IP (`GET accounts/:account_identifier/teamnet/routes/ip/:ip`)
* Create Route (`POST accounts/:account_identifier/teamnet/routes/network/:ip_network`)
* Update Route (`PATCH accounts/:account_identifier/teamnet/routes/network/:ip_network`)
* Delete Route (`DELETE accounts/:account_identifier/teamnet/routes/network/:ip_network`)

### Use cases

This would provide the support needed to allow the terraform provider to be able to manage tunnel routes.

### Potential cloudflare-go usage

```go
whatever the normal way is to expose api features.
```

### References

https://github.com/cloudflare/terraform-provider-cloudflare/issues/1086

tpickett66:
I started taking a pass at this but couldn't find anywhere that handled optional parameters for list operations. Am I missing something? If no, should I just invent something and open a PR for discussion?

Issue #755, opened 2021-12-14
prefix methods that fetch a single item with `Get`


tjstansell:
### Current cloudflare-go version

0.28.0

### Description

Similar to #612, perhaps all methods should use a standardized naming that has a verb prefix.  Since `List` already has its own issue, I think the remaining methods that do not already use a verb prefix are querying a single resource and should use the `Get` prefix, which is already used for some resources.

### Use cases

Similar to the argument for List, this could help with autocomplete

### Potential cloudflare-go usage

```go
just like now, but prefixed with Get.
```

### References

* #612

tjstansell:
Related to this, is there a deprecation framework for api-incompatible change like this?  These new function names could easily be added and then deprecation warnings on the old names could be introduced to warn folks they'll be removed at a future date.  Just not sure if there's already a process for this kind of thing in place.

jacobbednarz:
an approach like #758 is preferable and like you suggest; add new methods wrapping the old ones and eventually deprecate the old ones after a few releases.

jacobbednarz:
this is no longer required as we will be taking a different approach with the experimental client (for v2) with different, yet consistent interfaces.

see for initial musings https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md

Issue #612, opened 2021-03-31
change methods that fetch all resources to be prefixed with `List`


jacobbednarz:
### Current cloudflare-go version

0.15.0

Right now we have a myriad of methods that range from `ListX` and `Lists` (plural). This is hard to track down available methods for tools like autocomplete. Instead, we can consolidate the prefix to be consistent like the other CRUD operations.

favonia:
As another breaking change, I wonder if it makes sense for these functions to return a modified `ResultInfo` saying that "there's only one single page" (that is, `Page` is always `1` and `Count` matches `Total`, etc,), completely hiding pagination from the library users.

jacobbednarz:
See #615 for that. We're most likely going to remove it entirely. As it's not needed.

favonia:
Got it. I am also happy with complete removal of it from the library interface. :smiley: (Could not find the relevant changes in the cited issues/PRs, though.)

tjstansell:
If anyone is tracking this ticket, it seems the new [experimental](https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md) pieces will make this effort moot, as all of the methods will move down into functions on service objects.

PR #844, opened 2022-03-25
Add listing of tunnel routes without optional parameters.

tpickett66:
I'm opening this to add the tunnel routes API endpoints closing #748 in service of ultimately addressing the same resources in the terraform provider (https://github.com/cloudflare/terraform-provider-cloudflare/issues/1086)

It hasn't... yet.

## Types of changes

## Checklist:

### DIFF
--- a/tunnel_routes.go
+++ b/tunnel_routes.go
@@ -1,0 +1,43 @@
+package cloudflare
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+type TunnelRoute struct {
+	Network    string    `json:"network"`
+	TunnelId   string    `json:"tunnel_id"`
+	TunnelName string    `json:"tunnel_name"`
+	Comment    string    `json:"comment"`
+	CreatedAt  time.Time `json:"created_at"`
+	DeletedAt  time.Time `json:"deleted_at"`
+}
+
+// TunnelRouteListResponse is the API response for listing tunnel routes.
+type TunnelRouteListResponse struct {
+	Response
+	Result []TunnelRoute `json:"result"`
+}
+
+func (api *API) TunnelRoutes(ctx context.Context) ([]TunnelRoute, error) {
+	uri := fmt.Sprintf("/%s/%s/teamnet/routes", AccountRouteRoot, api.AccountID)
+	res, err := api.makeRequestContext(ctx, http.MethodGet, uri, nil)
+
+	if err != nil {
+		return []TunnelRoute{}, err
+	}
+
+	var resp TunnelRouteListResponse
+	err = json.Unmarshal(res, &resp)
+	if err != nil {
+		return []TunnelRoute{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return resp.Result, nil
+}
--- a/tunnel_routes_test.go
+++ b/tunnel_routes_test.go
@@ -1,0 +1,62 @@
+package cloudflare
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+//List Tunnel Routes (GET accounts/:account_identifier/teamnet/routes)
+func TestAPI_ListTunnelRoutes(t *testing.T) {
+	setup(UsingAccount(testAccountID))
+	defer teardown()
+
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, http.MethodGet, r.Method, "Expected method 'GET', got %s", r.Method)
+		w.Header().Set("content-type", "application/json")
+		fmt.Fprintf(w, `{
+			"success": true,
+			"errors": [],
+			"messages": [],
+			"result": [
+			  {
+			    "network": "ff01::/32",
+				"tunnel_id": "f70ff985-a4ef-4643-bbbc-4a0ed4fc8415",
+				"tunnel_name": "blog",
+				"comment": "Example comment for this route",
+				"created_at": "2021-01-25T18:22:34.317854Z",
+				"deleted_at": "2021-01-25T18:22:34.317854Z"
+              }
+            ]
+          }`)
+	}
+
+	mux.HandleFunc("/accounts/"+testAccountID+"/teamnet/routes", handler)
+
+	ts, _ := time.Parse(time.RFC3339Nano, "2021-01-25T18:22:34.317854Z")
+	want := []TunnelRoute{
+		{
+			"ff01::/32",
+			"f70ff985-a4ef-4643-bbbc-4a0ed4fc8415",
+			"blog",
+			"Example comment for this route",
+			ts,
+			ts,
+		},
+	}
+
+	got, err := client.TunnelRoutes(context.Background())
+
+	if assert.NoError(t, err) {
+		assert.Equal(t, want, got)
+	}
+}
+
+//Get Tunnel Route by IP (GET accounts/:account_identifier/teamnet/routes/ip/:ip)
+//Create Route (POST accounts/:account_identifier/teamnet/routes/network/:ip_network)
+//Update Route (PATCH accounts/:account_identifier/teamnet/routes/network/:ip_network)
+//Delete Route (DELETE accounts/:account_identifier/teamnet/routes/network/:ip_network)


tunnel_routes.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+// TunnelRoute is the full record for a route.
+type TunnelRoute struct {
+	Network    string    `json:"network"`
+	TunnelId   string    `json:"tunnel_id"`
+	TunnelName string    `json:"tunnel_name"`
+	Comment    string    `json:"comment"`
+	CreatedAt  time.Time `json:"created_at"`
+	DeletedAt  time.Time `json:"deleted_at"`
```
jacobbednarz, line 21: these `time.Time` references should be pointers

```suggestion
	CreatedAt  *time.Time `json:"created_at"`
	DeletedAt  *time.Time `json:"deleted_at"`
```

tunnel_routes.go
```diff
@@ -0,0 +1,137 @@
...
+		return TunnelRoute{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return routeResponse.Result, nil
+}
+
+// TunnelRoutes lists all defined routes for tunnels in the account.
+//
+// See: https://api.cloudflare.com/#tunnel-route-list-tunnel-routes
+func (api *API) TunnelRoutes(ctx context.Context) ([]TunnelRoute, error) {
```
jacobbednarz, line 52: this method should take into consideration the [`PaginationOptions`](https://github.com/cloudflare/cloudflare-go/blob/365afd256caf70673bc294886b1eeb08cddae7ca/cloudflare.go#L458-L461) for when the number of routes is over the default pager.

```suggestion
func (api *API) TunnelRoutes(ctx context.Context, pageOpts PaginationOptions) ([]TunnelRoute, ResultInfo, error) {
```

you'll also need to factor these in accordingly in the method below.

tunnel_routes.go
```diff
@@ -0,0 +1,137 @@
...
+func (api *API) TunnelRouteForIp(ctx context.Context, ip string) (TunnelRoute, error) {
+	uri := fmt.Sprintf("/%s/%s/teamnet/routes/ip/%s", AccountRouteRoot, api.AccountID, ip)
+	return extractTunnelRouteResponse(api.makeRequestContext(ctx, http.MethodGet, uri, nil))
+}
+
+// CreateTunnelRoute add a new route to the account's routing table for the given tunnel
+//
+// See: https://api.cloudflare.com/#tunnel-route-create-route
+func (api *API) CreateTunnelRoute(ctx context.Context, tunnelId string, ipNetwork string) (TunnelRoute, error) {
+	return api.createTunnelRoute(ctx, tunnelId, ipNetwork, "")
```
jacobbednarz, line 81: we don't really need these internal methods on this resource. other resources use it where there is `account` and `zone` level resources that need management.

tunnel_routes.go
```diff
@@ -0,0 +1,137 @@
...
+
+	return resp.Result, nil
+}
+
+// TunnelRouteForIp finds the Tunnel Route that encompasses the given IP
+//
+// See: https://api.cloudflare.com/#tunnel-route-get-tunnel-route-by-ip
+func (api *API) TunnelRouteForIp(ctx context.Context, ip string) (TunnelRoute, error) {
+	uri := fmt.Sprintf("/%s/%s/teamnet/routes/ip/%s", AccountRouteRoot, api.AccountID, ip)
+	return extractTunnelRouteResponse(api.makeRequestContext(ctx, http.MethodGet, uri, nil))
```
jacobbednarz, line 74: i don't really like this pattern; you're passing in a full HTTP request without checking the return value and then propagating that to the method here. if the API call fails, this is going to break in all sorts of ways. let's just repeat and handle the errors in the method itself.

tunnel_routes.go
```diff
@@ -0,0 +1,137 @@
...
+		return []TunnelRoute{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return resp.Result, nil
+}
+
+// TunnelRouteForIp finds the Tunnel Route that encompasses the given IP
+//
+// See: https://api.cloudflare.com/#tunnel-route-get-tunnel-route-by-ip
+func (api *API) TunnelRouteForIp(ctx context.Context, ip string) (TunnelRoute, error) {
```
jacobbednarz, line 72:
```suggestion
func (api *API) TunnelRouteForIP(ctx context.Context, ip string) (TunnelRoute, error) {
```

tunnel_routes.go
```diff
@@ -0,0 +1,137 @@
...
+	uri := fmt.Sprintf("/%s/%s/teamnet/routes/ip/%s", AccountRouteRoot, api.AccountID, ip)
+	return extractTunnelRouteResponse(api.makeRequestContext(ctx, http.MethodGet, uri, nil))
+}
+
+// CreateTunnelRoute add a new route to the account's routing table for the given tunnel
```
jacobbednarz, line 77:
```suggestion
// CreateTunnelRoute add a new route to the account's routing table for the given tunnel.
```

tunnel_routes.go
```diff
@@ -0,0 +1,137 @@
...
+//
+// See: https://api.cloudflare.com/#tunnel-route-create-route
+func (api *API) CreateTunnelRoute(ctx context.Context, tunnelId string, ipNetwork string) (TunnelRoute, error) {
+	return api.createTunnelRoute(ctx, tunnelId, ipNetwork, "")
+}
+
+// CreateTunnelRouteWithComment add a new route to the account's routing table for the given tunnel
```
jacobbednarz, line 84:
```suggestion
// CreateTunnelRouteWithComment add a new route to the account's routing table for the given tunnel.
```

tunnel_routes.go
```diff
@@ -0,0 +1,137 @@
...
+//
+// See: https://api.cloudflare.com/#tunnel-route-create-route
+func (api *API) CreateTunnelRouteWithComment(ctx context.Context, tunnelId string, ipNetwork string, comment string) (TunnelRoute, error) {
```
jacobbednarz, line 87: why do we have this method? shouldn't the logic just be contained in `CreateTunnelRoute` and set the comment accordingly if provided?

tunnel_routes_test.go
```diff
@@ -0,0 +1,205 @@
...
+	"io/ioutil"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+//List Tunnel Routes (GET accounts/:account_identifier/teamnet/routes)
+func TestAPI_ListTunnelRoutes(t *testing.T) {
```
jacobbednarz, line 16:
```suggestion
func TestListTunnelRoutes(t *testing.T) {
```

tunnel_routes_test.go
```diff
@@ -0,0 +1,205 @@
...
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+//List Tunnel Routes (GET accounts/:account_identifier/teamnet/routes)
```
jacobbednarz, line 15: we don't need these comments here; the methods themselves hold that information

```suggestion
```

tunnel_routes_test.go
```diff
@@ -0,0 +1,205 @@
...
+func TestAPI_ListTunnelRoutes(t *testing.T) {
+	setup(UsingAccount(testAccountID))
```
jacobbednarz, line 17: we don't need the call `UsingAccount` here, we're passing in the account details directly.

```suggestion
	setup()
```

jacobbednarz (changes requested):
i've performed a first pass on this one but it looks like we need a  bit of work to get this into a mergeable state so i'm going to hold off on finishing until we address those other underlying bits.

thanks for what we have so far!

tunnel_routes.go
tpickett66: If you notice the response handling function accepts both the response and and an error, immediately returning if the error is present. I'm happy to change it but I don't think this will fail in the way you're describing.

tpickett66: I've inlined this function everywhere it was used

tunnel_routes_test.go
tpickett66: Not including `UsingAccount` leaves `api.AccountID` as the zero value (empty string) causing all of these tests to fail. Should I not be using `api.AccountID`?

jacobbednarz: no, you shouldn't. instead the method should accept the account ID as a part of the method signature or parameters.

jacobbednarz: there will be a lint rule coming for this soon but for now, i've just removed it.

jacobbednarz:
i've gone ahead and got this one in order to limit the back and forth but hopefully it covers the direction we are moving in for the upcoming [experimental changes](https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md) where we are leveraging more of structs for parameters instead of long method parameter lists.

tpickett66:
The changes you made make sense and look good, I'll check it out in a little bit. My use case here is the terraform provider so I'm happy to make that PR work with this however it needs to.

One question though, why does the `API` struct have an `AccountID` field but it seems like the preference is to pass that value into functions?

jacobbednarz:
the context is largely found in cloudflare/terraform-provider-cloudflare#873 however, the tl;dr is that configuring the account ID at the client level doesn't work for tenants (or anyone with multiple accounts really) and instead, we explicitly pass it into API calls instead.

jacobbednarz:
thanks for the effort here! great to see community contributions!

tunnel_routes.go
```diff
@@ -0,0 +1,186 @@
...
+
+type tunnelRouteResponse struct {
+	Response
+	Result TunnelRoute `json:"result"`
+}
+
+// TunnelRoutes lists all defined routes for tunnels in the account.
+//
+// See: https://api.cloudflare.com/#tunnel-route-list-tunnel-routes
+func (api *API) TunnelRoutes(ctx context.Context, params TunnelRoutesListParams) ([]TunnelRoute, error) {
```
tjstansell, line 73: shouldn't this have been called `ListTunnelRoutes()`?  I thought there was an effort to move all of the functions like this to be prefixed with `List`.

tunnel_routes.go
```diff
@@ -0,0 +1,186 @@
...
+		return []TunnelRoute{}, errors.Wrap(err, errUnmarshalError)
+	}
+
+	return resp.Result, nil
+}
+
+// TunnelRouteForIP finds the Tunnel Route that encompasses the given IP.
+//
+// See: https://api.cloudflare.com/#tunnel-route-get-tunnel-route-by-ip
+func (api *API) TunnelRouteForIP(ctx context.Context, params TunnelRoutesForIPParams) (TunnelRoute, error) {
```
tjstansell, line 97: similarly, shouldn't this have been `GetTunnelRouteForIP()`?

tunnel_routes.go
jacobbednarz: we can however, i'm less concerned about this now that we have some progress on the upcoming [experimental](https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md) client that will do away with everything sitting on the same top level namespace.

we'll be adding that in pre-v2 and then dropping support for the older methods once v2 is released.

tjstansell: I wasn't aware there was a new plan... so nevermind.

tjstansell: I was simply going by #612 and #755, but with the new experimental bits it doesn't really matter, as everything will move to lower level functions.  And if Jacob isn't worried neither am I.

PR #848, opened 2022-03-30
Rename tunnel route get methods to conform to convention

tpickett66:
After #844 was merged @tjstansell raised that these methods were probably misnamed so I'm correcting it now.


tpickett66:
I closed it because I thought it was unnecessary based on the discussion in the other PR: https://github.com/cloudflare/cloudflare-go/pull/844#discussion_r837991532

I'm ok with either really

jacobbednarz:
it's not necessary but if we have the PR, we may as well until the newer client lands.

PR #850, opened 2022-04-04
s/api.AccountID/accountID

### LINKED ISSUE #694: IP List methods should accept an account ID parameter

master

### Description

At the moment, the IP List methods all assume `api.AccountID` is set and remains static (which is the case for `terraform apply` runs). 

https://github.com/cloudflare/cloudflare-go/blob/a6793ca87754dc58afb86cd7a65dd1216b3e786a/ip_list.go#L295

This causes issues with end users trying to execute Terraform operations in parallel (cloudflare/terraform-provider-cloudflare#1140). 

### Use cases

Trying to execute Terraform operations in parallel (cloudflare/terraform-provider-cloudflare#1140) with different account IDs. 

all `*IPList` methods should accept (and reference) `accountID` instead

```diff
- func (api *API) GetIPList(ctx context.Context, id string) (IPList, error) {
+ func (api *API) GetIPList(ctx context.Context, accountID, id string) (IPList, error) {
-	uri := fmt.Sprintf("/accounts/%s/rules/lists/%s", api.AccountID, id)
+	uri := fmt.Sprintf("/accounts/%s/rules/lists/%s", accountID, id)
	res, err := api.makeRequestContext(ctx, http.MethodGet, uri, nil)
	if err != nil {
		return IPList{}, err
	}

	result := IPListResponse{}
	if err := json.Unmarshal(res, &result); err != nil {
		return IPList{}, errors.Wrap(err, errUnmarshalError)
	}

	return result.Result, nil
}
```

### References

_No response_

jacobbednarz:
Updates all methods that were relying on `api.AccountID` via the client
constructor to pass in explicit account IDs for the methods instead.

Even though the Workers APIs still use `api.AccountID`, I'm leaving them
as the whole thing needs an overhaul now that the API has settled.

Closes #694

### DIFF
--- a/ip_address_management.go
+++ b/ip_address_management.go
@@ -62,8 +62,8 @@ type AdvertisementStatusUpdateRequest struct {
 // ListPrefixes lists all IP prefixes for a given account
 //
 // API reference: https://api.cloudflare.com/#ip-address-management-prefixes-list-prefixes
-func (api *API) ListPrefixes(ctx context.Context) ([]IPPrefix, error) {
-	uri := fmt.Sprintf("/accounts/%s/addressing/prefixes", api.AccountID)
+func (api *API) ListPrefixes(ctx context.Context, accountID string) ([]IPPrefix, error) {
+	uri := fmt.Sprintf("/accounts/%s/addressing/prefixes", accountID)
 	res, err := api.makeRequestContext(ctx, http.MethodGet, uri, nil)
 	if err != nil {
 		return []IPPrefix{}, err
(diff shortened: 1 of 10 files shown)


jacobbednarz:
cc @cehrig @tarnfeld

CryptoGodfatherVA4 (approved):

CryptoGodfatherVA4 (approved):

CryptoGodfatherVA4 (approved):

CryptoGodfatherVA4 (approved):

CryptoGodfatherVA4 (approved):

CryptoGodfatherVA4 (approved):

CryptoGodfatherVA4 (approved):

CryptoGodfatherVA4 (approved):

jacobbednarz:
yep, that eventually needs to be removed but i've left it in place as `ZoneIDByName`/`ListZonesContext` will be gone in the next major release since the `List` method will have some search options included as outlined in the [experimental client docs](https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md).

we can change it now but i'm not too worried if it hangs around for now.

PR #868, opened 2022-04-27
work around teams list API bug that only returns 50 items by default

tjstansell:
This patch works around an internal API bug where the teams list API returns only 50 items.  There is an internal ticket with the gateway engineering team to fix this at the API itself, but it has been open for months with no progress.  This is a temporary workaround to [allow the terraform provider](https://github.com/cloudflare/terraform-provider-cloudflare/issues/1422) to actually manage teams lists larger than 50 items.  Given that [lists have a max size of 5000 for enterprise accounts](https://developers.cloudflare.com/cloudflare-one/policies/lists/) we just use the undocumented `limit` parameter to make sure we request the entire list.  This can get reverted once they fix the bug in the API.

### DIFF
--- a/teams_list.go
+++ b/teams_list.go
@@ -107,7 +107,8 @@ func (api *API) TeamsList(ctx context.Context, accountID, listID string) (TeamsL
 //
 // API reference: https://api.cloudflare.com/#teams-lists-teams-list-items
 func (api *API) TeamsListItems(ctx context.Context, accountID, listID string) ([]TeamsListItem, ResultInfo, error) {
-	uri := fmt.Sprintf("/%s/%s/gateway/lists/%s/items", AccountRouteRoot, accountID, listID)
+	// limit parameter works around bug in the API
+	uri := fmt.Sprintf("/%s/%s/gateway/lists/%s/items?limit=10000", AccountRouteRoot, accountID, listID)
 
 	res, err := api.makeRequestContext(ctx, http.MethodGet, uri, nil)
 	if err != nil {


jacobbednarz:
what bug in the API are you referring to? from what i can see, this is just missing pagination options.

tjstansell:
There is no pagination in this API. It returns 50 items without the use of the undocumented limit parameter.

jacobbednarz:
this is odd; if we have a `limit` there is pagination _somewhere_ otherwise changing the value wouldn't have any effect. i don't have a test account setup with this, how is the UI doing pagination?

i've got a ticket internally that has some details on this but not enough to pin point the issues so i'll see whether i can get some 👀 on it to see if we can resolve this.

tjstansell:
The UI uses the undocumented `limit` parameter, which is how I found out about it.  I opened support ticket number 2431495 after having discussed this with the Zero Trust/Teams engineering teams/PMs and not getting any resolution to this after a couple months.  They had verbally agreed it was broken but they haven't fixed it.  The support ticket was updated to say:
> We have an open ticket with the Gateway engineering team on this issue as multiple customers requested this feature.
>
> I will place this ticket on-hold for the time being and send update once we receive further updates from the engineering team.

Given the lack of traction and the fact that [my terraform provider issue](https://github.com/cloudflare/terraform-provider-cloudflare/issues/1422) has seen no traction, I decided to try to implement a workaround here myself.

tjstansell:
Specifically, the Zero Trust UI uses `&limit=1000` to get the full list ... which isn't sufficient, actually, because in our enterprise account the limit is 5000 and we have a list with 1240 items on it.  It shows 10 items per page and if I page forward ... and finally get to the page with item 1000, it shows me an empty broken page.  All of this has been reported to the Zero Trust team internally... nearly 2 months ago... with no fixes.  I'm tired of waiting.

tjstansell:
FWIW, if you can rattle some cages on the API/engineering side instead of having to use a workaround here, great!

jacobbednarz:
i'll see where it is at and whether this workaround is suitable. this won't land if it risks causing performance or availability issues though. 

if you need this immediately, I'd recommend forking this repository with your changes and running that.

tjstansell:
Um... not sure I understand the stance where absolutely broken for lists over 50 items is preferable... but it's your call.

jacobbednarz:
closing this off as the team as already committed to addressing the underlying bug in an upcoming sprint which will allow use of pagination.

PR #869, opened 2022-04-28
Add Origin Override rulesets support

WowVeryLogin:
Adding support for the new phase with rulesets to Override Origin and HostHeader of the request

## Description

Added new structures and calls specifically for Origin Override rulesets and Route Action

Added test

### DIFF
--- a/origin_override_rulesets.go
+++ b/origin_override_rulesets.go
@@ -1,0 +1,199 @@
+package cloudflare
+
+import (
+	"context"
+	"encoding/json"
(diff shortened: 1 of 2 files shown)


jacobbednarz (changes requested):
thanks for the PR, however, this isn't how the Rulesets resource is structured for additions so we won't be accepting it. 

if you's like to update this in line with the conventions for Rulesets, please see the existing Rulesets go file.

PR #878, opened 2022-05-10
add next_event_prequeue_start_time and next_event_start_time 

broswen:
- Add `next_event_prequeue_start_time` and `next_event_start_time` to Waiting Room.

- Use time.Time pointer for nullable Waiting Room Event `prequeue_start_time` property.

## Description
The waiting room schema now exposes the prequeue start time and start time of the next event or the event that is currently in progress.
This is not documented in the official API yet.

The `prequeue_start_time` property on `WaitingRoomEvent`  should be a pointer. It is optional and if not specified should serialize to `null`.
I think switching to a pointer might be a breaking change, especially for the Terraform provider. 

Updated unit tests and pass.
Tested with real resources using the API.

## Types of changes

## Checklist:

### DIFF
--- a/waiting_room.go
+++ b/waiting_room.go
@@ -12,21 +12,23 @@ import (
 
 // WaitingRoom describes a WaitingRoom object.
 type WaitingRoom struct {
-	CreatedOn             time.Time `json:"created_on,omitempty"`
-	ModifiedOn            time.Time `json:"modified_on,omitempty"`
-	Path                  string    `json:"path"`
-	Name                  string    `json:"name"`
-	Description           string    `json:"description,omitempty"`
-	CustomPageHTML        string    `json:"custom_page_html,omitempty"`
-	Host                  string    `json:"host"`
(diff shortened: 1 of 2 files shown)


jacobbednarz:
realistically, we should only be using `*time.Time` (the next major release will likely only use pointers for all fields) so I'm happy to fix that here. the change is internal to Terraform so 👍 from me. 

let's wait for the public documentation though to ensure this endpoint is fully documented before getting too fair ahead of ourselves.

broswen:
> realistically, we should only be using `*time.Time` (the next major release will likely only use pointers for all fields) so I'm happy to fix that here. the change is internal to Terraform so +1 from me.
> 
> let's wait for the public documentation though to ensure this endpoint is fully documented before getting too fair ahead of ourselves.

Public documentation is live. Visible on the Object Definition. https://api.cloudflare.com/#waiting-room-list-waiting-rooms

jacobbednarz:
awesome! thank you @broswen 👏

PR #888, opened 2022-05-18
AS-8430: Update history API to accept timestamp query params

priyakvs:
Alert history API now have the ability to query by timestamp filter. This PR is to add that functionality to cloudflare-go integration as well.

### DIFF
--- a/notifications.go
+++ b/notifications.go
@@ -405,19 +405,37 @@ func (api *API) GetAvailableNotificationTypes(ctx context.Context, accountID str
 	return r, nil
 }
 
+// TimeRange is an object for filtering the alert history based on timestamp.
+type TimeRange struct {
+	Since  string `json:"since,omitempty"`
+	Before string `json:"before,omitempty"`
+}
+
+// AlertHistoryFilter is an object for filtering the alert history response from the api.
+type AlertHistoryFilter struct {
+	TimeRange  TimeRange
+	Pagination PaginationOptions
+}
+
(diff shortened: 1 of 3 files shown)


notifications.go
```diff
@@ -405,19 +405,37 @@ func (api *API) GetAvailableNotificationTypes(ctx context.Context, accountID str
 	return r, nil
 }
 
+// TimeRange is an object for filtering the alert history based on timestamp.
+type TimeRange struct {
+	Since  string `json:"since,omitempty"`
```
jacobbednarz, line 410: something we can do here is apply `url` tags to these fields and instead automatically `Encode` them.

example from `tunnel` - https://github.com/cloudflare/cloudflare-go/blob/fcb3cd41e72cbaeeeb952df36b69d0c4bbff1775/tunnel.go#L105-L113

I might even drop this onto `PaginationOptions` too. WDTY?

jacobbednarz:
i pushed up a change in 19c9713 that swaps to using go-querystring (already in the package) for building and sending the URI parameters to save us manually doing it. do you want to do some quick tests locally with it and confirm it works as you'd expect before we merge this?

jacobbednarz:
all good! i got this working locally using your test and confirmed it on a test account. 

thanks for the PR 🌟

priyakvs:
Thanks a lot for updating this PR @jacobbednarz . Will keep this in mind for all future work :)

PR #896, opened 2022-05-20
support empty custom metadata

sbfaulkner:
## Description

When updating custom metadata, the omitempty prevents go from marshalling an empty map as needed by PATCH semantics for updates (eg. to delete the metadata).

Using a pointer for the map (as is done for the SSL member, and has started being adopted elsewhere in this pkg) resolves this.

Additional tests have been added (and the existing tests cover the normal cases).

## Types of changes

This is sort of a breaking change - any applications using the CustomMetadata field will need to pass an address instead, but since Go is typed they should be aware of the change.

Also, since IIRC this is an enterprise feature, and only after requesting it be activated, I suspect the number of users impacted is minimal. (totally a guess on my part)

## Checklist:


jacobbednarz:
this seems fine to me; @nickysemenza any reservations?

PR #897, opened 2022-05-23
teams_list: add pagination support for items endpoint

jacobbednarz:
unblocks a fix for cloudflare/terraform-provider-cloudflare#1422

### DIFF
--- a/teams_list.go
+++ b/teams_list.go
@@ -7,8 +7,11 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/google/go-querystring/query"
 	"github.com/pkg/errors"
 )
+
+var ErrMissingListID = errors.New("required missing list ID")
 
 // TeamsList represents a Teams List.
 type TeamsList struct {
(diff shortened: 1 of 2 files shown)


jacobbednarz:
public API docs will be released tomorrow.

teams_list.go
```diff
@@ -106,10 +116,24 @@ func (api *API) TeamsList(ctx context.Context, accountID, listID string) (TeamsL
...
+	if params.AccountID == "" {
+		return []TeamsListItem{}, ResultInfo{}, ErrMissingAccountID
+	}
 
-	res, err := api.makeRequestContext(ctx, http.MethodGet, uri, nil)
+	if params.ListID == "" {
+		return []TeamsListItem{}, ResultInfo{}, ErrMissingListID
+	}
+
+	v, _ := query.Values(params)
```
tjstansell, line 128: shouldn't this only be referencing `params.PaginationOptions`?  otherwise, won't it include the `AccountId` and `ListId` fields too and produce incorrect query parameters?

jacobbednarz: no, internally, we use https://github.com/google/go-querystring for query parameter marshalling based on struct field tags. using `url:"-"` allows defining the field but ignoring it for the purposes of building the URL. see https://github.com/google/go-querystring/blob/v1.1.0/query/encode.go#L48-L52 for the full details.

the second parameter to `TeamsListItems` is `TeamsListItemsParams` as per the [experimental client documentation](https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md) where we are making the move to for all methods, eventually.

PR #908, opened 2022-05-26
Updating API usage of IPsec tunnels

lungiman:
Signed-off-by: Amos Paul <amos@cloudflare.com>

## Description

Cloudflare-go is used by Terraform resource provider for Cloudflare. Updating API calls to Magic Transit IPsec Tunnels to support Terraform.

The changes are mostly w.r.t parsing response objects from API calls to cloudflare. This has been tested using unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/magic_transit_ipsec_tunnel.go
+++ b/magic_transit_ipsec_tunnel.go
@@ -16,16 +16,31 @@ const (
 	errMagicTransitIPsecTunnelNotDeleted  = "When trying to delete IPsec tunnel, API returned deleted: false"
 )
 
+type RemoteIdentities struct {
+	HexID  string `json:"hex_id"`
+	FQDNID string `json:"fqdn_id"`
+	UserID string `json:"user_id"`
+}
+
+// MagicTransitIPsecTunnelPskMetadata contains metadata associated with PSK.
+type MagicTransitIPsecTunnelPskMetadata struct {
+	LastGeneratedOn *time.Time `json:"last_generated_on,omitempty"`
+}
+
 // MagicTransitIPsecTunnel contains information about an IPsec tunnel.
 type MagicTransitIPsecTunnel struct {
-	ID                 string     `json:"id,omitempty"`
-	CreatedOn          *time.Time `json:"created_on,omitempty"`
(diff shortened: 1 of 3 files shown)


magic_transit_ipsec_tunnel.go
```diff
@@ -16,70 +16,97 @@ const (
...
+	Name               string                         `json:"name"`
+	CustomerEndpoint   string                         `json:"customer_endpoint"`
+	CloudflareEndpoint string                         `json:"cloudflare_endpoint"`
+	InterfaceAddress   string                         `json:"interface_address"`
+	Description        string                         `json:"description,omitempty"`
+	HealthCheck        *MagicTransitTunnelHealthcheck `json:"health_check,omitempty"`
+}
+
+// MagicTransitIPsecTunnelReq contains information needed to create IPsec tunnel.
+type MagicTransitIPsecTunnelReq struct {
```
jacobbednarz, line 38: is there a reason this (and the response equivalent) cannot be added to `MagicTransitIPsecTunnel` struct with `omitempty` attributes? doing that would avoid a breaking change for method signatures.

jacobbednarz: (may need to be pointers as well, but that is up to the implementation)

lungiman: > is there a reason this (and the response equivalent) cannot be added to `MagicTransitIPsecTunnel` struct with `omitempty` attributes? doing that would avoid a breaking change for method signatures.

Ultimately that is the goal. Currently there are 2 fields sharing the same name. ie, For a request `psk` is a `string` and for response `psk` is the same object as `psk_metadata`. We are in the process of tidying up the responses to be more consistent.

jacobbednarz: do we want to wait for that? i ask because we're changing the return types which require changes from anyone using this library (terraform, cf-terraforming, customers). i'd like to limit the number of disruptive changes if we can.

lungiman: I think waiting will be the cleanest approach.

jacobbednarz:
pending some UI/API changes before we merge this one.

PR #918, opened 2022-05-31
Generalize IP lists to lists of arbitrary kind.

orium:
There are now more list kinds, including `redirect` lists, which we want to support soon.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -26,8 +26,9 @@ const (
 )
 
 var (
-	ErrMissingAccountID = errors.New("required missing account ID")
-	ErrMissingZoneID    = errors.New("required missing zone ID")
+	ErrMissingAccountID          = errors.New("required missing account ID")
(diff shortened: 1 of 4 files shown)


jacobbednarz:
as this is removing a bunch of functionality, we need to first deprecate the old stuff and give folks time to migrate off it before the next major version bump. a simple `Deprecated: Use ...` in the method is suffice for now and we'll add it to the cleanup docs later.

for the new methods, are you able to check out [tunnel_routes](https://github.com/cloudflare/cloudflare-go/blob/master/tunnel_routes.go) and use the struct as parameters approach we've defined in the [experimental docs](https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md) to limit the number of parameters we are passing to methods?

orium:
Thanks @jacobbednarz.  I've updated the PR based on your comments.

jacobbednarz:
thanks! this is great 👏

PR #919, opened 2022-06-01
Add PKCE flag to Identity Provider configuration

GreenStage:
Adds PKCE to Identity Providers Configurations

## Description

Some OAuth Identity providers support PKCE for confidential clients. This improves Cloudflare ZT Access overall strength against some attacks. 

Yes

## Types of changes

## Checklist:

### DIFF
--- a/access_identity_provider.go
+++ b/access_identity_provider.go
@@ -43,6 +43,7 @@ type AccessIdentityProviderConfiguration struct {
 	SsoTargetURL       string   `json:"sso_target_url,omitempty"`
 	SupportGroups      bool     `json:"support_groups,omitempty"`
 	TokenURL           string   `json:"token_url,omitempty"`
+	PKCEEnabled        *bool    `json:"pkce_enabled,omitempty"`
 }
 
 // AccessIdentityProvidersListResponse is the API response for multiple


access_identity_provider.go
```diff
@@ -43,6 +43,7 @@ type AccessIdentityProviderConfiguration struct {
 	SsoTargetURL       string   `json:"sso_target_url,omitempty"`
 	SupportGroups      bool     `json:"support_groups,omitempty"`
 	TokenURL           string   `json:"token_url,omitempty"`
+	PKCEEnabled        bool     `json:"pkce_enabled,omitempty"`
```
jacobbednarz, line 46: is this value always present in the request/responses? do we need a `*bool` here?

jacobbednarz (approved):
LGTM with one question before we merge.

access_identity_provider.go
GreenStage: Nope, but neither are any of the other fields. The server only returns values that are configured. The same applies for pkce_enabled=false

Do you think this will cause issues?

jacobbednarz: as long as "not present" in the payload is an ok state and we default accordingly. we have other APIs at the moment that need `true`/`false`/`uninitialised` and it does cause an issue in Terraform due to the inability to set null values for booleans.

GreenStage: Oh, we should be fine then. There is no need to distinguish false from uninitialised

PR #923, opened 2022-06-08
New API for Managed Headers

WowVeryLogin:
Add support of List and Update methods for Managed Headers Product

## Description

This product exists only on a Zone level. No need for Create of Delete methods: managed headers API always provides a list of possible headers. The only difference is the state of particular header rule in this list, which can be changed by Update method.

Unittests are added

### DIFF
--- a/managed_headers.go
+++ b/managed_headers.go
@@ -1,0 +1,68 @@
+package cloudflare
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/pkg/errors"
+)
+
+type ListManagedHeadersResponse struct {
+	Response
+	Result ManagedHeaders `json:"result"`
+}
+
+type ManagedHeaders struct {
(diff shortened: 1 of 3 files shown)


managed_headers.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+type ManagedHeader struct {
+	ID            string   `json:"id"`
+	Enabled       bool     `json:"enabled"`
+	HasCoflict    bool     `json:"has_conflict,omitempty"`
+	ConflictsWith []string `json:"conflicts_with,omitempty"`
+}
+
+func (api *API) ListZoneManagedHeaders(ctx context.Context, zoneID string) (ManagedHeaders, error) {
```
jacobbednarz, line 29: instead of only accepting zoneID, can we make it accept a struct (with the single param) as a bit of future proofing and consistency with the newer approach?

WowVeryLogin: Makes sense, fixed it

PR #926, opened 2022-06-10
Added support for redirect lists.

orium:


go.mod
```diff
@@ -14,6 +14,7 @@ require (
 )
 
 require (
+	github.com/AlekSi/pointer v1.2.0 // indirect
```
jacobbednarz, line 17: we can get rid of this dependency, we already maintain our own via `convert_types.go`

list_test.go
```diff
@@ -301,14 +302,14 @@ func TestListsItemsIP(t *testing.T) {
 	want := []ListItem{
 		{
 			ID:         "2c0fc9fa937b11eaa1b71c4d701ab86e",
-			IP:         "192.0.2.1",
+			IP:         pointer.ToString("192.0.2.1"),
```
jacobbednarz, line 305: once the dependency is cleaned up, we can use the following

```suggestion
			IP:         cloudflare.StringPtr("192.0.2.1"),
```

jacobbednarz:
thanks a bunch! 🥇

PR #931, opened 2022-06-14
load_balancing: revamp methods for consistent usage

jacobbednarz:
Updates the load balancing methods to:

- Ensure we have a consistent experience for user, zone and account
  level resources.
- A consistent method signature where `ctx` and `params` are available
  to all methods and filtering is performed based on the values passed
  in. See [experimental docs] for details.
- `Modify..` => `Update` to match other method conventions.
- Adds pagination to all `List*` methods via the param structs.

[experimental docs]: https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md

### DIFF
--- a/cloudflare_test.go
+++ b/cloudflare_test.go
@@ -153,7 +153,7 @@ func TestClient_RetryCanSucceedAfterErrors(t *testing.T) {
             "success": true,
             "errors": [],
             "messages": [],
-            "result": 
+            "result":
                 {
                     "id": "17b5962d775c646f3f9725cbc7a53df4",
                     "created_on": "2014-01-01T05:20:00.12345Z",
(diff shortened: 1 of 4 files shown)


gofish:
Pools and monitors are accessed by account, load balancers are accessed by zone.
https://api.cloudflare.com/#account-load-balancer-pools-properties
https://api.cloudflare.com/#account-load-balancer-monitors-properties

These are the only existing (semi-deprecated) `/user` endpoints:
https://api.cloudflare.com/#load-balancer-pools-properties
https://api.cloudflare.com/#load-balancer-monitors-properties

The `/user/load_balancers` endpoint doesn't exist for access to load balancers, only `/zones/%s/load_balancers`
https://api.cloudflare.com/#load-balancers-properties


So neither `...PoolParams` nor `...MonitorParams` should have `ZoneID` fields, `...LoadBalancerParams` should not have `AccountID` fields, and `ZoneID` is required.

PR #937, opened 2022-06-17
fix marshalling of LogpushJob with no filter

sbfaulkner:
## Description

A LogpushJob with no filter incorrectly marshals the filter as `"{\"where\":{}}"` instead of omitting it.

Changing the field to a pointer allows us to test for presence and omit it entirely.

I've added a marshalling test that initially tests the valid and not present cases.

## Types of changes

> **Warning** this is technically a breaking change and will require changes in the terraform provider to use the pointer instead of the value as has been done in other updates.

## Checklist:

### DIFF
--- a/logpush.go
+++ b/logpush.go
@@ -12,18 +12,18 @@ import (
 
 // LogpushJob describes a Logpush job.
 type LogpushJob struct {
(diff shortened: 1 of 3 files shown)


jacobbednarz:
thanks a bunch!

PR #942, opened 2022-06-21
Add filtering to managed headers API

WowVeryLogin:
Add option to filter managed headers API response

Added test


jacobbednarz:
closing as we discussed internally and want to push the filtering to the API and allow the various implementations to control that.

PR #948, opened 2022-06-23
Add UpdateTunnelConfiguration func

EngHabu:
Signed-off-by: Haytham Abuelfutuh <haytham@afutuh.com>

Implement UpdateTunnelConfiguration as described [here](https://api.cloudflare.com/#cloudflare-tunnel-configuration-put-configuration).

## Types of changes

## Checklist:

### DIFF
--- a/tunnel.go
+++ b/tunnel.go
@@ -47,6 +47,13 @@ type TunnelDetailResponse struct {
 	Response
 }
 
+// TunnelConfigurationResponse is used for representing the API response payload for
+// a single tunnel.
+type TunnelConfigurationResponse struct {
+	Result TunnelConfigurationParams `json:"result"`
+	Response
+}
+
 // TunnelTokenResponse is  the API response for a tunnel token.
 type TunnelTokenResponse struct {
 	Result string `json:"result"`
@@ -68,6 +75,13 @@ type TunnelUpdateParams struct {
 	AccountID string `json:"-"`
 	Name      string `json:"name,omitempty"`
 	Secret    string `json:"tunnel_secret,omitempty"`
+}
+
+type TunnelConfigurationParams struct {
+	AccountID string `json:"-"`
+	TunnelID  string `json:"-"`
+	Config    string `json:"config,omitempty"`
+	Version   int    `json:"version,omitempty"`
 }
 
 type TunnelDeleteParams struct {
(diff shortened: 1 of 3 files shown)


tunnel.go
```diff
@@ -211,6 +225,29 @@ func (api *API) UpdateTunnel(ctx context.Context, params TunnelUpdateParams) (Tu
...
 }
 
+// UpdateTunnelConfiguration updates an existing tunnel for the account.
+//
+// API reference: https://api.cloudflare.com/#cloudflare-tunnel-configuration-properties
+func (api *API) UpdateTunnelConfiguration(ctx context.Context, params TunnelConfigurationParams) (Tunnel, error) {
+	if params.AccountID == "" {
+		return Tunnel{}, ErrMissingAccountID
+	}
+
```
jacobbednarz, line 235: you'll also want to check `TunnelID` is provided too

jacobbednarz (changes requested):
thanks! we'll also need to add a unit test for this new endpoint

tunnel.go
```diff
@@ -70,6 +77,87 @@ type TunnelUpdateParams struct {
...
+// OriginRequestConfig is a set of optional fields that users may set to
+// customize how cloudflared sends requests to origin services. It is used to set
+// up general config that apply to all rules, and also, specific per-rule
+// config.
+// Note: To specify a time.Duration in go-yaml, use e.g. "3s" or "24h".
+// This is copied directly from: https://github.com/cloudflare/cloudflared/blob/eed7d7bbc90ccaca129bc6b6c3fd18b19e6bf856/config/configuration.go#L189
+type OriginRequestConfig struct {
+	// HTTP proxy timeout for establishing a new connection
+	ConnectTimeout *time.Duration `yaml:"connectTimeout"`
```
jacobbednarz, line 95: we aren't using YAML anywhere so all these tags can go. best stick to the publicly documented endpoint https://api.cloudflare.com/#cloudflare-tunnel-configuration-put-configuration

tunnel.go
EngHabu: Done

EngHabu:
@jacobbednarz mind taking another look?

jacobbednarz:
i ended up just fixing this myself as there were some foundational issues with this PR that would have made debugging difficult. 

- the structs had incorrect relationships. `TunnelConfiguration` has `OriginRequest` at the top level (not nested under `IngressRules`).
- the test wasn't testing the code path in use, it was only testing that the JSON was parsable which is was but not for the typed struct in use.
- the payload has stringified JSON instead of a typed response. updated both these methods to double unmarshal the payloads (first the v4 envelope then the `Config` after). i'm unsure why this decision was made but i'll look to see if we can change this as it doesn't make sense that we have a structured object stringified in a structured response.

EngHabu:
Thank you @jacobbednarz

PR #958, opened 2022-06-28
Added support for SNI override for the `route` action.

orium:
This adds support for SNI overrides in the `route` action (ruleset's rules).

Marked as a draft PR because the functionality is not yet released.


rulesets.go
```diff
@@ -206,6 +206,7 @@ type RulesetRuleActionParameters struct {
 	Response                *RulesetRuleActionParametersBlockResponse        `json:"response,omitempty"`
 	HostHeader              string                                           `json:"host_header,omitempty"`
 	Origin                  *RulesetRuleActionParametersOrigin               `json:"origin,omitempty"`
+	Sni                     string                                           `json:"sni,omitempty"`
```
jacobbednarz, line 209:
```suggestion
	SNI                     string                                           `json:"sni,omitempty"`
```

jacobbednarz (approved):

PR #962, opened 2022-06-29
Revert DNSRecord.Proxied type into bool and remove `omitempty` tag

zeithrold:
## Description

Here are some reasons that motivated me to do this commitments.

### Previously pointer type causes difficulty to approach to the API

Current version require a _pointer_ to assign boolean value into DNSRecord.Proxied. So to assign a truthy value, instead of `true`, I must use `&[]bool{true}[0]` or any other equal item, which is a strange view in code written in golang.

### Clearer is better

In this pull request I removed `omitempty` tag and added `default` tag. The reason is `proxied` attribute in CloudFlare DNS API (Client Payload) is either required or optional but default is falsy. I supposed that it might be a better solution to make `DNSRecord.Proxied` explicit in order to call functions more clearly.

### BTW

 - It's a breaking change
 - Related pull request is: #595 

Yes, via given GitHub Actions workflows.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/flarectl/dns.go
+++ b/cmd/flarectl/dns.go
@@ -19,7 +19,7 @@ func formatDNSRecord(record cloudflare.DNSRecord) []string {
 		record.Content,
 		strconv.FormatInt(int64(record.TTL), 10),
 		strconv.FormatBool(record.Proxiable),
-		strconv.FormatBool(*record.Proxied),
+		strconv.FormatBool(record.Proxied),
 		strconv.FormatBool(record.Locked),
 	}
 }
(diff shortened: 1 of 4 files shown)


jacobbednarz:
thanks for the PR however, i'm afriad we won't be accepting it. dependent systems, such as the Terraform provider, rely on this to send a payload explicitly without the `proxied` value when setting the TTL.

> So to assign a truthy value, instead of true, I must use &[]bool{true}[0] or any other equal item, which is a strange view in code written in golang.

check out the [type conversion helpers](https://github.com/cloudflare/cloudflare-go/blob/master/convert_types.go) we ship with the library; they help with the syntax issues you're having.

zeithrold:
Thank you for your reply! But I might think that it's necessary to remind that in document.

PR #967, opened 2022-07-02
Intelligence API

### LINKED ISSUE #965: Add support for intelligence

v0.43.0

### Description

Add support for intelligence endpoint

https://api.cloudflare.com/#domain-intelligence-get-domain-details

https://api.cloudflare.com/#ip-intelligence-properties

https://api.cloudflare.com/#asn-intelligence-properties 

### Use cases

Be able to query intelligence cloudflare-go

```go
result, err := cloudflare.IntelligenceIP(...)
```

### References

Happy to work on this. Wanted to make sure that the beta status was okay with this project.

Cyb3r-Jak3:
Adds support for the intelligence API

## Description

Resolves #965 

Unit tests have been added for all functions.

## Types of changes

## Checklist:


cloudflare_test.go
```diff
@@ -534,3 +534,8 @@ func TestCheckResultInfo(t *testing.T) {
 		})
 	}
 }
+
+// testHTTPMethod makes sure that a request's method is what is expected.
+func testHTTPMethod(t *testing.T, expected string, r *http.Request) {
+	assert.Equal(t, expected, r.Method, "Expected method '%s', got %s", expected, r.Method)
+}
```
Cyb3r-Jak3, line 541: Lint fails because of this. I only added this to intelligence tests, which all use GET.

jacobbednarz: that's expected; the check here is saying that all uses of it are using the same value so it shouldn't be a parameter but instead hardcoded in. 

personally, i'm not a fan of methods like these as they take away the explictness of the tests. i'd rather for now that we continue the patterns we already have where this is asserted in the test itself.

intelligence_asn.go
```diff
@@ -0,0 +1,104 @@
...
+	Count        int      `json:"count,omitempty"`
+	Page         int      `json:"page,omitempty"`
+	PerPage      int      `json:"per_page,omitempty"`
+}
+
+// IntelligenceASNOverview get overview for an ASN number
+//
+// API Reference: https://api.cloudflare.com/#asn-intelligence-get-asn-overview
+func (api *API) IntelligenceASNOverview(ctx context.Context, params IntelligenceASNOverviewParameters) ([]ASNInfo, error) {
+	// Make sure Account ID is set
```
jacobbednarz, line 56: let's do without these style of comments. this [blog from stackoverlow](https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/) has a good write up of why comments like this are something to avoid and what to do about it if you think they are nessecary.

intelligence_asn_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+import (
+	"context"
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	"net/http"
+	"testing"
+)
+
+const TestASNNumber = "13335"
```
jacobbednarz, line 11: we don't want this constant exported

```suggestion
const testASNNumber = "13335"
```

jacobbednarz (approved):

PR #971, opened 2022-07-03
Add Tunnel API method to get connections for a tunnel

### LINKED ISSUE #970: Add API method to get the connections of a tunnel

0.43.0

### Description

This allows us to get the connections of a particular tunnel. Essentially, its a Go wrapper for the following REST endpoint
`GET accounts/:account_identifier/cfd_tunnel/:tunnel_id/connections`

### Use cases

This endpoint is not documented in here https://api.cloudflare.com/#cloudflare-tunnel-properties but I found this endpoint getting used in [cloudflared](https://github.com/cloudflare/cloudflared/blob/d2cb80333652cb24f7d128e81485a195a300f2ad/cfapi/tunnel.go#L170). This endpoint gives more data than the `Connections` property returned by `cloudflare.Tunnel()` and hence is very useful. Especially, it provides the agent architecture. Moreover, it provides the connections as a sub-array of each agent which can be very useful.

```go
func (api *API) TunnelConnections(ctx context.Context, params TunnelConnectionParams) ([]Connection, error) {
	if params.AccountID == "" {
		return []Connection{}, ErrMissingAccountID
	}

	if params.ID == "" {
		return []Connection{}, errors.New("missing tunnel ID")
	}

	uri := fmt.Sprintf("/accounts/%s/cfd_tunnel/%s/connections", params.AccountID, params.ID)

	if err != nil {
		return []Connection{}, err
	}

	var argoDetailsResponse TunnelConnectionResponse
	err = json.Unmarshal(res, &argoDetailsResponse)
	if err != nil {
		return []Connection{}, errors.Wrap(err, errUnmarshalError)
	}
	return argoDetailsResponse.Result, nil
}
```

### References
Providing a PR for this #971

SayakMukhopadhyay:
## Description

Closes #970 

Not yet

## Types of changes

## Checklist:


jacobbednarz:
we need to add test coverage for this change as well please.

SayakMukhopadhyay:
@jacobbednarz test has been added.

tunnel.go
```diff
@@ -407,6 +429,33 @@ func (api *API) GetTunnelConfiguration(ctx context.Context, params GetTunnelConf
...
+// TunnelConnections gets all connections on a tunnel.
+//
+// API reference: https://api.cloudflare.com/#cloudflare-tunnel-clean-up-cloudflare-tunnel-connections
+func (api *API) TunnelConnections(ctx context.Context, params TunnelConnectionParams) ([]Connection, error) {
+	if params.AccountID == "" {
+		return []Connection{}, ErrMissingAccountID
+	}
+
+	if params.ID == "" {
+		return []Connection{}, errors.New("missing tunnel ID")
```
jacobbednarz, line 441:
```suggestion
		return []Connection{}, ErrMissingTunnelID
```

tunnel_test.go
```diff
@@ -309,6 +309,78 @@ func TestGetTunnelConfiguration(t *testing.T) {
...
+				]
+			}]
+		}
+		`)
+	}
+
+	mux.HandleFunc("/accounts/"+testAccountID+"/cfd_tunnel/f174e90a-fafe-4643-bbbc-4a0ed4fc8415/connections", handler)
+
+	//createdAt, _ := time.Parse(time.RFC3339, "2009-11-10T23:00:00Z")
+	//deletedAt, _ := time.Parse(time.RFC3339, "2009-11-10T23:00:00Z")
```
jacobbednarz, line 352:
```suggestion
```

tunnel.go
```diff
@@ -407,6 +429,33 @@ func (api *API) GetTunnelConfiguration(ctx context.Context, params GetTunnelConf
...
+
+	res, err := api.makeRequestContext(ctx, http.MethodGet, uri, nil)
+	if err != nil {
+		return []Connection{}, err
+	}
+
+	var argoDetailsResponse TunnelConnectionResponse
+	err = json.Unmarshal(res, &argoDetailsResponse)
+	if err != nil {
+		return []Connection{}, errors.Wrap(err, errUnmarshalError)
```
jacobbednarz, line 454: you shouldn't use `errors.Wrap`, use `fmt.Errorf` instead.

tunnel.go
```diff
@@ -407,6 +429,33 @@ func (api *API) GetTunnelConfiguration(ctx context.Context, params GetTunnelConf
 	return tunnelDetails, nil
 }
 
+// TunnelConnections gets all connections on a tunnel.
+//
+// API reference: https://api.cloudflare.com/#cloudflare-tunnel-clean-up-cloudflare-tunnel-connections
```
jacobbednarz, line 434: this is incorrect; we don't yet have public API documentation for this endpoint. do you have a support ticket i can track this against?

tunnel.go
SayakMukhopadhyay: Ah! seems like this change was merged yesterday for all other components. I will rebase to the upstream master and fix this.

tunnel.go
SayakMukhopadhyay: Yeah, this endpoint is not documented but as mentioned in the linked issue, it is still getting used by cloudflared, that's how I found the existence of this endpoint. Actually, cloudflared won't be able to list all endpoints if this endpoint doesn't work, so I am inclined to believe that this endpoint is stable, just not documented.

I haven't created a ticket (since I am a free user) so I have created a community [post](https://community.cloudflare.com/t/documentation-unavailable-for-tunnel-connections-endpoint/397023). I am willing to contribute to the documentation too. Let me know what I can do to get this PR merged.

jacobbednarz: i'll poke the team but we really need public documentation here. `cloudflared` is run by the team that owns this service so they can make changes and know the impacts, however, we need to make sure the endpoints in use are also stable for public consumption; not just tools that the team maintain.

SayakMukhopadhyay: I understand, the main motivation for this PR is a [tool](https://github.com/beezlabs-org/cloudflare-tunnel-operator) that I am working on which would benefit from having this endpoint in the SDK. Moreover, I am just hoping that this endpoint was mistakenly not documented and hence the associated issue and this PR. I really hope this becomes documented though as this endpoint actually provides useful information. 

Moreover, the URL associated with the API already has a DELETE method associated with it. Having the GET method seems like a no brainer.

jacobbednarz: docs are coming and will land in the next couple of days.

jacobbednarz, line 434:
```suggestion
// API reference: https://api.cloudflare.com/#cloudflare-tunnel-list-cloudflare-tunnel-connections
```

SayakMukhopadhyay:
The docs have been updated 🎉

jacobbednarz (approved):

jacobbednarz:
thanks for getting this over the line @SayakMukhopadhyay 👏

SayakMukhopadhyay:
Thanks for approving the changes and making necessary fixes! 🙏

PR #989, opened 2022-07-13
tests: provide a `loadFixture` helper to clean up JSON outputs

jacobbednarz:
Provide a simpler approach to API payloads in tests.


Jjajsja (commented):
utils.go

PR #795, opened 2022-01-31
adds support for worker scripts using module syntax

kayteh:
## Description

Adds a `Module` field to WorkerScriptParams that changes the Content-Type of the resulting request to `application/javascript+module`. This isn't documented, but it is exactly what the Workers editor in the dashboard seems to do.

The only breaking change is on UploadWorker, which I believe we could avoid by omitting that entire change, or adding an UploadWorkerWithParams abstraction. I'm happy to hear either.

Closes #794 

## Has your change been tested?

Unit tests were added to cover at least one case for with and without bindings, I don't believe it's necessary to duplicate every UploadWorker test, as a single one can test the code flows correctly.

## Types of changes

What sort of change does your code introduce/modify?

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [ ] This change is using publicly documented (api.cloudflare.com or developers.cloudflare.com) and stable APIs.

Issue #1007, opened 2022-07-28
Ability to round-trip worker bindings


sodabrew:
### Current cloudflare-go version

0.45

### Description

I am writing a utility to download a script, modify it, and upload it again to a new name. 

It would be useful if the input and output formats for the Bindings structure matched up. I am reading the code to see if there's a simple enough way to transform the structures, but I figured I'd file this before I got too deep.

### Use cases

This is similar to the Workers Environments use case, so future work on that product feature might subsume this eventually, but download-modify-upload is still a valuable capability.

### Potential cloudflare-go usage

```go
cfapi, err := cloudflare.New(config.APIKey, config.APIEmail, cloudflare.UsingAccount(config.AccountID))
script, err := cfapi.DownloadWorker(ctx, &cloudflare.WorkerRequestParams{ ScriptName: "template" })
bindings, err := cfapi.ListWorkerBindings(ctx, &cloudflare.WorkerRequestParams{ ScriptName: "template" })
response, err := cfapi.UploadWorkerWithBindings(ctx,
  &cloudflare.WorkerRequestParams{
    ScriptName: "output",
  },
  &cloudflare.WorkerScriptParams{
    Script: script.Script,
    Bindings: bindings.BindingList,
  },
)
```

Unfortunately, this won't compile:
```
cannot use bindings.BindingList (variable of type []cloudflare.WorkerBindingListItem)
  as type map[string]cloudflare.WorkerBinding in struct literal
```

### References

Aha, this might do it. I feel like this comment is a clue that maybe the data structure mismatch ought to have been resolved in the code review process. How would you feel if I put up a PR to make UploadWorkerWithBindings accept an array instead of a map? Was the idea to use a map to avoid the case of two bindings with the same name? In that case, should the ListWorkerBindings return an array? Maybe a convenience method to munge between them?

https://github.com/cloudflare/cloudflare-go/blob/9a9f293157bf887f5e207fc9a3535401c3ce7f62/workers_test.go#L233-L240

sodabrew:
Got it. Right, so this feels totally unnecessary impedance mismatch. Guidance on a PR to not need this code would be greatly appreciated!
```go
bindingMap:= make(map[string]cloudflare.WorkerBinding)
for _, binding := range bindings.BindingList {
        bindingMap[binding.Name] = binding.Binding
}
```

End-to-end working example:

```go
[...]

PR #1010, opened 2022-08-02
add module upload support

### LINKED ISSUE #794: Support for UploadWorker as a module

0.31.0

### Description

Workers supports uploading ES Module syntax now. This might be lightly documented in API docs, but ESM-based workers are uploaded with `Content-Type: application/javascript+module`, as seen when uploading from the Workers editor in the dashboard.

cloudflare-go doesn't have any way to make the distinction, and would need to be added around [here for UploadWorker](https://github.com/cloudflare/cloudflare-go/blob/master/workers.go#L495) and [here via UploadWorkerWithBindings](https://github.com/cloudflare/cloudflare-go/blob/master/workers.go#L552)

The likely best option, in my opinion, is to

- Add a `Module: bool` field to `WorkerScriptParams`
- Change the call signature of `UploadWorker` to `UploadWorker(ctx context.Context, requestParams *WorkerRequestParams, data *WorkerScriptParams)`, and ignore the `Bindings` field.

Attempting to upload a valid worker without this sort of change results in a response: 
```json
{
  "code": 10021,
  "message": "Uncaught SyntaxError: Unexpected token 'export'\n  at line 1\n"
}
```

### Use cases

Uploading Worker assets from Terraform

```go
scriptParams := &WorkerScriptParams{
  Script: `export default { async fetch() { return new Response('hello-world') } }`,
  Bindings: workerBindings, // ignored in UploadWorker()
  Module: true, // changes `application/javascript` to `application/javascript+module`
}

client.UploadWorkerWithBindings(ctx, requestParams, scriptParams)
client.UploadWorker(ctx, requestParams, scriptParams)
```

### References

- cloudflare/terraform-provider-cloudflare#1417

jgodson:
PR to upstream the changes from https://github.com/Shopify/cloudflare-go/pull/16 that adds support for uploading scripts that use the module syntax.

It builds upon the changes started [here](https://github.com/cloudflare/cloudflare-go/pull/795)

Closes https://github.com/cloudflare/cloudflare-go/issues/794

### DIFF
--- a/.changelog/1010.txt
+++ b/.changelog/1010.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+workers: Add support to upload module workers
+```
--- a/workers.go
+++ b/workers.go
@@ -26,6 +26,10 @@ type WorkerRequestParams struct {
 type WorkerScriptParams struct {
 	Script string
 
+	// Module changes the Content-Type header to specify the script is an
+	// ES Module syntax script.
+	Module bool
+
 	// Bindings should be a map where the keys are the binding name, and the
 	// values are the binding content
 	Bindings map[string]WorkerBinding
@@ -592,11 +596,16 @@ func (api *API) ListWorkerScripts(ctx context.Context) (WorkerListResponse, erro
 // UploadWorker push raw script content for your worker.
 //
 // API reference: https://api.cloudflare.com/#worker-script-upload-worker
-func (api *API) UploadWorker(ctx context.Context, requestParams *WorkerRequestParams, data string) (WorkerScriptResponse, error) {
-	if requestParams.ScriptName != "" {
-		return api.uploadWorkerWithName(ctx, requestParams.ScriptName, "application/javascript", []byte(data))
+func (api *API) UploadWorker(ctx context.Context, requestParams *WorkerRequestParams, params *WorkerScriptParams) (WorkerScriptResponse, error) {
+	if params.Module {
+		return api.UploadWorkerWithBindings(ctx, requestParams, params)
 	}
-	return api.uploadWorkerForZone(ctx, requestParams.ZoneID, "application/javascript", []byte(data))
+
+	contentType := "application/javascript"
+	if requestParams.ScriptName != "" {
+		return api.uploadWorkerWithName(ctx, requestParams.ScriptName, contentType, []byte(params.Script))
+	}
+	return api.uploadWorkerForZone(ctx, requestParams.ZoneID, contentType, []byte(params.Script))
 }
 
 // UploadWorkerWithBindings push raw script content and bindings for your worker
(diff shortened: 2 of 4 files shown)


sodabrew:
Awesome! Don't say I never shopped on Shopify :) 

Since I'm just working on a round-trip use case over in #1007, I'll note that `DownloadWorker` also needs to handle module-format workers. There is only an example for a plain worker at https://api.cloudflare.com/#worker-script-download-worker

```
$ curl <auth headers> https://api.cloudflare.com/client/v4/accounts/<accountid>/workers/scripts/<scriptname>

Content-Type: application/json
[rest of the headers]

self.addEventListener... [rest of the script]
```

The same API call for a Module-format worker returns multipart that would need to be parsed. What should happen if there are multiple parts?
```
$ curl <auth headers> https://api.cloudflare.com/client/v4/accounts/<accountid>/workers/scripts/<scriptname>

Content-Type: multipart/form-data; boundary=abcdefg
[rest of the headers]

--abcdefg
Content-Disposition: form-data; name="worker.js"

export default {
  fetch(request) {
    return new Response("Hello world")
  },
};
--abcdefg--
```

The return type from `cloudflare.DownloadWorker` is
```
type WorkerScriptResponse struct {
        Response
        WorkerScript `json:"result"`
}
```

But the code for the method actually does this:
```
...
        var r WorkerScriptResponse
        r.Script = string(res)
        r.Success = true
        return r, nil
}
```

I'm not sure I follow what's going on there with how Go is inferring and passing types from that method (limitations of my Go knowledge)

sodabrew:
Got it, the struct fields are being promoted from nested structs
```
type WorkerScriptResponse struct {
        Response
        WorkerScript `json:"result"`
}

type WorkerScript struct {
        WorkerMetaData
        Script     string `json:"script"`
        UsageModel string `json:"usage_model,omitempty"`
}

type WorkerMetaData struct {
        ID         string    `json:"id,omitempty"`
        ETAG       string    `json:"etag,omitempty"`
        Size       int       `json:"size,omitempty"`
        CreatedOn  time.Time `json:"created_on,omitempty"`
        ModifiedOn time.Time `json:"modified_on,omitempty"`
}
```

Thing is, most of these fields are not set by the `DownloadWorker` response!

So back to the issue at hand: is it possible for a DownloadWorker to respond with multiple module files? If so, what's the best way to represent that as a Go structure? (I'm thinking `map[string]WorkerScript` where the key is the module file name)

dotansimha (approved):

jgodson:
> Is it possible for a DownloadWorker to respond with multiple module files?

That is an excellent question. I'm not really sure as I haven't tried that/needed to do that yet. 

I believe that the upload functionality at least is covered here, so if download could be implemented separately, this should be good to go as is (at least it has been working for us so far).

jacobbednarz:
@jgodson i'm hoping to getting to review/test this in the coming days. are you able to address the CHANGELOG entry so there is one less thing to do before we merge?

jacobbednarz:
thanks for your patience here @jgodson. the PR looks great and my (basic) testing of this internally is looking great.

sodabrew:
@jacobbednarz Glad to see this land! Where are you at in terms of thinking about how to solve for DownloadWorker to return something sensible when a module worker is fetched? https://github.com/cloudflare/cloudflare-go/pull/1010#issuecomment-1203209116

jacobbednarz:
haven't really dug into the issues with `DownloadWorker` and i'd have to chat with the team about how they intend to solve it for multiple workers.

PR #1014, opened 2022-08-03
Support for attaching a Worker to a Domain

sodabrew:
## Description

Support for the new Custom Domain API at https://api.cloudflare.com/#worker-domain-attach-to-domain

Works for me locally, unit tests will be added in PR review.

## Types of changes

## Checklist:

- [x] This change is using publicly documented (api.cloudflare.com or developers.cloudflare.com) 
- [ ] ... and stable APIs. **unclear**.

### DIFF
--- a/workers.go
+++ b/workers.go
@@ -850,7 +850,7 @@ func getRouteEndpoint(route WorkerRoute) (string, error) {
 	return "routes", nil
 }
 
-type WorkerDomain struct {
+type WorkerDomainParams struct {
 	ZoneID      string `json:"zone_id"`
 	Hostname    string `json:"hostname"`
 	Service     string `json:"service"`
@@ -874,16 +874,18 @@ type WorkerDomainResponse struct {
 // AttachWorkerToDomain attaches a worker to a zone and hostname
 //
 // API reference: https://api.cloudflare.com/#worker-domain-attach-to-domain
-func (api *API) AttachWorkerToDomain(ctx context.Context, params *WorkerDomain) (WorkerDomainResponse, error) {
-	uri := fmt.Sprintf("/accounts/%s/workers/domains", api.AccountID)
+func (api *API) AttachWorkerToDomain(ctx context.Context, rc *ResourceContainer, params *WorkerDomainParams) (WorkerDomainResponse, error) {
+	uri := fmt.Sprintf("/accounts/%s/workers/domains", rc.Identifier)
 	res, err := api.makeRequestContext(ctx, http.MethodPut, uri, params)
 	if err != nil {
 		return WorkerDomainResponse{}, err
 	}
+
 	var r WorkerDomainResponse
 	err = json.Unmarshal(res, &r)
 	if err != nil {
 		return WorkerDomainResponse{}, fmt.Errorf("%s: %w", errUnmarshalError, err)
 	}
+
 	return r, nil
 }
--- a/workers.go
+++ b/workers.go
@@ -851,18 +851,18 @@ func getRouteEndpoint(route WorkerRoute) (string, error) {
 }
 
 type WorkerDomain struct {
-	ZoneID string `json:"zone_id"`
-	Hostname string `json:"hostname"`
-	Service string `json:"service"`
+	ZoneID      string  `json:"zone_id"`
+	Hostname    string  `json:"hostname"`
+	Service     string  `json:"service"`
 	Environment *string `json:"environment,omitempty"`
 }
 
 type WorkerDomainResult struct {
-	ID string `json:"id"`
-	ZoneID string `json:"zone_id"`
-	ZoneName string `json:"zone_name"`
-	Hostname string `json:"hostname"`
-	Service string `json:"service"`
+	ID          string `json:"id"`
+	ZoneID      string `json:"zone_id"`
+	ZoneName    string `json:"zone_name"`
+	Hostname    string `json:"hostname"`
+	Service     string `json:"service"`
 	Environment string `json:"environment"`
 }
 
@@ -877,9 +877,9 @@ type WorkerDomainResponse struct {
 func (api *API) AttachWorkerToDomain(ctx context.Context, zoneID, hostname, service string, environment *string) (WorkerDomainResponse, error) {
 	uri := fmt.Sprintf("/accounts/%s/workers/domains", api.AccountID)
 	body := WorkerDomain{
-		ZoneID: zoneID,
-		Hostname: hostname,
-		Service: service,
+		ZoneID:      zoneID,
+		Hostname:    hostname,
+		Service:     service,
 		Environment: environment,
 	}
 	res, err := api.makeRequestContext(ctx, http.MethodPut, uri, body)
--- a/workers.go
+++ b/workers.go
@@ -849,3 +849,47 @@ func getRouteEndpoint(route WorkerRoute) (string, error) {
 
 	return "routes", nil
 }
+
+type WorkerDomain struct {
+	ZoneID string `json:"zone_id"`
+	Hostname string `json:"hostname"`
+	Service string `json:"service"`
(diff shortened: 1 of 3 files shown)


sodabrew:
Working on tests next.

@jacobbednarz should I switch the arguments to be inside a struct? Some of the methods use a strict while others take direct arguments. I'm leaning towards switching to the struct pattern. Let me know!

jacobbednarz:
check out https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md, #1028, and https://github.com/cloudflare/cloudflare-go/blob/master/tunnel.go for prior art on how we should be adding new methods.

sodabrew:
@jacobbednarz Updated per your examples, please review!

It seems like `cloudflare.ZoneIdentifier` could be used for this new call, but there's no other user of it in workers.go. https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md

In workers.go, I followed the example of DeleteWorker, DownloadWorker, etc. that have ZoneID in the respective params structs.

jacobbednarz:
> It seems like `cloudflare.ZoneIdentifier` could be used for this new call, but there's no other user of it in workers.go.

there are currently no other implementation of `ResourceContainer` in the workers methods because they 1) predate the change 2) are pretty messy when it comes to what can live at what resource level. it hasn't been touched (but really should be improved) as the Workers API has evolved quite a bit since it was introduced and most of it is no longer needed.

going forward, we want to use the updated method signature for all new methods to help future proof against these issues.

sodabrew:
> going forward, we want to use the updated method signature for all new methods to help future proof against these issues.

Just to be clear, does that mean I should add the new method using the new style in a manner differently than my latest commit, or that you'll circle back to workers.go at a later date and redo all the method signatures in one go?

PS - I would expect to squash all commits in this PR, let me know if you'd like to do that as part of a review and merge or if I should force push my branch. I am very comfortable with rebase workflow, so this is no hardship at my end.

jacobbednarz:
> does that mean I should add the new method using the new style in a manner differently than my latest commit

sorry if it wasn't clear. you should do this (i'll add a note inline as a reminder).

> PS - I would expect to squash all commits in this PR, let me know if you'd like to do that as part of a review and merge or if I should force push my branch. I am very comfortable with rebase workflow, so this is no hardship at my end.

no need, i'm happy to merge/squash at the end.

workers.go
```diff
@@ -849,3 +849,41 @@ func getRouteEndpoint(route WorkerRoute) (string, error) {
...
+	if err != nil {
+		return WorkerDomainResponse{}, err
+	}
+	var r WorkerDomainResponse
+	err = json.Unmarshal(res, &r)
+	if err != nil {
+		return WorkerDomainResponse{}, fmt.Errorf("%s: %w", errUnmarshalError, err)
+	}
+	return r, nil
+}
```
jacobbednarz:
```suggestion
func (api *API) AttachWorkerToDomain(ctx context.Context, rc *ResourceContainer, params *WorkerDomainParams) (WorkerDomainResponse, error) {
	uri := fmt.Sprintf("/accounts/%s/workers/domains", rc.Identifier)
	res, err := api.makeRequestContext(ctx, http.MethodPut, uri, params)
	if err != nil {
		return WorkerDomainResponse{}, err
	}
	
	var r WorkerDomainResponse
	err = json.Unmarshal(res, &r)
	if err != nil {
		return WorkerDomainResponse{}, fmt.Errorf("%s: %w", errUnmarshalError, err)
	}
	
	return r, nil
}
```

workers.go
```diff
@@ -849,3 +849,41 @@ func getRouteEndpoint(route WorkerRoute) (string, error) {
 
 	return "routes", nil
 }
+
+type WorkerDomain struct {
```
jacobbednarz:
```suggestion
type WorkerDomainParams struct {
```

workers_test.go
```diff
@@ -1112,3 +1125,26 @@ func TestWorkers_UpdateWorkerRouteWithNoScript(t *testing.T) {
 	_, err := client.UpdateWorkerRoute(context.Background(), "foo", "e7a57d8746e74ae49c25994dadb421b1", route)
 	assert.NoError(t, err)
 }
+
+func TestWorkers_AttachWorkerToDomain(t *testing.T) {
+	setup(UsingAccount("foo"))
```
jacobbednarz:
```suggestion
```

workers_test.go
```diff
@@ -1112,3 +1125,26 @@ func TestWorkers_UpdateWorkerRouteWithNoScript(t *testing.T) {
...
+	defer teardown()
+
+	mux.HandleFunc("/accounts/foo/workers/domains", func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, http.MethodPut, r.Method, "Expected method 'PUT', got %s", r.Method)
+		w.Header().Set("content-type", "application/json")
+		fmt.Fprintf(w, deleteWorkerResponseData) //nolint
+	})
+	res, err := client.AttachWorkerToDomain(context.Background(), &WorkerDomain{
```
jacobbednarz:
```suggestion
	res, err := client.AttachWorkerToDomain(context.Background(), AccountIdentifier(testAccountID), &WorkerDomainParams{
```

workers_test.go
```diff
@@ -1112,3 +1125,26 @@ func TestWorkers_UpdateWorkerRouteWithNoScript(t *testing.T) {
...
+		ZoneID:      "foo",
```
jacobbednarz:
```suggestion
		ZoneID:      testZoneID,
```

workers_test.go
```diff
@@ -1112,3 +1125,26 @@ func TestWorkers_UpdateWorkerRouteWithNoScript(t *testing.T) {
 	_, err := client.UpdateWorkerRoute(context.Background(), "foo", "e7a57d8746e74ae49c25994dadb421b1", route)
 	assert.NoError(t, err)
 }
+
+func TestWorkers_AttachWorkerToDomain(t *testing.T) {
+	setup(UsingAccount("foo"))
+	defer teardown()
+
+	mux.HandleFunc("/accounts/foo/workers/domains", func(w http.ResponseWriter, r *http.Request) {
```
jacobbednarz:
```suggestion
	mux.HandleFunc("/accounts/"+testAccountID+"/workers/domains", func(w http.ResponseWriter, r *http.Request) {
```

workers_test.go
```diff
@@ -165,6 +165,18 @@ const (
   "success": true,
   "errors": [],
   "messages": []
+}`
+	attachWorkerToDomainResponse = `{
+    "result": {
+        "id": "e7a57d8746e74ae49c25994dadb421b1",
+	"zone_id":"foo",
```
jacobbednarz: you'll want to use `testZoneID` here instead.

sodabrew: Need to move out of the const block to the var block in order to use fmt.Sprintf here

workers_test.go
sodabrew: This is incorrect; without this line, the tests fail to connect to localhost:
```
--- FAIL: TestWorkers_AttachWorkerToDomain (0.00s)
    workers_test.go:1148:
        	Error Trace:	/Users/aaron/Code/cloudflare-go/workers_test.go:1148
        	Error:      	Received unexpected error:
        	            	HTTP request failed: Put "http://127.0.0.1:59180/accounts/01a7362d577a6c3019a474fd6f485823/workers/domains": dial tcp 127.0.0.1:59180: connect: connection refused
        	Test:       	TestWorkers_AttachWorkerToDomain
FAIL
FAIL	github.com/cloudflare/cloudflare-go	1.076s
FAIL
```

jacobbednarz:
thanks! appreciate the effort here!

sodabrew:
Thank you!

PR #1016, opened 2022-08-04
filter/firewall-rules fix for double API calls

### LINKED ISSUE #1024: Fix pagination for filter/firewall-rules API
### Confirmation

### cloudflare-go version

latest

### Go environment

GO111MODULE=""
GOARCH="arm64"
GOBIN=""
GOCACHE="/Users/tjozsa/Library/Caches/go-build"
GOENV="/Users/tjozsa/Library/Application Support/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="arm64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/tjozsa/go/pkg/mod"
GONOPROXY="code.cfops.it"
GONOSUMDB="code.cfops.it"
GOOS="darwin"
GOPATH="/Users/tjozsa/go"
GOPRIVATE="code.cfops.it"
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/opt/homebrew/Cellar/go/1.18.5/libexec"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/opt/homebrew/Cellar/go/1.18.5/libexec/pkg/tool/darwin_arm64"
GOVCS=""
GOVERSION="go1.18.5"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/dev/null"
GOWORK=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -arch arm64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/81/zyf3xcqd6wv1nh70f4pspr900000gp/T/go-build1439680234=/tmp/go-build -gno-record-gcc-switches -fno-common"

### Expected output

`api.FirewallRules` and `api.Filters` calling the public API only once with required params for paginations.

### Actual output

They submit at least two requests:
```
/zones/b1fbb152bbde3bd28919a7f4bdca841f/firewall/rules
/zones/b1fbb152bbde3bd28919a7f4bdca841f/firewall/rules?page=1&per_page=50
```

I've modified cloudflare-go a bit to present the problem I'm seeing. A simple printing out the `uri` we are calling in cloudflare.go `makeRequestContext`

```
func (api *API) makeRequestContext(ctx context.Context, method, uri string, params interface{}) ([]byte, error) {
	fmt.Println(uri)
	return api.makeRequestWithAuthType(ctx, method, uri, params, api.authType)
}
```

Then when I called in my example code:
```
fwRules, _, err := api.FirewallRules(context.Background(), cloudflare.ZoneIdentifier("b1fbb152bbde3bd28919a7f4bdca841f"), cloudflare.FirewallRuleListParams{})
```

stdout follows:
```
/zones/b1fbb152bbde3bd28919a7f4bdca841f/firewall/rules
/zones/b1fbb152bbde3bd28919a7f4bdca841f/firewall/rules?page=1&per_page=50
```

### Steps to reproduce

see code demonstration

### References

I found this while I was upgrading cf-terraforming to use the latest cloudflare-go. My tests were not passing, because they needed an additional interaction:
- https://github.com/cloudflare/cf-terraforming/blob/49cc60900f2562c287dfa68fec6d90352268b17f/testdata/cloudflare/cloudflare_filter.yaml#L4-L10
- https://github.com/cloudflare/cf-terraforming/blob/49cc60900f2562c287dfa68fec6d90352268b17f/testdata/cloudflare/cloudflare_filter.yaml#L43-L49

tamas-jozsa:
## Description

Recent pagination updates to these APIs are causing endpoints to be called twice. This PR aims to fix them.

Unit tests are passing, and I tried this version in cf-terraforming, where it's working as expected

## Types of changes

## Checklist:

Closes #1024

### DIFF
--- a/.changelog/1016.txt
+++ b/.changelog/1016.txt
@@ -1,0 +1,7 @@
+```release-note:bug
+firewall_rule: fix double endpoint calls & moving towards common method signature
+```
+
+```release-note:bug
+filter: fix double endpoint calls & moving towards common method signature
+```
--- a/.changelog/1016.txt
+++ b/.changelog/1016.txt
@@ -1,7 +1,7 @@
-```release-note:bug
+```release-note:enhancement
 firewall_rule: fix double endpoint calls & moving towards common method signature
 ```
 
-```release-note:bug
+```release-note:enhancement
 filter: fix double endpoint calls & moving towards common method signature
 ```
--- a/filter.go
+++ b/filter.go
@@ -60,6 +60,26 @@ type FilterValidationExpressionMessage struct {
 	Message string `json:"message"`
 }
 
+// FilterCreateParams contains required and optional params
+// for creating a filter.
+type FilterCreateParams struct {
+	ID          string `json:"id,omitempty"`
+	Expression  string `json:"expression"`
+	Paused      bool   `json:"paused"`
+	Description string `json:"description"`
+	Ref         string `json:"ref,omitempty"`
+}
+
+// FilterUpdateParams contains required and optional params
+// for updating a filter.
+type FilterUpdateParams struct {
+	ID          string `json:"id"`
+	Expression  string `json:"expression"`
+	Paused      bool   `json:"paused"`
+	Description string `json:"description"`
+	Ref         string `json:"ref,omitempty"`
+}
+
 type FilterListParams struct {
 	ResultInfo
 }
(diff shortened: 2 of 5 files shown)


filter.go
```diff
@@ -61,6 +61,8 @@ type FilterValidationExpressionMessage struct {
 }
 
 type FilterListParams struct {
+	PerPage uint16
```
jacobbednarz: this isn't required; `ResultInfo` already has an embedded pagination struct.

filter.go
```diff
@@ -91,20 +93,8 @@ func (api *API) Filter(ctx context.Context, rc *ResourceContainer, filterID stri
 //
 // API reference: https://developers.cloudflare.com/firewall/api/cf-filters/get/#get-all-filters
 func (api *API) Filters(ctx context.Context, rc *ResourceContainer, params FilterListParams) ([]Filter, *ResultInfo, error) {
-	uri := buildURI(fmt.Sprintf("/zones/%s/filters", rc.Identifier), params)
```
jacobbednarz, line 94: this won't work for the multiple pages condition as it requires the `ResultInfo` to have the pagination populated to move the pager. 

what is the issue you're trying to solve here? right now, I'm not particularly fussed with the additional HTTP call providing it's not duplicating results in the array. I have an idea to remove it however, it requires making all endpoints use pagination by default which will need some digging internally to ensure we don't break things.

jacobbednarz (changes requested):

filter.go
```diff
@@ -128,6 +118,18 @@ func (api *API) Filters(ctx context.Context, rc *ResourceContainer, params Filte
 			params.ResultInfo = fResponse.ResultInfo.Next()
 		}
```
tamas-jozsa: @jacobbednarz this part of the code handles the multi-page condition: https://github.com/cloudflare/cloudflare-go/pull/1016/files#diff-8a6290ede135299ac187338deaf28c9f658a4e45634d78c9e2ef113bbbd7ef67R101-R119 

The problem I intend to solve is a duplicate call of 
```
```

When you call `Filters` like this:
```
var filterListParams = FilterListParams{
	PerPage: 25,
	Page:    1,
}

actual, _, err := client.Filters(context.Background(), ZoneIdentifier("d56084adb405e0b7e32c52321bf07be6"), filterListParams)
```

filter.go
tamas-jozsa: I've opened a bug describing my findings: https://github.com/cloudflare/cloudflare-go/issues/1024

tamas-jozsa: And I also updated my PR and verified that the changes work as expected.

jacobbednarz: the `List*` style methods need to support automatic pagination _and_ if someone passes in their own pagination options. i've pulled this locally and confirmed it's still not fixed for the scenario where we allow manual pagination. below is an example of it not working:

```go
package main

import (
	"context"
	"log"

	"github.com/cloudflare/cloudflare-go"
	"github.com/davecgh/go-spew/spew"
)

func main() {

	api, err := cloudflare.New(os.Getenv("CLOUDFLARE_API_KEY"), os.Getenv("CLOUDFLARE_API_EMAIL"))
	if err != nil {
		log.Fatal(err)
	}
	ctx := context.Background()

	filters, res, _ := api.Filters(ctx, cloudflare.ZoneIdentifier("0da42c8d2132a9ddaf714f9e7c920711"), cloudflare.FilterListParams{ResultInfo: cloudflare.ResultInfo{Page: 2}})

	spew.Dump(len(filters), res) 
    // => filters should be > 0 however it is empty and `res` shows an empty ResultInfo
    // 
	// (*cloudflare.ResultInfo)(0xc0000e82d8)({
	// 	Page: (int) 0,
	// 	PerPage: (int) 0,
	// 	TotalPages: (int) 0,
	// 	Count: (int) 0,
	// 	Total: (int) 0,
	// 	Cursor: (string) "",
	// 	Cursors: (cloudflare.ResultInfoCursors) {
	// 		Before: (string) "",
	// 		After: (string) ""
	// 	}
	// })
```

tamas-jozsa: Gotcha, I missed that requirement. I've then introduced an `autoPaginate` var that I'll set to false in case either `Page` or `PerPage` is provided. It looks better know I think.

However, I noticed, that not all `PerPage` values are supported. For example, if you provide `3`, you get 
```
filters.api.unknown_error
```
Wonder if we could handle this better. 

Also, I further updated method signatures as discussed.

jacobbednarz: > Gotcha, I missed that requirement. I've then introduced an autoPaginate var that I'll set to false in case either Page or PerPage is provided. It looks better know I think.

definite improvement however, i'm really not happy with how we are handling pagination as a whole because this is quite verbose to add in all endpoints that need to support it. i'd really prefer if we had a way in the `ListParams` to explicitly say we only intended to fetch a single result. something we can work on with the experimental client down the track i guess.

> Wonder if we could handle this better.

something to raise with the service team; i suspect there are minimums in the per page counts as < 5 (or even 10) don't really make sense for pagination.

jacobbednarz:
if you add a CHANGELOG entry (per the automation), we'll get this merged. thanks!

.changelog/1016.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:bug
```
jacobbednarz:
```suggestion
```release-note:enhancement
```

.changelog/1016.txt
```diff
@@ -0,0 +1,7 @@
...
+firewall_rule: fix double endpoint calls & moving towards common method signature
+```
+
+```release-note:bug
```
jacobbednarz:
```suggestion
```release-note:enhancement
```

filter.go
tamas-jozsa: I agree with you. We can do better and abstract away pagination logic in a more efficient and reusable manner. 

FWIW, you can retrieve a single results with the current setup from any page. 

For example, if you want one result from page 2:
```
fwRules, resultInfo, err := api.Filters(context.Background(), 
  cloudflare.ZoneIdentifier("b1fbb152bbde3bd28919a7f4bdca841f"), 
  cloudflare.FilterListParams{ResultInfo: cloudflare.ResultInfo{PerPage: 1, Page: 2}})
```

Or 1 result from page 1
```
fwRules, resultInfo, err := api.Filters(context.Background(), 
  cloudflare.ZoneIdentifier("b1fbb152bbde3bd28919a7f4bdca841f"), 
  cloudflare.FilterListParams{ResultInfo: cloudflare.ResultInfo{PerPage: 1}})
```

Or everything from page 2
```
fwRules, resultInfo, err := api.Filters(context.Background(), 
  cloudflare.ZoneIdentifier("b1fbb152bbde3bd28919a7f4bdca841f"), 
  cloudflare.FilterListParams{ResultInfo: cloudflare.ResultInfo{Page: 2}})
```

PR #1017, opened 2022-08-04
zone-lockdown: paginating list response

### LINKED ISSUE #1023: Pagination for zone-lockdown

latest

### Description

`ListZoneLockdowns` should paginate through the results and return all `ZoneLockdowns`. It is especially important in cf-terraforming, that we can generate every zone lockdowns.

Also, taking the chance to bring methods to a common method signature, based on https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md 

@jacobbednarz the documentation mentions `Nested methods and services` like `client.Zones.List()`. But I haven't found any case like that. Shall I update ZoneLockedowns similarly? `client.ZoneLockdown.List()`

Also, in the example you shared: https://github.com/cloudflare/cloudflare-go/blob/master/tunnel.go some method signatures differ from what the documentation suggests (sort of).
```
func (api *API) CreateTunnel(ctx context.Context, rc *ResourceContainer, params TunnelCreateParams) (Tunnel, error) 
func (api *API) UpdateTunnel(ctx context.Context, rc *ResourceContainer, params TunnelUpdateParams) (Tunnel, error)
```

Should we focus on a similar signature where we define different params type similar to (`TunnelCreateParams, TunnelUpdateParams`)

Or should we use a more standard example like:
```
func (api *API) CreateZoneLockdown(ctx context.Context, rc *ResourceContainer, ld ZoneLockdown) (ZoneLockdown, error)
func (api *API) UpdateZoneLockdown(ctx context.Context, rc *ResourceContainer, ld ZoneLockdown) (ZoneLockdown, error)
```

### Use cases

cf-terraforming has to generate every zone lockdown.

```go
-
```

### References

_No response_

tamas-jozsa:
## Description

I'm adding pagination for `ListZoneLockdowns` API. Additionally, I'm bringing Lockdown APIs to a common pattern. Instead of returning `ZoneLockdownResponse` they will return `ZoneLockdown` so they are easier to use.

Unit tests are working

## Types of changes

## Checklist:

### DIFF
--- a/lockdown.go
+++ b/lockdown.go
@@ -5,8 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"net/url"
-	"strconv"
 	"time"
 )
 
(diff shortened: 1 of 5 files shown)


jacobbednarz (changes requested):
instead of introducing a change that takes us further away from the [experimental method signatures](https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md), let's update it to move towards it.

check out https://github.com/cloudflare/cloudflare-go/blob/master/tunnel.go for an example of where we should head.

lockdown.go
```diff
@@ -46,109 +44,129 @@ type ZoneLockdownListResponse struct {
 	ResultInfo `json:"result_info"`
 }
 
+type LockdownListParams struct {
+	PerPage uint16
+	Page    uint16
+	ResultInfo
```
jacobbednarz, line 67: `ResultInfo` already has pagination structs embedded so this isn't required - https://github.com/cloudflare/cloudflare-go/blob/6fc060e04d39e4eda023336e4854871352a1edf4/cloudflare.go#L427-L435

tamas-jozsa:
> instead of introducing a change that takes us further away from the [experimental method signatures](https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md), let's update it to move towards it.
> 
> check out https://github.com/cloudflare/cloudflare-go/blob/master/tunnel.go for an example of where we should head.

That's very useful, thanks Jacob! I've opened an issue for further clarifications: [1023](https://github.com/cloudflare/cloudflare-go/issues/1023)

PR #1019, opened 2022-08-05
Removing read-only managed ruleset phases

### LINKED ISSUE #1022: Managed header ruleset phases shouldn't be part of cloudflare_rulesets

latest

### Description

We faced a couple issues in TF because of how the managed header rulesets were being treated/evolving. 

The initial problem was the two phases http_response_headers_transform_managed and http_request_late_transform_managed were introduced, so support was added in TF. Then those phases became read-only, which created an issue when we were attempting to copy the config from one zone to another (for the monarch project). The final state there was that those would be hidden from the GET rulesets response, which is great because they are exposed as a settings endpoint now, which has TF support.

This API is stable, GA and not going to change.  The kind=managed rulesets will remain read-only, and the kind=zone,root,custom will remain read-write.

The key bit is that the kind=managed rulesets aren't resources that belong to customer accounts—they're resources which a customer account can refer to in rules, but not modify. In summary: I think that terraform should ignore any ruleset of kind=managed, always.  But in any other case, it should be able to take ownership of resources in a ruleset it can see, regardless of the phase.  Internal phases should simply never be seen in the public API.

I also see, that there is a new resource for managed headers: https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/managed_headers 
I believe, it's going to be used for changing settings, that we were doing with these ruleset resources before:

- http_response_headers_transform_managed
- http_response_firewall_managed
- http_request_late_transform_managed
- http_request_firewall_managed

Though when I try to generate `cf-terraforming generate --resource-type cloudflare_managed_headers`
I see, that it needs some work in cf_terraforming.
```
"cloudflare_managed_headers" is not yet supported for automatic generation
```

### Use cases

Since the managed phases are now hidden from the API, it doesn't cause any problem, but we should clean them up, so it wouldn't lead to misunderstandings.

```go
-
```

### References

_No response_

tamas-jozsa:
## Description

Managed ruleset phases are read-only, therefore it does not make sense to keep them here. The Cloudflare terraform provider would fail, if it'd receive such resources to create.

This PR removes the following managed ruleset phases:

- http_response_headers_transform_managed
- http_response_firewall_managed
- http_request_late_transform_managed
- http_request_firewall_managed

no

## Types of changes

## Checklist:

### DIFF
--- a/rulesets.go
+++ b/rulesets.go
@@ -17,27 +17,23 @@ const (
 	RulesetKindSchema  RulesetKind = "schema"
 	RulesetKindZone    RulesetKind = "zone"
 
-	RulesetPhaseDDoSL4                              RulesetPhase = "ddos_l4"
-	RulesetPhaseDDoSL7                              RulesetPhase = "ddos_l7"
-	RulesetPhaseHTTPCustomErrors                    RulesetPhase = "http_custom_errors"
-	RulesetPhaseHTTPLogCustomFields                 RulesetPhase = "http_log_custom_fields"
-	RulesetPhaseHTTPRequestCacheSettings            RulesetPhase = "http_request_cache_settings"
-	RulesetPhaseHTTPRequestFirewallCustom           RulesetPhase = "http_request_firewall_custom"
-	RulesetPhaseHTTPRequestFirewallManaged          RulesetPhase = "http_request_firewall_managed"
-	RulesetPhaseHTTPRequestLateTransform            RulesetPhase = "http_request_late_transform"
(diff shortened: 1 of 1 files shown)


jacobbednarz (changes requested):
the assumptions here are not entirely correct; these phases are required for managing overrides within a configuration. you can see examples of an abstracted usage via Terraform at https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/ruleset.

tamas-jozsa:
Cancelling PR based on the feedback in https://github.com/cloudflare/cloudflare-go/issues/1022
I should rather filter down to resources from cf-terraforming, instead of couldflare-go.

PR #1032, opened 2022-08-10
certificate_packs: deprecate "custom" in favour of ACM

jacobbednarz:
Triggered by cloudflare/cf-terraforming#443 validation mismatches, I
went ahead and updated `CertificatePacks` to only reference ACM
configuration now that dedicated custom/custom certificates are no more.


jacobbednarz:
@nickysemenza do you mind eyeballing this one for me? we had some failures in cf-terraforming which lead me here to clean up the structs and methods.

PR #1035, opened 2022-08-11
Should fix the R2 bucket create endpoint

### LINKED ISSUE #1033: R2 bucket creation is broken due to API change
### Confirmation

### Expected output

Able to create R2 bucket

### Actual output

✘ [ERROR] A request to the Cloudflare API (/accounts/<account id>/r2/buckets/bucketname) failed.

  No route matches this url. [code: 10015]

### Steps to reproduce

Attempt to create R2 bucket with API

### References

https://github.com/cloudflare/wrangler2/issues/1654

Cyb3r-Jak3:
## Description

Should fix the R2 bucket creation API request

See https://github.com/cloudflare/wrangler2/pull/1653
Fixes #1033 

Unit tests pass

## Types of changes

## Checklist:

Should wait for wrangler to be merged to confirm that this is the correct fix.

### DIFF
--- a/r2_bucket_test.go
+++ b/r2_bucket_test.go
@@ -15,7 +15,7 @@ func TestR2_CreateBucket(t *testing.T) {
 	setup()
 	defer teardown()
 
-	mux.HandleFunc(fmt.Sprintf("/accounts/%s/r2/buckets/%s", testAccountID, testBucketName), func(w http.ResponseWriter, r *http.Request) {
+	mux.HandleFunc(fmt.Sprintf("/accounts/%s/r2/buckets", testAccountID), func(w http.ResponseWriter, r *http.Request) {
 		assert.Equal(t, http.MethodPost, r.Method, "Expected method 'POST', got %s", r.Method)
 		w.Header().Set("content-type", "application/json")
 		fmt.Fprintf(w, `{
@@ -26,17 +26,17 @@ func TestR2_CreateBucket(t *testing.T) {
 }`)
 	})
 
-	err := client.CreateR2Bucket(context.Background(), AccountIdentifier(""), "")
+	err := client.CreateR2Bucket(context.Background(), AccountIdentifier(""), CreateR2BucketParameters{})
 	if assert.Error(t, err) {
 		assert.Equal(t, ErrMissingAccountID, err)
 	}
 
-	err = client.CreateR2Bucket(context.Background(), AccountIdentifier(testAccountID), "")
+	err = client.CreateR2Bucket(context.Background(), AccountIdentifier(testAccountID), CreateR2BucketParameters{})
 	if assert.Error(t, err) {
 		assert.Equal(t, ErrMissingBucketName, err)
 	}
 
-	err = client.CreateR2Bucket(context.Background(), AccountIdentifier(testAccountID), "example-bucket")
+	err = client.CreateR2Bucket(context.Background(), AccountIdentifier(testAccountID), CreateR2BucketParameters{testBucketName})
 	assert.NoError(t, err)
 }
 
--- a/r2_bucket_test.go
+++ b/r2_bucket_test.go
@@ -31,12 +31,12 @@ func TestR2_CreateBucket(t *testing.T) {
 		assert.Equal(t, ErrMissingAccountID, err)
 	}
 
-	err = client.CreateR2Bucket(context.Background(), AccountIdentifier(testAccountID), CreateR2BucketParameters{})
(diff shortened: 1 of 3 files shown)


Cyb3r-Jak3:
Also, do wrangler API calls count as public documentation?
There is
https://github.com/cloudflare/wrangler2/blob/7ae059b3dcdd9dce5f03110d8ff670022b8ccf02/packages/wrangler/src/r2.ts#L17-L24
which doesn't exist in the API docs

jacobbednarz:
thanks, i'll with the R2 folks internally and see what we can do here.

> Also, do wrangler API calls count as public documentation?

afraid not; the service team only maintains the API/developer docs. the wrangler/workers folks are in a bit of privileged position where they _usually_ own the APIs they are using so are across the upcoming changes.

sodabrew:
Was the PR you just got merged 3 days ago in #1028 already broken by an API that was removed without notice?

Cyb3r-Jak3:
The API worked when I tested it. The back ended seemed to have changed without noticed as it broke, creating buckets from wrangler as well.

sodabrew:
I will raise an issue with enterprise support. Pulling a public API that is actively in use is ridiculous.

It's still in the public docs: https://api.cloudflare.com/#r2-bucket-create-bucket
image

I will note that this API was mis-documented anyways. The documentation says to POST, but wrangler2 was using PUT. As a general rule, one should not POST to a non-existent resource to create it (as the documentation says), the correct HTTP verb for this case is PUT (as wrangler2 was doing). A more common approach is to POST to the collection URL, e.g. `/accounts/<accountid>/r2/buckets` which is what the two fix-it PRs are doing.

I will say again, though, under no circumstances should an in-use API have been yanked from production without notice.

jacobbednarz:
this was somewhat unintentional and the change is being rolled back so this isn't needed. thanks anyway!

sodabrew:
Ok, that's good. But also the discrepancy I observed re: PUT vs POST is worth sorting out. The docs say POST, but wrangler2 is doing a PUT. The better verb here is indeed PUT as it implies creation. Here's a good write up of the differences: https://restfulapi.net/rest-put-vs-post/

Cyb3r-Jak3:
@jacobbednarz This should be reopened. The API is changing to `POST /r2/buckets` with a JSON body of `{"name": "<bucket>"}`.

jacobbednarz:
I'll confirm with the team as they were looking into with issues when using tokens last I checked and were backing out the change until they could dig in further.

jacobbednarz:
there is still no ETA on the `POST` equivalent. i'll keep my ear to ground internally and see if we can get it added when it lands instead of having a PR stagnate.

r2_bucket_test.go
```diff
@@ -26,17 +26,17 @@ func TestR2_CreateBucket(t *testing.T) {
...
 	}
 
-	err = client.CreateR2Bucket(context.Background(), AccountIdentifier(testAccountID), "")
+	err = client.CreateR2Bucket(context.Background(), AccountIdentifier(testAccountID), CreateR2BucketParameters{})
 	if assert.Error(t, err) {
 		assert.Equal(t, ErrMissingBucketName, err)
 	}
 
-	err = client.CreateR2Bucket(context.Background(), AccountIdentifier(testAccountID), "example-bucket")
+	err = client.CreateR2Bucket(context.Background(), AccountIdentifier(testAccountID), CreateR2BucketParameters{testBucketName})
```
jacobbednarz:
```suggestion
	err = client.CreateR2Bucket(context.Background(), AccountIdentifier(testAccountID), CreateR2BucketParameters{Name: testBucketName})
```

r2_bucket_test.go
```diff
@@ -26,17 +26,17 @@ func TestR2_CreateBucket(t *testing.T) {
...
 	})
 
-	err := client.CreateR2Bucket(context.Background(), AccountIdentifier(""), "")
+	err := client.CreateR2Bucket(context.Background(), AccountIdentifier(""), CreateR2BucketParameters{})
 	if assert.Error(t, err) {
 		assert.Equal(t, ErrMissingAccountID, err)
 	}
 
-	err = client.CreateR2Bucket(context.Background(), AccountIdentifier(testAccountID), "")
+	err = client.CreateR2Bucket(context.Background(), AccountIdentifier(testAccountID), CreateR2BucketParameters{})
```
jacobbednarz:
```suggestion
	err = client.CreateR2Bucket(context.Background(), AccountIdentifier(testAccountID), CreateR2BucketParameters{Name: ""})
```

jacobbednarz:
the documentation for fixing the POST payloads is coming in the next couple of days. we'll revive this as it will be the way forward.

PR #1040, opened 2022-08-13
Add support for multipart downloads for module-format Worker scripts

sodabrew:
## Description

Since #1010 allows uploading a Worker script in module format, calling DownloadWorker on the same script will return a multipart/form-data response.

### Asks

1. _Update: still need this; for the moment I'm assuming that any multipart worker is in module format._ Cloudflare API for download a worker should add a Content-Type header to each multipart/form-data section to specify whether the script in that section is application/javascript or application/javascript+module.
2. _Update: took this approach, please review!_ Guidance on a deep change: `makeRequest...` et al should return a content-type header not just the body.
3. _Update: didn't like this approach:_ ~Guidance on a workaround: `makeRequest...` et al continue to just return the response body, so consumers assume the format, or detect the format with a heuristic. For example, `DownloadWorker` can check the first two bytes and if they are `--` then assume the rest of that line is the multipart boundary string and parse the body accordingly. This is so gross.~

### Details

This approach digs into the `makeRequest...` family (`makeRequestWithAuthTypeAndHeaders` is the full implementation) and adds support to return the first body of a multipart response. `makeRequest`, et al, does not return the mime-type of the response. Consuming functions mostly assume it is application/json, except DownloadWorker which implicitly assumes it's application/javascript.

While I've solved the initial problem with #1010 that you can upload a module-format worker but if you try to download it, the DownloadWorker method return raw multipart/form-data, I'm stuck on the next problem: the Module true/false status is _implied_ by whether the download script API returns a javascript content-type or a multipart content-type.

There's no further clue because the Cloudflare API for downloading a worker only gives a Content-Disposition header on its response for a module worker:
```
Content-Type: multipart/form-data; boundary=mimeboundary

--mimeboundary
Content-Disposition: form-data; name="worker.js"

export default { etc etc etc }
--mimeboundary
```
vs. non-module format:
```
Content-Type: application/json

addEventListener( etc etc )
```

This isn't consistent with **uploading** a module worker, where the Content-Type of the mime-part must be set to `application/javascript+module` like this:
```
--mimeboundary
Content-Disposition: form-data; name="worker.js"
Content-Type: application/javascript+module

export default { etc etc etc }
--mimeboundary
```

Sidenote: why was the method `UploadWorkerWithBindings` added? The bindings are in a param structure, isn't the point to add new params without creating a new method? And in fact if you call UploadWorker with Module-format worker, it shunts to UploadWorkerWithBindings! The salient difference is whether the upload is in multipart format or a direct script. The code path I can see that isn't in multipart format is a non-module worker without bindings (the original product, so probably also the most common case). _Per manual testing, the Cloudflare API will accept a multipart/form-data with a plain script as its body -- that is, **UploadWorker** is sufficient for all cases and **UploadWorkerWithBindings** is duplicative._

_**Works**_ but creates an awkward roundtrip problem in the current iteration:

1. `UploadWorker` with Module: true because it's a module script.
```
        scriptFile, err := os.ReadFile("cool-script.js")
        scriptText := string(scriptFile)

        UploadWorker(ctx,
                &cloudflare.WorkerRequestParams{
                        ScriptName: "modtest",
                },
                &cloudflare.WorkerScriptParams{
                        Script: scriptText,
                        Module: true,
                },
        )
```
4. `DownloadWorker` with the script name and returns the script body. But what format is it?
```
        scriptText = DownloadWorker(ctx,
                &cloudflare.WorkerRequestParams{
                        ScriptName: "modtest",
                },
        )
```
5. `UploadWorker` ... with Module true or false? There's no indication from DownloadWorker which to use.
```
        strings.Replace(scriptText, "foo", "bar", -1)

        UploadWorker(ctx,
                &cloudflare.WorkerRequestParams{
                        ScriptName: "modtest",
                },
                &cloudflare.WorkerScriptParams{
                        Script: scriptText,
                        Module: true,   <---- ?????????????????                },
        )
```

## Types of changes

## Checklist:

### DIFF
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -176,21 +176,42 @@ func (api *API) makeRequestContextWithHeaders(ctx context.Context, method, uri s
 	return api.makeRequestWithAuthTypeAndHeaders(ctx, method, uri, params, api.authType, headers)
 }
 
-// Deprecated: Use `makeRequestContextWithHeaders` instead.
-//nolint:unused
-func (api *API) makeRequestWithHeaders(method, uri string, params interface{}, headers http.Header) ([]byte, error) {
-	return api.makeRequestWithAuthTypeAndHeaders(context.Background(), method, uri, params, api.authType, headers)
-}
-
 func (api *API) makeRequestWithAuthType(ctx context.Context, method, uri string, params interface{}, authType int) ([]byte, error) {
 	return api.makeRequestWithAuthTypeAndHeaders(ctx, method, uri, params, authType, nil)
 }
 
+// Struct for makeRequestWithAuthTypeAndHeaders's return value.
+// Wrapper methods can return just the Body if that's all the caller needs.
+//
+// Allows callers to see the Content-Type header and HTTP Status codes in case
+// there is additional meaning for particular API calls. This is especially
+// useful if the response is a multipart/form-data because the boundary string
+// is stored as a parameter of the content type header. Caller can parse using mime/multipart.
+type APIResponse struct {
+	Body        []byte
+	ContentType string
+	StatusCode  int
+}
+
 func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, uri string, params interface{}, authType int, headers http.Header) ([]byte, error) {
+	res, err := api.makeRequestWithAuthTypeAndHeadersComplete(ctx, method, uri, params, api.authType, headers)
+	if err != nil {
+		return nil, err
+	}
+	return res.Body, err
+}
+
+// Use this method if an API response can have different Content-Type headers and different body formats.
+func (api *API) makeRequestContextWithHeadersComplete(ctx context.Context, method, uri string, params interface{}, headers http.Header) (*APIResponse, error) {
+	return api.makeRequestWithAuthTypeAndHeadersComplete(ctx, method, uri, params, api.authType, headers)
+}
+
+func (api *API) makeRequestWithAuthTypeAndHeadersComplete(ctx context.Context, method, uri string, params interface{}, authType int, headers http.Header) (*APIResponse, error) {
 	var err error
 	var resp *http.Response
 	var respErr error
 	var respBody []byte
+
 	for i := 0; i <= api.retryPolicy.MaxRetries; i++ {
 		var reqBody io.Reader
 		if params != nil {
@@ -277,13 +298,15 @@ func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, u
 			}
 			break
 		}
-	}
+	} // for loop
+
+	// still had an error after all retries
 	if respErr != nil {
 		return nil, respErr
 	}
 
 	if api.Debug {
-		fmt.Printf("cloudflare-go [DEBUG] RESPONSE StatusCode:%d Body:%#v RayID:%s\n", resp.StatusCode, string(respBody), resp.Header.Get("cf-ray"))
+		fmt.Printf("cloudflare-go [DEBUG] RESPONSE StatusCode:%d RayID:%s ContentType:%s Body:%#v\n", resp.StatusCode, resp.Header.Get("cf-ray"), resp.Header.Get("content-type"), string(respBody))
 	}
 
 	if resp.StatusCode >= http.StatusBadRequest {
@@ -341,7 +364,11 @@ func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, u
 		}
 	}
 
-	return respBody, nil
+	return &APIResponse{
+		Body:        respBody,
+		StatusCode:  resp.StatusCode,
+		ContentType: resp.Header.Get("content-type"),
+	}, nil
 }
 
 // request makes a HTTP request to the given API endpoint, returning the raw
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -180,17 +180,16 @@ func (api *API) makeRequestWithAuthType(ctx context.Context, method, uri string,
 	return api.makeRequestWithAuthTypeAndHeaders(ctx, method, uri, params, authType, nil)
 }
 
-// Struct for makeRequestWithAuthTypeAndHeaders's return value.
-// Wrapper methods can return just the Body if that's all the caller needs.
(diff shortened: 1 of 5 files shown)


cloudflare.go
```diff
@@ -265,6 +267,17 @@ func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, u
 				api.logger.Printf("Error performing request: %s %s : %s \n", method, uri, respErr.Error())
 			}
 			continue
+		}
+
+		// Successful request, let's read the body and return the first part
```
sodabrew: Need to fix this comment, as the code is returning the _last_ part if there are multiple parts.

cloudflare.go
```diff
@@ -273,13 +286,14 @@ func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, u
 			}
 			break
 		}
-	}
+	} // for loop
```
sodabrew: I added this comment to help understand where in the control flow the reader block was taking place. I actually think the body reading block should move out on control level -- that is, a successful API response breaks the for loop, then read the body.

workers_test.go
```diff
@@ -179,15 +179,38 @@ const (
...
+    return fetch(request)
+    }
+}
+--workermoduledownloadresponse--
+`
 )
 
 var (
 	successResponse               = Response{Success: true, Errors: []ResponseInfo{}, Messages: []ResponseInfo{}}
-	workerScript                  = "addEventListener('fetch', event => {\n    event.passThroughOnException()\nevent.respondWith(handleRequest(event.request))\n})\n\nasync function handleRequest(request) {\n    return fetch(request)\n}"
```
sodabrew, line 186: Way easier to read using backticks than embedded newlines, IMO. No functional change here.

workers_test.go
```diff
@@ -384,13 +407,33 @@ func TestWorkers_DownloadWorkerWithNameErrorsWithoutAccountId(t *testing.T) {
...
+}
+
 func TestWorkers_ListWorkerScripts(t *testing.T) {
 	setup(UsingAccount("foo"))
 	defer teardown()
 
 	mux.HandleFunc("/accounts/foo/workers/scripts", func(w http.ResponseWriter, r *http.Request) {
 		assert.Equal(t, http.MethodGet, r.Method, "Expected method 'GET', got %s", r.Method)
-		w.Header().Set("content-type", "application-json")
+		w.Header().Set("content-type", "application/json")
```
sodabrew: `makeRequestWithAuthTypeAndHeaders` doesn't look at the content-type of the response, so it never flagged the incorrect "`application-json`"[sic] content-type before. I think it would be helpful to be stricter about only parsing JSON responses if the response content-type is set appropriately.

sodabrew:
@jacobbednarz thanks for quickly merging the two test fixits that I pulled out of this PR into #1048 and #1049. I wrote a massive update to the OP with examples and problems of implied content-type. Looking forward to your comments!

cloudflare.go
```diff
@@ -341,6 +339,32 @@ func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, u
...
+		}
+	} else {
+		respBody, err = ioutil.ReadAll(resp.Body)
+		defer resp.Body.Close()
+		if err != nil {
+			return nil, fmt.Errorf("could not read response body: %w", err)
+		}
+	}
+
+	if api.Debug {
```
jacobbednarz: this probably shouldn't move as the debug is intended to also allow seeing non-200 HTTP responses to help problem diagnosis.

sodabrew: Moved this back where it was and moved the multipart parsing to DownloadWorker itself. Check out this new approach and let me know what you think!

sodabrew: ps ... what's going on in cloudflare_experimental.go? It looks like a subtly different API client implementation and is only used in the Zone code.

jacobbednarz:
this PR has become quite significantly larger than the intended scope and i'm afraid, i won't be merging it as is with such a large refactor of the core methods. leading into your next question, the experimental client (see https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md) will focus on these larger, sweeping improvements for the next major release. we'll be intentionally shaking away alot of the older library functionality that we have now which would be a better place for type of changes you're proposing here. i wouldn't recommend opening PRs for it at this stage, but, just be aware that it exists.

for now, if you want this functionality, we need to do it within the confines of the library API as it stands todya.

sodabrew:
Ok, thanks for that feedback. I've adjusted the scope of changes to be a lot narrower, please take another look and let me know.

jacobbednarz:
i've gone ahead and updated `APIResponse` to be more `http.Response` like as that will be the direction for the experimental client when we start porting functionality to it. this looks fine in the tests but do you want to confirm locally that this meets what you're after here before we merge?

cloudflare.go
```diff
@@ -176,21 +176,41 @@ func (api *API) makeRequestContextWithHeaders(ctx context.Context, method, uri s
...
 func (api *API) makeRequestWithAuthTypeAndHeaders(ctx context.Context, method, uri string, params interface{}, authType int, headers http.Header) ([]byte, error) {
+	res, err := api.makeRequestWithAuthTypeAndHeadersComplete(ctx, method, uri, params, api.authType, headers)
+	if err != nil {
+		return nil, err
+	}
+	return res.Body, err
+}
+
+// Use this method if an API response can have different Content-Type headers and different body formats.
+func (api *API) makeRequestContextWithHeadersComplete(ctx context.Context, method, uri string, params interface{}, headers http.Header) (*APIResponse, error) {
```
jacobbednarz, line 204: i don't love the name here however, i suspect this will be cleaned up in the experimental client so i'm not too phase but will want to keep it's usage _extremely_ limited until then.

sodabrew:
Looks great! I've rebased and squashed down to three commits: the core change I arrived at, your edits thereupon, and worker test script whitespace changes. You might need to repush your last commit with your own signing key (I see an "Unverified" note) -- or I could squash your final commit into mine with a Co-authored-by?

jacobbednarz:
done! thanks for the PR and getting it over the line. i've made some notes internally around the issues here with headers and the request methods not exposing everything which hopefully we can address nicely in the experimental client later on.

PR #1027, opened 2022-08-08
Add worker account settings

Cyb3r-Jak3:
## Description

Adds support for workers account settings: https://api.cloudflare.com/#worker-account-settings-properties

## Has your change been tested?

Created units tests and they pass

## Types of changes

What sort of change does your code introduce/modify?

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.
- [x] This change is using publicly documented (api.cloudflare.com or developers.cloudflare.com) and stable APIs.

PR #1056, opened 2022-08-19
First pass at attempting to flesh out account-scoped worker APIs.

MrJoy:
This is an attempt at fleshing this library out so that the Worker-related APIs are able to use the `/accounts/:account_id/workers/` endpoint namespace.

## Description

I wanted to get this PR opened quickly so I could get feedback on whether my approach is harmonious with the goals of the maintainers (and how to correct, if not) -- before I went too far down the wrong path.  I expect there's still quite a bit to do before it's ready for merging.  Feedback and guidance would be appreciated.

I've run, and updated as appropriate, the regression tests.  I have not yet tried using the library directly.

## Types of changes

## Checklist:

### DIFF
--- a/.changelog/1056.txt
+++ b/.changelog/1056.txt
@@ -1,0 +1,3 @@
+```release-note:
+Expand Worker methods to support `/accounts/:account_id/workers` API endpoints.
+```
--- a/workers.go
+++ b/workers.go
@@ -18,6 +18,7 @@ import (
 
 // WorkerRequestParams provides parameters for worker requests for both enterprise and standard requests.
 type WorkerRequestParams struct {
+	AccountID  string
 	ZoneID     string
 	ScriptName string
 }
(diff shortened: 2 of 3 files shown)


.changelog/1056.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:
```
jacobbednarz, line 1: the release note needs a type such as `enhancement`, `bug`, `note`. see https://github.com/cloudflare/cloudflare-go/blob/master/docs/changelog-process.md

jacobbednarz:
in general, the Workers methods need quite an overhaul. they need to:

- remove a bunch of old functionality (determining whether to use zone/account level based on the presence of a script name is one big one)
- adhere to https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md. see #1027 for prior art.
- update method names to better reflect their purpose
- remove reliance on helpers like `UsingAccount` that rely on global scope

it is quite an undertaking and the Workers team have some notes for an approach internally that they may be tackling. i'm unsure how much of this will be duplicated effort but if we tick off those issues, we can always trim the to do list internally if you decide to pursue them.

MrJoy:
I'd prefer to focus this PR on the scope necessary to move [this issue](https://github.com/cloudflare/terraform-provider-cloudflare/issues/1646) forward, if possible.  Any work done to that end should be additive towards larger refactoring goals by virtue of implementing the new codepaths that will be needed under a new interface.  I'm somewhat less comfortable with removing APIs or making breaking changes since I'm an outsider to this project and don't have full context for the impact of those changes.

Would you be amenable to a PR that focuses just on enabling use of the new endpoints without making breaking/structural changes for now?

jacobbednarz:
you'd actually be making more work for yourself by trying to work around it and introduce a weird API for consumers. 

if you start with the method signature changes, you'll see what I mean. the older stuff disappears as we have account/zone `ResourceContainer` and the parameters become explicit instead of implied by internals. 

all good if you're not comfortable making the changes but at this point I'd rather we address these few things to clarify the Workers API as it's very hard to use and maintain at the moment.

MrJoy:
I'll see what I can do later this week..

PR #1068, opened 2022-08-29
Expose raw api call with context and raw context with headers

23doors:
## Description

Simple change, .Raw() is already exposed so it makes sense to also expose version with context and with headers. Possibly also full context+headers+complete version should be included.

## Types of changes

- New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/.changelog/1068.txt
+++ b/.changelog/1068.txt
@@ -1,3 +1,3 @@
 ```release-note:enhancement
-api: expose .RawContext and .RawContextWithHeaders methods
+api: addded context and headers to Raw method
 ```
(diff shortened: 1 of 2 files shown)


23doors:
I think that due to relatively niche usage of .Raw, this change doesn't necessarily warrant a changelog entry. Added a temporary one though.

jacobbednarz:
given the low usage of this method, i would probably just update the method signature to include `ctx` and `headers` since i would imagine most would already want this. what do you think?

23doors:
I agree and having just 1 signature for this is enough as long as it's not as limiting as the current one. For some it may be a backwards incompatible change though. But I guess it doesnt really matter for that kind of method. 

Updated PR

PR #1072, opened 2022-08-30
Fix nil dereference panic after 8563c7a

NicholasAsimov:
## Description

Fixes nil panic regression introduced by 8563c7a2, if `respErr != nil` then body is most likely `nil`. 

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x73bc4d]
goroutine 1 [running]:
github.com/cloudflare/cloudflare-go.(*API).makeRequestWithAuthTypeAndHeaders(0xc000328000, {0xc33d38, 0xc00027bcc0}, {0xb42aa6, 0xd}, {0xc0004a83c0, 0x17}, {0x0, 0x0}, 0x1, ...)
/go/pkg/mod/github.com/cloudflare/cloudflare-go@v0.48.0/cloudflare.go:254 +0xa2d
github.com/cloudflare/cloudflare-go.(*API).makeRequestWithAuthType(...)
/go/pkg/mod/github.com/cloudflare/cloudflare-go@v0.48.0/cloudflare.go:186
github.com/cloudflare/cloudflare-go.(*API).makeRequestContext(0xb45349, {0xc33d38, 0xc00027bcc0}, {0xb42aa6, 0x54a}, {0xc0004a83c0, 0x0}, {0x0, 0x0})
/go/pkg/mod/github.com/cloudflare/cloudflare-go@v0.48.0/cloudflare.go:172 +0x57
github.com/cloudflare/cloudflare-go.(*API).Accounts(0xb43d08, {0xc33d38, 0xc00027bcc0}, {{0x0, 0x1}, {0xc000333600, 0x54a}})
/go/pkg/mod/github.com/cloudflare/cloudflare-go@v0.48.0/accounts.go:59 +0xe7
main.main()
```

N/A

## Types of changes

## Checklist:

### DIFF
--- a/.changelog/1072.txt
+++ b/.changelog/1072.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cloudflare: fix nil dereference error in makeRequestWithAuthTypeAndHeaders
+```
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -271,7 +271,7 @@ func (api *API) makeRequestWithAuthTypeAndHeadersComplete(ctx context.Context, m
 		// retry if the server is rate limiting us or if it failed
 		// assumes server operations are rolled back on failure
 		if respErr != nil || resp.StatusCode == http.StatusTooManyRequests || resp.StatusCode >= 500 {
-			if resp.StatusCode == http.StatusTooManyRequests {
+			if resp != nil && resp.StatusCode == http.StatusTooManyRequests {
 				respErr = errors.New("exceeded available rate limit retries")
 			}
 


jacobbednarz:
are you able to provide a reproduction case where this would first fail that we can put into a regression test?

NicholasAsimov:
I'm afraid not, it only happens on production environment which would be a bit complicated. Anyways you can emulate any network error that will return `resp [nil], respErr [error]`

jacobbednarz:
the testing i've tried with this using a variety of 4xx and 5xx being returned but none have actually surfaced the panic. in all cases, `resp` is non-nil (like the check above it) so i'm unsure how this is popping up at the moment.

jacobbednarz:
there is no harm adding the additional safety here however, i'm curious to find out where this is popping up to handle it specifically instead of as a generic error.

amoiseiev:
@jacobbednarz I think the fix is appropriate and it's a good practice to be defensive, especially when you interact with remote APIs. 

Having that said, after upgrading to 0.49 I realized it's likely related to context timeouts. Now, in the same environment, I'm getting: "operation aborted during backoff: context deadline exceeded", which may explain why you were getting resp with live API interaction but likely not hitting (or setting) context timeout.

jacobbednarz:
i don't think anyone was disagreeing the change was a good idea. my reservation was with adding code and hoping it helped instead of having a solid example of it addressing the problem.

the context deadline is a good lead; thanks. have you got an example to share that is demonstrating the issue? if not, i'll have to have a play and see if i can pull something together for it.

amoiseiev:
I created https://github.com/cloudflare/cloudflare-go/pull/1080 with a potential improvement.

Basically, resp is "nil" on context timeout in go http client. The way to reproduce is to impose an artificially short timeout to force the code going through this path (example in the pull request), crashes with 0.48 and generates a bit ambiguous error in 0.49, so exiting earlier will likely cover well for such issue and follow the spirit of less processing /. less bugs approach.

PR #1080, opened 2022-09-04
Additional context deadline handling in makeRequestWithAuthTypeAndHeadersComplete

amoiseiev:
Currently, the request retry / backoff logic is being applied to requests that reached context timeout. While it still works as we account for context deadline, it may be better to exit immediately. It also indirectly helps to safeguard from regressions like the ones introduced in v0.48.0 (fixed by https://github.com/cloudflare/cloudflare-go/pull/1072) as the response body is nil when go http client exits on context deadline (less processing/potentially less bugs).

Now, the error will just say that we reached the deadline for that specific request and doesn't imply we applied backoff logic or retried the request. It doesn't take away from the ability to surface more specific errors.

Also updating TestContextTimeout to:

1) better readability
2) making use of an actual http endpoint and increasing timeout values to make sure we get to api.request function and not timing out on rate limiter
3) ability to run the test individually

Ran a request with short context timeout to reproduce the issue, confirmed it the expected result on timeout is long enough for a request to finish

```
	ctx, cncl := context.WithTimeout(context.Background(), time.Millisecond*time.Duration(50))
	defer cncl()

	// Getting existing A records from the zone
	_, err = api.DNSRecords(ctx, "<zone id>", cloudflare.DNSRecord{
		Type: "A",
		Name: "domain",
	})
```

## Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X ] My code follows the code style of this project.
- [ X] All new and existing tests passed.

### DIFF
--- a/.changelog/1080.txt
+++ b/.changelog/1080.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-cloudflare: exiting closer to the source on context timeouts to better defend from potential edge cases
+cloudflare: exiting closer to the source on context timeouts to improve error messaging and better defend from potential edge cases
 ```
--- a/.changelog/1080.txt
+++ b/.changelog/1080.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cloudflare: exiting closer to the source on context timeouts to better defend from potential edge cases
+```
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -285,6 +285,8 @@ func (api *API) makeRequestWithAuthTypeAndHeadersComplete(ctx context.Context, m
 
 				api.logger.Printf("Request: %s %s got an error response %d: %s\n", method, uri, resp.StatusCode,
 					strings.Replace(strings.Replace(string(respBody), "\n", "", -1), "\t", "", -1))
+			} else if errors.Is(respErr, context.DeadlineExceeded) {
+				return nil, respErr
 			} else {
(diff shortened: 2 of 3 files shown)


jacobbednarz:
thanks for digging into this one. let's add this in a test with the assertion that we get the correct exception. once we have that, we're good to go.

amoiseiev:
> thanks for digging into this one. let's add this in a test with the assertion that we get the correct exception. once we have that, we're good to go.

So, we actually have a test called "TestContextTimeout" that is checking for error type, the reason it didn't catch v0.48 regression is because when we set context timeout to "0", the deadline gets reached on https://github.com/cloudflare/cloudflare-go/pull/1080/files#diff-432b285f81cb984bfd1ce00fa760be757c96784135a1b9ecd49d3b6c9a26522bR250 , even before we make a request. 

The error that's currently being returned (before this change) is also of the right type, it's just extra wrapped in "backoff" error. Once the change is merged, it won't be wrapped by the "backoff" error but with retain context.DeadlineExceeded var type. I'm not sure it's worth looking for an unwrapped error in the test but happy to make the change if you have strong opinion on it.

I tweaked the test to spin up an actual endpoint to timeout on and increased the timeout value so we can get to api.request block, which makes it more realistic practical scenario. As a bonus, it's easier to understand what the test is doing and which URL it's calling. It also makes it run individually just fine now.

jacobbednarz:
i think that updated test is great, thank you.

PR #1082, opened 2022-09-06
fix: make request with auth type

23doors:
## Description

Simple fix for regression introduced in https://github.com/cloudflare/cloudflare-go/commit/873a91b42be9d4efa192605043b91b8cb9666fb2

Currently all makeRequestWithAuthType calls (so all origin_ca.go calls) are failing cause it's using api.authType rather then overridden AuthUserService auth type in this case.

## Types of changes

## Checklist:

### DIFF
--- a/.changelog/1082.txt
+++ b/.changelog/1082.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+origin certificate: Fix API auth type used
+```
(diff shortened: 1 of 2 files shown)


jacobbednarz:
thanks @23doors! 💯

PR #1095, opened 2022-09-29
ACCT-4459: Support for domain scoped roles

ianmobbs:
As we [begin to GA domain-scoped roles](https://blog.cloudflare.com/domain-scoped-roles-ga/), we'd like to add support for these new controls to the Cloudflare Go SDK.

## Description

The primary changes here are:
- Update AccountMember to contain both AccountRoles and Policies, our new struct for representing access control
- Add structs for Policies and all required sub-structs

Yes, we have extended existing tests, added new tests, and written a verification script

n/a

## Types of changes

## Checklist:


account_members.go
```diff
@@ -81,18 +83,73 @@ func (api *API) AccountMembers(ctx context.Context, accountID string, pageOpts P
...
+	return api.CreateAccountMemberInternal(ctx, accountID, invite)
+}
+
+// CreateAccountMember invites a new member to join an account with roles.
+// The member will be placed into "pending" status and receive an email confirmation.
+// NOTE: If you are currently enrolled in Domain Scoped Roles, your roles will be converted to policies
+// upon member invitation. We recommend upgrading to CreateAccountMemberWithPolicies to use policies.
+//
+// API reference: https://api.cloudflare.com/#account-members-add-member
+func (api *API) CreateAccountMember(ctx context.Context, accountID string, emailAddress string, roles []string) (AccountMember, error) {
```
jacobbednarz: if we're overhauling the invitations, let's update `CreateAccountMember` method signature to match the [experimental client](https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md). this will also allow you to cut down on some of the boilerplate code we're repeating and putting into `CreateAccountMemberInternal`.

(incomplete but hopefully enough to get you started example)

```go
type CreateAccountMemberParams struct {
	Email    string   `json:"email"`
	Roles    []string `json:"roles,omitempty"`
	Policies []Policy `json:"policies,omitempty"`
	Status   string   `json:"status,omitempty"`
}

func (api *API) CreateAccountMember(ctx context.Context, rc *ResourceContainer, params CreateAccountMemberParams) (AccountMember, error) {
	invite := AccountMemberInvitation{
		Email: params.Email,
	}

	if len(params.Roles) > 0 {
		invite.Roles = params.Roles
	}

	// ... etc
}
```

account_members.go
```diff
@@ -81,18 +83,73 @@ func (api *API) AccountMembers(ctx context.Context, accountID string, pageOpts P
...
+		Policies: policies,
+		Status:   "",
+	}
+	return api.CreateAccountMemberInternal(ctx, accountID, invite)
+}
+
+// CreateAccountMemberInternal allows you to provide a raw AccountMemberInvitation to be processed
+// and contains the logic for other CreateAccountMember* methods
+func (api *API) CreateAccountMemberInternal(ctx context.Context, accountID string, invite AccountMemberInvitation) (AccountMember, error) {
+	// make sure we have account
```
jacobbednarz: let's remove comments like this one. i find it more of a smell since if we need these style of comments, the code isn't clear enough. a good overview of where we aim to go with code comments in this repository: https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/

errors.go
```diff
@@ -17,6 +17,7 @@ const (
 	errUnmarshalErrorBody                     = "error unmarshalling the JSON response error body"
 	errRequestNotSuccessful                   = "error reported by API"
 	errMissingAccountID                       = "required missing account ID"
+	errMissingMemberRolesOrPolicies           = "account member must be created with roles or policies (not both)"
```
jacobbednarz: these are fine to put in the `account_members` files too if they are specific to just that resource.

resource_group.go
```diff
@@ -0,0 +1,53 @@
+package cloudflare
```
jacobbednarz, line 1: just a note that we already have a resource container concept in this library that may be a little confusing here (even though i know IAM had resource groups first 😛 ). i wonder if there is a way we can better differentiate the two as unrelated? in the future versions of the library (see https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md) we'll have a clearer line as the IAM component will likely live in it's own subpackage/namespace as opposed to the top level concepts so this may not be a thing to solve now.

account_members.go
ianmobbs: Good call, thanks! Can you confirm that this is the right way to go about things?
- Re-write `CreateAccountMember` to match the experimental client
- Remove the `CreateAccountMemberInternal` method
- Re-factor utility methods (`CreateAccountMemberWithStatus`, `CreateAccountMemberWithRoles`, etc) to just call `CreateAccountMember` - the alternative I supposed is just removing the utility methods?

Here's my current `CreateAccountMember` impl, let me know if it looks okay and the above is correct and I'll push :) 
```go
func (api *API) CreateAccountMember(ctx context.Context, rc *ResourceContainer, params CreateAccountMemberParams) (AccountMember, error) {
	if rc.Level != AccountRouteLevel {
		return AccountMember{}, fmt.Errorf(errInvalidResourceContainerAccess, rc.Level)
	}

	invite := AccountMemberInvitation{
		Email:  params.EmailAddress,
		Status: params.Status,
	}

	roles := []AccountRole{}
	for i := 0; i < len(params.Roles); i++ {
		roles = append(roles, AccountRole{ID: invite.Roles[i]})
	}
	err := validateRolesAndPolicies(roles, params.Policies)
	if err != nil {
		return AccountMember{}, err
	}

	if params.Roles != nil {
		invite.Roles = params.Roles
	} else if params.Policies != nil {
		invite.Policies = params.Policies
	}

	uri := fmt.Sprintf("/accounts/%s/members", rc.Identifier)
	res, err := api.makeRequestContext(ctx, http.MethodPost, uri, invite)
	if err != nil {
		return AccountMember{}, err
	}

	var accountMemberListResponse AccountMemberDetailResponse
	err = json.Unmarshal(res, &accountMemberListResponse)
	if err != nil {
		return AccountMember{}, fmt.Errorf("%s: %w", errUnmarshalError, err)
	}

	return accountMemberListResponse.Result, nil
}
```

resource_group.go
ianmobbs: Oof, yeah - I think let's work on this when re-packing the IAM components. I'm not sure we'll be moving away from the `ResourceGroup` terminology any time soon (and had a discussion today about extending it, actually) but will brainstorm ways to reduce confusion in this package.

account_members.go
jacobbednarz: looks spot on! nice one.

account_members_test.go
```diff
@@ -215,8 +252,12 @@ func TestCreateAccountMemberWithStatus(t *testing.T) {
 	}
 
 	mux.HandleFunc("/accounts/01a7362d577a6c3019a474fd6f485823/members", handler)
+	accountResource := &ResourceContainer{
```
jacobbednarz: from the code docs - https://github.com/cloudflare/cloudflare-go/blob/master/resource.go#L12-L14

```
// ResourceContainer defines an API resource you wish to target. Should not be
// used directly, use `UserIdentifier`, `ZoneIdentifier` and `AccountIdentifier`
// instead.
```

while you can do this, i'd recommend using the helpers instead incase we expand this struct. it's also a bit nicer to read 😄 `AccountIdentifier("01a7362d577a6c3019a474fd6f485823")`

PR #1111, opened 2022-10-18
DLP-631 Add support for DLP Endpoints

GreenStage:
Adds support for Zero Trust DLP endpoints

## Description

We are adding DLP endpoints support to cloudflare-go and terraform, as we prepare to go live in production.

Manually, through a simple program running DLP endpoint calls against my account.

## Types of changes

- [ X ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ X ] My code follows the code style of this project.
- [ X ] My change requires a change to the documentation.
- [ X  ] I have updated the documentation accordingly.
- [ X ] I have added tests to cover my changes.
- [ X ] All new and existing tests passed.
- [ X ] This change is using publicly documented (api.cloudflare.com or developers.cloudflare.com) and stable APIs.  --> **PENDING RELEASE**

### DIFF
--- a/.changelog/1111.txt
+++ b/.changelog/1111.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+dlp: Adds support for DLP resources
+```
--- a/dlp_profile.go
+++ b/dlp_profile.go
@@ -24,7 +24,7 @@ type DLPEntry struct {
 	ID        string `json:"id,omitempty"`
 	Name      string `json:"name,omitempty"`
 	ProfileID string `json:"profile_id,omitempty"`
-	Enabled   bool   `json:"enabled,omitempty"`
+	Enabled   *bool  `json:"enabled,omitempty"`
 	Type      string `json:"type,omitempty"`
 
 	// The following fields are only present for custom entries.
--- a/dlp_profile.go
+++ b/dlp_profile.go
@@ -1,0 +1,194 @@
+package cloudflare
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+)
+
+// DLPPattern represents a DLP Pattern that matches an entry.
+type DLPPattern struct {
+	Regex      string `json:"regex,omitempty"`
+	Validation string `json:"validation,omitempty"`
+}
+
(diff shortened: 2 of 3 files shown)


jacobbednarz:
i went ahead and updated the methods to match the [`experimental`](https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md) approach for method signatures and conventions to keep it in line with where we are heading for these methods. hopefully, the simplification of these methods to be around the generic `DLPProfile` with the option to target custom or predefined types demonstrates the benefits of this type of approach.

i've also swapped `Enabled` to be a boolean pointer. we've had too many issues with Terraform to use a straight boolean so this allows us some flexibility when we get to that integration.

GreenStage:
Hi @jacobbednarz , thanks for helping out!
One small concern:
You can only create and delete **custom** profiles. So changing the signature of the CreateDLPCustomProfile/DeleteDLPCustomProfile to CreateDLPProfile/DeleteDLPProfile may promote confusion.

jacobbednarz:
it's definitely a concern but something we can rely on the API/docs to help guide (also, the `Delete` is custom only for now via the URL). do we currently have API responses in place to help with this?

the thing I'm aiming to avoid here is down the track if we allow predefined profiles modifiable/deletable and we have to change the method signatures again. it's a pattern that keeps popping up where we have initial constraints but eventually loosen them off and cause SDK users an annoying migration.

GreenStage:
Totally understand you.

As you said, both the create endpoint to delete endpoint urls are specifically under the /custom path.  The API does not have routes for DELETE /predefined/{id} or POST /predefined. This will result in 404s.

I agree with your concern for flexibility and future retro-compatibility, but we need to figure out a way to avoid confusion here

GreenStage:
the way i see it, we have 2 options:
1) add a if check on the create/delete function for "custom" type;
2) change the signatures to CreateDLPCustomProfile and DeleteDLPCustomProfile

jacobbednarz:
let's go with #1. we can have it throw an exception for requests that are not custom and keep that logic internal to the method. 

I'll push up a PR for this later today.

jacobbednarz:
have made this update for the `Create` in #1113. the `Delete` is already using `custom` in the path so would already only apply to those profiles.

PR #1117, opened 2022-10-22
chore: remove github.com/pkg/errors dependency

cornelk:
Follow up from https://github.com/cloudflare/cloudflare-go/pull/972


jacobbednarz:
thank you @cornelk 🏆

PR #1133, opened 2022-11-23
DS-12892: Add Workers Analytics Engine binding

jimhawkridge:
This is required by https://github.com/cloudflare/terraform-provider-cloudflare/pull/2051 and was manually tested as part of that.
I've also extended an existing unit test similar to what was done for the R2 binding.


jacobbednarz:
this looks good @jimhawkridge . if you can add the CHANGELOG, we can get this merged.

jacobbednarz:
thanking you!

PR #1136, opened 2022-11-30
Update for Functions GA

WalshyDev:
## Description
For [Functions GA](http://blog.cloudflare.com/pages-function-goes-ga/) we released a ton of features. This brings the SDK up to date with those.

**Docs will be released tomorrow (30th Nov)**

Pages project & deployment tests both pass.

## Types of changes

## Checklist:

- [-] This change is using publicly documented (api.cloudflare.com or developers.cloudflare.com) and stable APIs.

### DIFF
--- a/pages_project.go
+++ b/pages_project.go
@@ -67,18 +67,17 @@ type PagesProjectDeploymentConfigs struct {
 
 // PagesProjectDeploymentConfigEnvironment represents the configuration for preview or production deploys.
 type PagesProjectDeploymentConfigEnvironment struct {
-	EnvVars            map[string]PagesProjectDeploymentVar `json:"env_vars"`
-	CompatibilityDate  string                               `json:"compatibility_date,omitempty"`
-	CompatibilityFlags []string                             `json:"compatibility_flags,omitempty"`
-	KvNamespaces       NamespaceBindingMap                  `json:"kv_namespaces,omitempty"`
-	DoNamespaces       NamespaceBindingMap                  `json:"durable_object_namespaces,omitempty"`
-	D1Databases        D1BindingMap                         `json:"d1_databases,omitempty"`
-	R2Bindings         R2BindingMap                         `json:"r2_buckets,omitempty"`
-}
-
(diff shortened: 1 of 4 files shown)


pages_project.go
```diff
@@ -67,18 +67,17 @@ type PagesProjectDeploymentConfigs struct {
...
-	KvNamespaces       NamespaceBindingMap                  `json:"kv_namespaces,omitempty"`
-	DoNamespaces       NamespaceBindingMap                  `json:"durable_object_namespaces,omitempty"`
-	D1Databases        D1BindingMap                         `json:"d1_databases,omitempty"`
-	R2Bindings         R2BindingMap                         `json:"r2_buckets,omitempty"`
-}
-
-// PagesProjectDeploymentVar represents a deployment environment variable.
-type PagesProjectDeploymentVar struct {
-	Value string `json:"value"`
+	EnvVars                          EnvironmentVariableMap `json:"env_vars,omitempty"`
```
jacobbednarz, line 70: in a typed language, this is a bit of a code smell. instead of including the type (map) in the name, could we call it EnvVars that expects a `[]map[string]EnvironmentVariable`?

pages_project.go
```diff
@@ -67,18 +67,17 @@ type PagesProjectDeploymentConfigs struct {
...
+	KvNamespaces                     NamespaceBindingMap    `json:"kv_namespaces,omitempty"`
```
jacobbednarz, line 71: similar with these types. as we are breaking changes already, can we fix these up?

jacobbednarz (changes requested):

pages_project.go
WalshyDev: We use these types internally, so I'd personally like to keep this consistent. I can change if you really want though

jacobbednarz: alright, let's keep them for consistency but the only caveat is once we start looking at codegen'ing the library, this will be changed again to be the concrete type of `map[string]*EnvironmentVariable` so we'll be inconsistent there once that happens.

PR #1137, opened 2022-11-30
Modernise Worker methods and conventions

jacobbednarz:
For the 4.x release of the Terraform provider, we are removing the global `api.AccountID` that is configured on the client due to ongoing issues with it (see cloudflare/terraform-provider-cloudflare#1646). Workers are the last set of methods that rely on this behaviour 🌟

In doing this, we're also modernising the methods to match our future conventions and method signatures [outlined for the experimental client](https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md). 

This _should_ be pretty well 1:1 functionality wise as everything Workers related is now accessible on the account resource level instead of the older split between zone/account for non-ent and enterprise configurations. The other major change here is that now the various parts of the Workers ecosystem are in their own files so that we can start to assign specific ownership rules instead of lumping it all together.


workers_routes.go
```diff
@@ -0,0 +1,161 @@
...
+func (api *API) CreateWorkerRoute(ctx context.Context, rc *ResourceContainer, params CreateWorkerRouteParams) (WorkerRouteResponse, error) {
+	if rc.Level != ZoneRouteLevel {
+		return WorkerRouteResponse{}, fmt.Errorf(errInvalidResourceContainerAccess, ZoneRouteLevel)
+	}
+
+	if rc.Identifier == "" {
+		return WorkerRouteResponse{}, ErrMissingIdentifier
+	}
+
+	uri := fmt.Sprintf("/zones/%s/workers/routes", rc.Identifier)
```
jacobbednarz, line 40: worker routes still use the `/workers/routes` endpoint instead of the services route as that isn't documented. once we have that documented, we can swap it over.

PR #1147, opened 2022-12-15
Update api shield to allow empty payload

jafowler:
## Description

It has recently come to our attention that `auth_id_characteristics` can be empty when sent to the API. This change will fix the payload that goes to the API to send `{ "auth_id_characteristics" : [] }` when the payload is empty. 

I added an additional test case for sending an empty auth_id_characteristics payload.

## Types of changes

### DIFF
--- a/api_shield.go
+++ b/api_shield.go
@@ -74,7 +74,9 @@ func (api *API) UpdateAPIShieldConfiguration(ctx context.Context, rc *ResourceCo
 func validateAPIShieldConfigurationUpdateParams(params UpdateAPIShieldParams) UpdateAPIShieldParams {
 	// edge case for when we get an empty auth_id_characteristics. We want to send `{auth_id_characteristics: []}` to the API.
 	if len(params.AuthIdCharacteristics) <= 1 {
-		if params.AuthIdCharacteristics[0].Name == "" && params.AuthIdCharacteristics[0].Type == "" {
+		if len(params.AuthIdCharacteristics) == 0 {
+			params.AuthIdCharacteristics = make([]AuthIdCharacteristics, 0)
+		} else if params.AuthIdCharacteristics[0].Name == "" && params.AuthIdCharacteristics[0].Type == "" {
 			params.AuthIdCharacteristics = make([]AuthIdCharacteristics, 0)
 		}
 	}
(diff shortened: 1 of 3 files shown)


jacobbednarz:
you'll need to update your branch here (presumably, your fork). it's removing a bunch of stuff it shouldn't be.

jafowler:
> you'll need to update your branch here (presumably, your fork). it's removing a bunch of stuff it shouldn't be.

oh weird, let me check

jafowler:
fixed, sorry about that.

api_shield_test.go
```diff
@@ -82,3 +82,33 @@ func TestPutAPIShield(t *testing.T) {
 		assert.Equal(t, want, actual)
 	}
 }
+
+func TestPutAPIShieldNoAuthIDCharacteristics(t *testing.T) {
+	setup()
+	defer teardown()
+
+	// now lets do a PUT
```
jacobbednarz, line 90:
```suggestion
```

api_shield.go
```diff
@@ -68,3 +70,15 @@ func (api *API) UpdateAPIShieldConfiguration(ctx context.Context, rc *ResourceCo
 
 	return asResponse, nil
 }
+
+func validateAPIShieldConfigurationUpdateParams(params UpdateAPIShieldParams) UpdateAPIShieldParams {
+	// edge case for when we get an empty auth_id_characteristics. We want to send `{auth_id_characteristics: []}` to the API.
```
jacobbednarz, line 75:
```suggestion
```

api_shield.go
```diff
@@ -56,6 +56,8 @@ func (api *API) GetAPIShieldConfiguration(ctx context.Context, rc *ResourceConta
 func (api *API) UpdateAPIShieldConfiguration(ctx context.Context, rc *ResourceContainer, params UpdateAPIShieldParams) (Response, error) {
 	uri := fmt.Sprintf("/zones/%s/api_gateway/configuration", rc.Identifier)
 
+	params = validateAPIShieldConfigurationUpdateParams(params)
```
jacobbednarz, line 59: I don't know how I feel about this given that an empty array for a PUT request may remove anything existing. Is there a reason we can't rely on end users to explicitly set this instead?

api_shield.go
jacobbednarz: if this is important context, let's ensure it is in the git commit, not code comments - https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/

api_shield.go
jafowler: Fair, i think this same approach can be done in the provider.

PR #1151, opened 2022-12-22
Support for DNS comments and tags

Cyb3r-Jak3:
## Description

Adds support for the new DNS comments and tags. There is a breaking change because I updated DNSRecords to support the new listing pagination.

New test has been created and all tests pass

## Types of changes

## Checklist:

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -29,6 +29,8 @@ type DNSRecord struct {
 	Proxied    *bool       `json:"proxied,omitempty"`
 	Proxiable  bool        `json:"proxiable,omitempty"`
 	Locked     bool        `json:"locked,omitempty"`
+	Comment    string      `json:"comment,omitempty"`
+	Tags       []string    `json:"tags,omitempty"`
 }
 
 // DNSRecordResponse represents the response from the DNS endpoint.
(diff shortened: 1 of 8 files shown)


dns.go
```diff
@@ -104,30 +139,71 @@ func (api *API) DNSRecords(ctx context.Context, zoneID string, rr DNSRecord) ([]
 		v.Set("content", rr.Content)
 	}
 
+	if rr.Proxied != nil {
+		v.Set("proxied", strconv.FormatBool(*rr.Proxied))
```
jacobbednarz: we don't need all of these `v.Set` calls as we can instead, pass it to `buildURI` and it will use the struct field configuration to build this out.

dns.go
```diff
@@ -90,7 +125,7 @@ func (api *API) CreateDNSRecord(ctx context.Context, zoneID string, rr DNSRecord
 // This takes a DNSRecord to allow filtering of the results returned.
 //
 // API reference: https://api.cloudflare.com/#dns-records-for-a-zone-list-dns-records
-func (api *API) DNSRecords(ctx context.Context, zoneID string, rr DNSRecord) ([]DNSRecord, error) {
+func (api *API) DNSRecords(ctx context.Context, zoneID string, rr DNSRecord, params DNSListParameters) ([]DNSRecord, *ResultInfo, error) {
```
jacobbednarz: if we're introducing a breaking change, let's update the method signature to use the `ResourceContainer` and be renamed to `ListDNSRecords`.

Cyb3r-Jak3: Do you want me to do that for all the other methods as well?

jacobbednarz: sure, if you're up for it!

dns.go
Cyb3r-Jak3: This would require adding ``url:""`` to all the fields of `DNSRecord` as that is used for some fields when creating the URL. There is also the tag matching, which wouldn't fit inside the standard `buildURI` function.

jacobbednarz: > This would require adding url:"" to all the fields of DNSRecord as that is used for some fields when creating the URL.

that's totally fine given it's only used when building query parameters.

> There is also the tag matching, which wouldn't fit inside the standard buildURI function.

i haven't looked at the API docs but this should be fine to add to the buildURI method or pass in a struct that allows nesting, etc.

Cyb3r-Jak3: Docs sections

> DNS record tag value. For advanced queries, the following alternative  parameters are available: 'tag.\<tag-name\>.present',  'tag.\<tag-name\>.absent', 'tag.\<tag-name\>.exact',  'tag.\<tag-name\>.contains', 'tag.\<tag-name\>.startswith',  'tag.\<tag-name\>.endswith'. Just 'tag.\<tag-name\>' is  equivalent to 'tag.\<tag-name\>.exact'. Replace \<tag-name\>  with the name of the tag to query. Example:  /dns_records?tag.important.present&tag.team=DNS. Note that  'tag.\<tag-name\>.present' and 'tag.\<tag-name\>.absent' do not require a value; if one is given anyway it is ignored.

This means that you can have a URL parameter of `tag.example` or `tag.example.startswith` which is
https://github.com/cloudflare/cloudflare-go/pull/1151/files#diff-cb1b8a01bbc5f64cf06dce0e3f8021ad8289078b402dc620afa04780484d7cd5R152-R160
exists

jacobbednarz: for simplicity sake, let's do all except for the tags. the others should be simple query parameters and we can look at incorporating this pattern later if it becomes a thing in more places.

ghost:
Wow, thanks for this @Cyb3r-Jak3! Just noticed the merge conflict on #1158, turns out because you already added it! :tada:

PR #1156, opened 2022-12-23
firewall_rules/filter: reset fResponse on each loop to avoid residual data

jafowler:
## Description
We found an issue were data from the previous page could populate into the current page because we were not reseting the struct on each loop.

This is something i'd love to have a conversation about. We need / should look into how to properly test our paginated APIs

## Types of changes
bugfix

### DIFF
--- a/.changelog/1156.txt
+++ b/.changelog/1156.txt
@@ -1,0 +1,6 @@
+```release-note:bug
+firewall_rules: use empty reponse struct on each page call
+```
+```release-note:bug
+filter: use empty reponse struct on each page call
+```
--- a/.changelog/1156.txt
+++ b/.changelog/1156.txt
@@ -1,6 +1,31 @@
 ```release-note:bug
 firewall_rules: use empty reponse struct on each page call
 ```
+
 ```release-note:bug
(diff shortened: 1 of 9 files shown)


manatarms (approved):

jacobbednarz:
have you checked the other list methods? they all use basically the same layout and may be impacted in a similar way. 

in an ideal world, this should be a helper method or utilise Go generics from 1.19. 

I'll give this change a run through a couple of projects I have using the pagination conditions as this needs to account for automatic and manual pagination.

jafowler:
> have you checked the other list methods? they all use basically the same layout and may be impacted in a similar way.
> 
> in an ideal world, this should be a helper method or utilise Go generics from 1.19.
> 
> I'll give this change a run through a couple of projects I have using the pagination conditions as this needs to account for automatic and manual pagination.

There are a couple more places where this should be implemented. I have them stashed locally but i'll look into how to do this with generics to make things more ideal.

jacobbednarz:
we'll address the move to a generics based helper once Go 1.20 lands and we drop support for 1.17.

PR #1158, opened 2022-12-28
Add support for DNS record comments and tags

ghost:
## Description

This adds support for DNS record comment and tags.

Blog post: https://blog.cloudflare.com/dns-record-comments/
Documentation: https://developers.cloudflare.com/dns/manage-dns-records/reference/record-attributes/

Updated unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/dns.go
+++ b/dns.go
@@ -29,6 +29,8 @@ type DNSRecord struct {
 	Proxied    *bool       `json:"proxied,omitempty"`
 	Proxiable  bool        `json:"proxiable,omitempty"`
 	Locked     bool        `json:"locked,omitempty"`
+	Comment    *string     `json:"comment,omitempty"`
+	Tags       []string    `json:"tags,omitempty"`
 }
 
 // DNSRecordResponse represents the response from the DNS endpoint.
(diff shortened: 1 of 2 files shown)


ghost:
Closing in favor of #1151 which looks good and has already been merged!

favonia:
@janik-cloudflare I have a question about the new DNS record comment API: How should I _clear_ a comment or set the comment to empty without recreating the DNS record?

ghost:
Hi @favonia! Comments can be removed by setting them to `null` in a PUT or PATCH request.

favonia:
@janik-cloudflare Thanks. I wonder if it's still possible the change the design to treat the empty string as clearing comment for the following reasons:

1. The Go interface needs to distinguish the following three cases. Treating `""` as clearing has the benefit of reusing Go's built-in JSON encoder for the type `*string` (with `omitempty`).
   a. Keep the current comment: `nil` (which would then be omitted)
   b. Clear the current comment: `&""`
   c. Set a new comment: `&"New comment"`
2. The behavior does not seem to be documented on the API website yet and the API is relatively new. Few, if any, applications depend on this behavior.
3. If strict backward compatibility is needed, we can treat both `""` and `null` as clearing the comment.

PS: I know `&"New comment"` is not really a valid Go expression because `"New comment"` is a constant.

ghost:
@favonia I've discussed this with the team. We agree that the current solution isn't ideal (e.g., Golang makes it too difficult to distinguish between `null` and an absent key) but we'd prefer not to change the API design to one we find less intuitive in order to work around language limitations.

If a PATCH request with `{"comment": ""}` removes the comment, I would expect missing comments to always be represented as `""` instead of `null` (in other words, I would be surprised to see `{"comment": null}` in an API response after sending `{"comment": ""}`). However, when creating a new record without including the `comment` field at all, that would require us to send `{"comment": ""}` in API responses. Although Go treats `""` as a default value for strings, most languages either use `null` or don't have an implicit default at all, so seeing an empty string without ever having set the field would feel unintuitive to me.

PR #1160, opened 2022-12-29
feat(workers): allow enabling logpush on worker uploads

ndisidore:
This allows cloudflare-go users to turn on/off [logpush support for Cloudflare Workers](https://developers.cloudflare.com/workers/platform/logpush/).
The analogous functionality in Wrangler can be found here: https://github.com/cloudflare/wrangler2/blob/7169142171b1c9b4ff2f19b8b46871932ef7d10a/packages/wrangler/src/create-worker-upload-form.ts#L294

## Description

This adds a new option to `CreateWorkerParams` that allows control over the logpush setting. Note its a nullable bool: the 3 states are true - on / false - off / nil - unchanged. Again, this is analogous to wrangler with its `boolean | undefined` type.

A new test was created to test this. 

## Types of changes

## Checklist:

### DIFF
--- a/workers.go
+++ b/workers.go
@@ -28,6 +28,10 @@ type CreateWorkerParams struct {
 	// ES Module syntax script.
 	Module bool
 
+	// Logpush opts the worker into Workers Logpush logging. A nil value leaves the current setting unchanged.
+	//  https://developers.cloudflare.com/workers/platform/logpush/
+	Logpush *bool
+
 	// Bindings should be a map where the keys are the binding name, and the
 	// values are the binding content
 	Bindings map[string]WorkerBinding
@@ -220,7 +224,7 @@ func (api *API) UploadWorker(ctx context.Context, rc *ResourceContainer, params 
 		err         error
 	)
 
-	if params.Module || len(params.Bindings) > 0 {
+	if params.Module || params.Logpush != nil || len(params.Bindings) > 0 {
 		contentType, body, err = formatMultipartBody(params)
 		if err != nil {
 			return WorkerScriptResponse{}, err
@@ -257,8 +261,10 @@ func formatMultipartBody(params CreateWorkerParams) (string, []byte, error) {
 		BodyPart   string              `json:"body_part,omitempty"`
 		MainModule string              `json:"main_module,omitempty"`
 		Bindings   []workerBindingMeta `json:"bindings"`
+		Logpush    *bool               `json:"logpush,omitempty"`
 	}{
 		Bindings: make([]workerBindingMeta, 0, len(params.Bindings)),
+		Logpush:  params.Logpush,
 	}
 
 	if params.Module {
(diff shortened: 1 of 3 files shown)


ndisidore:
Re: `My code follows the code style of this project.` - it looks like there are a number of lint errors present, but none that this change set has introduced
```
cloudflare-go on  nathan/feat/workers-logpush via 🐹 v1.19.4
❯ golangci-lint --new-from-rev=HEAD~1 run ./..
```

workers_test.go
```diff
@@ -823,3 +826,37 @@ func TestUploadWorker_WithServiceBinding(t *testing.T) {
...
+		assert.NoError(t, err)
+
+		expected := true
+		assert.Equal(t, &expected, mpUpload.Logpush)
+
+		w.Header().Set("content-type", "application/json")
+		fmt.Fprint(w, uploadWorkerResponseData)
+	})
+	logpush := true
+	res, err := client.UploadWorker(context.Background(), AccountIdentifier(testAccountID), CreateWorkerParams{ScriptName: "foo", Script: workerScript, Logpush: &logpush})
```
jacobbednarz:
```suggestion
	res, err := client.UploadWorker(context.Background(), AccountIdentifier(testAccountID), CreateWorkerParams{ScriptName: "foo", Script: workerScript, Logpush: BoolPtr(true)})
```

we have [`BoolPtr`](https://github.com/cloudflare/cloudflare-go/blob/master/convert_types.go#L77) helper to simlify this sort of usage.

workers.go
```diff
@@ -220,7 +224,7 @@ func (api *API) UploadWorker(ctx context.Context, rc *ResourceContainer, params
 		err         error
 	)
 
-	if params.Module || len(params.Bindings) > 0 {
+	if params.Module || params.Logpush != nil || len(params.Bindings) > 0 {
```
jacobbednarz, line 227: could you expand on this logic for me? i'm not familiar enough with the usage of logpush jobs. are you unable to configure logpush for moduled based workers?

ndisidore: Sure! Logpush can be enabled for both module & service worker formats, but the setting is set inside the metadata field which is only sent as part of the multipart body format here.

zebp (approved):

workers.go
jacobbednarz: this logic just doesn't feel right to me. if i'm reading this logic correctly and someone omits `params.Logpush` for a module based worker or has bindings, it won't ever end up calling `formatMultipartBody`, correct?

i think want we actually want here is a similar to https://github.com/cloudflare/cloudflare-go/blob/c071c00470ddeabd8cd9df4576cab11022be4622/workers.go#L270-L273 in the `formatMultipartBody` method where if it is defined, we want to append it to the struct.

```go
if params.Logpush != nil {
    meta.Logpush = params.Logpush
}
```

or am i missing something here?

jacobbednarz: holiday brain and reading this as AND, not OR. this should be 👍

PR #1169, opened 2023-01-06
Allow for autopagination with custom page size

arturhoo:
Signed-off-by: Artur Rodrigues <artur.rodrigues@lacework.net>

On https://github.com/cloudflare/cloudflare-go/pull/1151, `ListDNSRecordsParams` was introduced, which allows for users of the library to "search" for a specific page. What is not possible, however, is to have auto-pagination (i.e.: fetch all results) with a customer page size.

This level of flexibility is interesting for those seeking to reduce the number of API calls made to Cloudflare if they're facing rate limits.

## Description

On https://github.com/cloudflare/cloudflare-go/pull/1151, `ListDNSRecordsParams` was introduced, which embeds the [`ResultInfo`](https://github.com/cloudflare/cloudflare-go/blob/63d7aa85c4216e423033ce2f3315a672b657afd6/cloudflare.go#L473-L482) struct.

This bring the library to the same level of flexibility as the [API itself](https://api.cloudflare.com/#dns-records-for-a-zone-list-dns-records). However, users opting for the auto-pagination feature are restricted to the default `PerPage` [value of 50](https://github.com/cloudflare/cloudflare-go/blob/master/dns.go#L173).

If users decide to pass in a custom `PerPage` value, auto-pagination is silently turned off, with just the first page being returned.

This PR simply changes the logic in ListDNSRecords to turn off `autopaginate` only if `Page` is passed in. With the current auto-pagination logic, a larger `PerPage` can be used with no side-effects. This seems sensible, as auto-pagination is an added feature of the library on top of the API.

Imho, adding some documentation to the behaviour (current and new) is probably a good idea.

Added new unit test for the pagination behaviour with a custom `PerPage` value.
I've also validated it manually using `flarectl`.

N/A

## Types of changes

## Checklist:

### DIFF
--- a/.changelog/1169.txt
+++ b/.changelog/1169.txt
@@ -1,0 +1,3 @@
+```release-note:breaking-change
+dns: auto-pagination with custom per_page value and no page
+```
--- a/dns.go
+++ b/dns.go
@@ -165,7 +165,7 @@ func (api *API) ListDNSRecords(ctx context.Context, rc *ResourceContainer, param
 	}
 
 	autoPaginate := true
-	if params.PerPage >= 1 || params.Page >= 1 {
+	if params.Page >= 1 {
 		autoPaginate = false
 	}
 
--- a/dns_test.go
+++ b/dns_test.go
@@ -153,7 +153,7 @@ func TestCreateDNSRecord(t *testing.T) {
 	assert.Equal(t, want, actual)
 }
 
-func TestDNSRecords(t *testing.T) {
+func TestListDNSRecords(t *testing.T) {
 	setup()
 	defer teardown()
 
@@ -242,7 +242,7 @@ func TestDNSRecords(t *testing.T) {
 	assert.Equal(t, want, actual)
 }
 
-func TestDNSRecordsSearch(t *testing.T) {
+func TestListDNSRecordsSearch(t *testing.T) {
 	setup()
 	defer teardown()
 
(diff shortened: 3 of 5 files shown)


jacobbednarz (changes requested):
thanks for the PR but we won't be accepting it as is. the library has the convention that if you provide any pagination options, we don't paginate. this gives us the best of both worlds where those who don't care for internal pagination can easily get all values but if someone needs to control pagination, they have full control over it. we won't be changing this in the foreseeable future. 

if the per page count is problematic, we evaluate that on an endpoint by endpoint basis. generally, that is 50 records however, for DNS we can have a higher threshold given the service defaults to a much higher value (such as 1000).

arturhoo:
Thanks for the thoughtful response.

In this case, perhaps the default should be changed from 50 in the library to 100, matching the default value of the API itself? https://api.cloudflare.com/#dns-records-for-a-zone-list-dns-records

> the library has the convention that if you provide any pagination options, we don't paginate. this gives us the best of both worlds where those who don't care for internal pagination can easily get all values but if someone needs to control pagination, they have full control over it.

The consistency bit makes a lot of sense! However, I think the best of both worlds is still achieved: those wanting the flexibility of controlling pagination themselves only have to set `Page`, while those who don't can continue to call the method without passing the attributes from `ResultInfo`. Perhaps an explicit auto-pagination attribute in the `ResultInfo` struct would make it better? Let me know if you'd be open to that. Thanks,

jacobbednarz:
> In this case, perhaps the default should be changed from 50 in the library to 100, matching the default value of the API itself? https://api.cloudflare.com/#dns-records-for-a-zone-list-dns-records

apologies if it wasn't clear; this is what i suggesting above but only for the DNS records endpoint. others we will keep at 50 since that it a sensible default for the majority of resources.

> I think the best of both worlds is still achieved: those wanting the flexibility of controlling pagination themselves only have to set `Page`, while those who don't can continue to call the method without passing the attributes from `ResultInfo`

this assumption doesn't match the feedback we've received. Go SDK users want the ability to control both aspects of the pagination which is why the or conditional disables automatic pagination if either are provided.

> Perhaps an explicit auto-pagination attribute in the ResultInfo struct would make it better? Let me know if you'd be open to that.

maybe at a later date, but not right now. we'll be looking to transition to Go generics in the future for a pagination helper which will trim down the boilerplate we reuse however, we have no solid plans to replace the convention for now.

arturhoo:
> apologies if it wasn't clear; this is what i suggesting above but only for the DNS records endpoint. others we will keep at 50 since that it a sensible default for the majority of resources.

I'll open a new PR with only this change!

PR #1170, opened 2023-01-08
dns: UpdateDNSRecord: clean up params and drop record lookup

favonia:
## Description

This PR avoids the extra DNS record lookup for `name` and `type`; these two fields are not needed for the PATCH method and thus the lookup is useless. It also removes many parameters that are probably read-only.

PS: This PR does not imply the `data` field should be there---I did not touch it because I do not understand it. Also, I assume `meta` is read-only, which could be wrong. @jacobbednarz Should `data` be there? How about `meta`?

Yes. I can confirm the real Cloudflare API does not need `name` or `type` to update a DNS record.

## Types of changes

Even though it changes `UpdateDNSRecordParams`, any existing application which puts non-zero values in those deleted fields should have failed to update the DNS record.

## Checklist:


favonia:
@jacobbednarz the coverage comparison unfortunately is not very helpful because the data for the master branch have not been consistently uploaded to Codecov.

.changelog/1170.txt
```diff
@@ -0,0 +1,7 @@
+```release-note:note
+dns: clean up UpdateDNSRecordParams and skip record lookup
```
jacobbednarz:
```suggestion
dns: remove additional lookup from `Update` operations when `Name` or `Type` was omitted
```

.changelog/1170.txt
```diff
@@ -0,0 +1,7 @@
...
+```
+
+```release-note:breaking-change
+dns: many fields have been removed from UpdateDNSRecordParams
```
jacobbednarz: we need to be specific here what is removed for those looking through the CHANGELOG

dns.go
```diff
@@ -231,22 +224,6 @@ func (api *API) UpdateDNSRecord(ctx context.Context, rc *ResourceContainer, para
...
-
-		if params.Name == "" {
-			params.Name = rec.Name
-		}
-		if params.Type == "" {
-			params.Type = rec.Type
-		}
-	}
-
 	uri := fmt.Sprintf("/zones/%s/dns_records/%s", rc.Identifier, params.ID)
```
jacobbednarz, line 238: we should add a check above this to ensure `params.ID` is provided otherwise the update will fail at the incorrect URL.

dns.go
```diff
@@ -231,22 +224,6 @@ func (api *API) UpdateDNSRecord(ctx context.Context, rc *ResourceContainer, para
 
 	params.Name = toUTS46ASCII(params.Name)
 
-	// Populate the record name from the existing one if the update didn't
-	// specify it.
-	if params.Name == "" || params.Type == "" {
```
jacobbednarz, line 236: i suspect this was from a time when updating DNS records was a PUT and these values were required to prevent broken resources if the `Name` or `Type` was omitted.

jacobbednarz:
> This PR does not imply the data field should be there---I did not touch it because I do not understand it. Also, I assume meta is read-only, which could be wrong. @jacobbednarz Should data be there? 

data is what is used for complex record types (CAA, SRV, etc) instead of `content`.

> How about meta?

`meta` is read only. largely to inform if the record is managed by an internal service instead of customers.

dns.go
favonia: Yes, I think so. (Too lazy to check the git history.)

.changelog/1170.txt
favonia: Done.

dns.go
favonia: It seems I should add `ErrMissingDNSRecordID` or something like that, but should I put it in `dns.go` or the global `errors.go`?

jacobbednarz: `dns.go` is fine. `errors.go` is used for library wide errors.

favonia: Done. The same checking was added to `GetDNSRecord`.

jacobbednarz (approved):

jacobbednarz:
thanks as always, @favonia

Issue #1163, opened 2023-01-04
Hide pagination info when auto-pagination is enabled


favonia:
### Current cloudflare-go version

0.54.0

### Description

As mentioned in https://github.com/cloudflare/cloudflare-go/issues/612#issuecomment-890554597, I wonder if it makes sense for the functions with auto-pagination to return a modified `ResultInfo` saying that "there's only one big single page" (that is, `Page` is `1`, `TotalPages` is `1`, and `Count` matches `Total`, etc,). 

### Use cases

It should be an invariant that `Count` always matches the number of returned entries, hiding the details of pagination when auto-pagination is enabled.

### Potential cloudflare-go usage

```go
rs, ri, err := ListDNSRecords(ctx, ...)

// len(rs) == ri.Count should always hold whenever err == nil
```

### References

This was discussed in #612. The discussion there was dismissed because we intended to enable auto-pagination unconditionally back then. However, the new API allows library users to turn off auto-pagination and thus the same design question re-appears.

PR #1171, opened 2023-01-09
dns: update ListDNSRecords default `per_page` attribute to 100 records

arturhoo:
Bump the `PerPage` value used when auto-pagination is performed on `ListDNSRecords` to 100, matching that of the API: https://api.cloudflare.com/#dns-records-for-a-zone-list-dns-records

## Description

As outlined in https://github.com/cloudflare/cloudflare-go/pull/1169, when no pagination options are passed in to `ListDNSRecords`, auto- pagination is performed. However, the current `per_page` value doesn't match that of the official API (50 vs 100). This commit makes both values match. 

A new test was implemented that more explicitly tests the auto pagination behaviour.

N/A

## Types of changes

None of the above. I'd classify this as an enhancement, per the discussion on https://github.com/cloudflare/cloudflare-go/pull/1169

## Checklist:

### DIFF
--- a/.changelog/1171.txt
+++ b/.changelog/1171.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+dns: Match ListDNSRecords per_page value to the API's (100)
+```
--- a/.changelog/1171.txt
+++ b/.changelog/1171.txt
@@ -1,3 +1,3 @@
 ```release-note:enhancement
-dns: Match ListDNSRecords per_page value to the API's (100)
+dns: update default `per_page` attribute to 100 records
 ```
--- a/dns.go
+++ b/dns.go
@@ -89,6 +89,9 @@ type DNSListResponse struct {
 	ResultInfo `json:"result_info"`
 }
 
+// listDNSRecordsDefaultPageSize represents the default per_page size of the API.
+var listDNSRecordsDefaultPageSize int = 100
+
 // nontransitionalLookup implements the nontransitional processing as specified in
 // Unicode Technical Standard 46 with almost all checkings off to maximize user freedom.
 var nontransitionalLookup = idna.New(
@@ -152,7 +155,10 @@ func (api *API) CreateDNSRecord(ctx context.Context, rc *ResourceContainer, para
 	return recordResp, nil
 }
 
-// ListDNSRecords returns a slice of DNS records for the given zone identifier.
+// ListDNSRecords returns a slice of DNS records for the given zone
+// identifier. If params doesn't include any pagination (ResultInfo)
+// options, auto pagination is performed with the default page size
+// of 100 records per request.
 //
 // API reference: https://api.cloudflare.com/#dns-records-for-a-zone-list-dns-records
 func (api *API) ListDNSRecords(ctx context.Context, rc *ResourceContainer, params ListDNSRecordsParams) ([]DNSRecord, *ResultInfo, error) {
@@ -170,7 +176,7 @@ func (api *API) ListDNSRecords(ctx context.Context, rc *ResourceContainer, param
 	}
 
 	if params.PerPage < 1 {
-		params.PerPage = 50
+		params.PerPage = listDNSRecordsDefaultPageSize
 	}
 
 	if params.Page < 1 {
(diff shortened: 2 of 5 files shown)


dns.go
```diff
@@ -89,6 +89,9 @@ type DNSListResponse struct {
 	ResultInfo `json:"result_info"`
 }
 
+// listDNSRecordsDefaultPageSize represents the default per_page size of the API.
+var listDNSRecordsDefaultPageSize int = 100
```
favonia, line 93:
```suggestion
const listDNSRecordsDefaultPageSize int = 100
```

favonia: Ah, sorry, I did not realize you intentionally made it a variable so that the testing can be done without generating >100 entries.

dns_test.go
```diff
@@ -341,6 +341,39 @@ func TestDNSRecordsSearch(t *testing.T) {
...
+
+		var response string
+		switch page {
+		case "1":
+			response = loadFixture("dns", "list_page_1")
+			page1Called = true
+		case "2":
+			response = loadFixture("dns", "list_page_2")
+			page2Called = true
+		}
```
favonia, line 367: Should this have a `default` case that errs?

dns.go
```diff
@@ -152,7 +155,10 @@ func (api *API) CreateDNSRecord(ctx context.Context, rc *ResourceContainer, para
 	return recordResp, nil
 }
 
-// ListDNSRecords returns a slice of DNS records for the given zone identifier.
+// ListDNSRecords returns a slice of DNS records for the given zone
+// identifier. If params doesn't include any pagination (ResultInfo)
+// options, auto pagination is performed with the default page size
+// of 100 records per request.
```
jacobbednarz: given this is a library convention (and default) we can omit this

```suggestion
// ListDNSRecords returns a slice of DNS records for the given zone
// identifier. 
```

.changelog/1171.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+dns: Match ListDNSRecords per_page value to the API's (100)
```
jacobbednarz:
```suggestion
dns: update default `per_page` attribute to 100 records
```

dns.go
```diff
@@ -152,7 +155,8 @@ func (api *API) CreateDNSRecord(ctx context.Context, rc *ResourceContainer, para
 	return recordResp, nil
 }
 
-// ListDNSRecords returns a slice of DNS records for the given zone identifier.
+// ListDNSRecords returns a slice of DNS records for the given zone
+// identifier.
```
favonia:
```suggestion
// ListDNSRecords returns a slice of DNS records for the given zone identifier.
```
(Reducing diff noise.)

favonia:
@jacobbednarz @arturhoo I wonder how you feel about #1163? It you agree with my proposal, it seems we can address it in this PR as well. More specifically, we can add more assertions at the end of the new `TestListDNSRecordsPagination`.

arturhoo:
Thanks for taking the time to review this PR @favonia 

> @jacobbednarz @arturhoo I wonder how you feel about https://github.com/cloudflare/cloudflare-go/issues/1163? It you agree with my proposal, it seems we can address it in this PR as well. More specifically, we can add more assertions at the end of the new TestListDNSRecordsPagination.

I don't have a strong opinion. As I outlined in https://github.com/cloudflare/cloudflare-go/pull/1169, I'd be more in favor of an explicit auto-pagination param inside `ListDNSRecordsParams` (or another struct that is embedded in multiple other param structs with a better name than `ResultInfo`).

In any case, it seems this PR here is a good step forward despite the direction we end up going?

favonia:
> In any case, it seems this PR here is a good step forward despite the direction we end up going?

It is, but it would have saved me from creating another PR. :laughing:

arturhoo:
Feel free to assign that one to me if it ends up being accepted 😁

fwiw, `ResultInfo` being returned when auto-pagination is enabled also caused some initial confusion in https://github.com/kubernetes-sigs/external-dns/pull/3288#discussion_r1063548215

jacobbednarz:
> @jacobbednarz @arturhoo I wonder how you feel about #1163? It you agree with my proposal, it seems we can address it in this PR as well

let's not muddy the waters here and keep them separate. they are different concerns and need separate conversations. I'm not rushing to change that return value as we have some work coming up to revamp the internals where we can cover that off. 

> with a better name than ResultInfo

this has been a pet peeve of mine for a while and is addressed in the experimental client. read it as `PaginationOptions` for now 😉

PR #1173, opened 2023-01-10
dns: clean up ListDNSRecordsParams and actually support tags for searching

favonia:
## Description

This is the sequel to #1170 to clean up `ListDNSRecordsParams`.

My (free) Cloudflare plan cannot test the new features `tag-match` and `tag` enabled by this PR. Otherwise, this PR passes the minimum test case, which barely tests the tags beyond mocking.

@jacobbednarz @janik-cloudflare **Help is needed to confirm the "real" behavior of tags.**

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)\
  Previously, `tags` was listed as a parameter but the URL parameter name was `tags`, which should have been `tag`.
- [x] New feature (non-breaking change which adds functionality)\
  The parameter `tag-match` is added.
- [x] Breaking change (fix or feature that would cause existing functionality to change)\
  All the fields not documented in the public API are removed except `Priority`.

## Checklist:

### DIFF
--- a/.changelog/1173.txt
+++ b/.changelog/1173.txt
@@ -1,0 +1,11 @@
+```release-note:bug
+dns: the field `Tags` in `ListDNSRecordsParams` was not correctly serialized into URL queries
+```
+
+```release-note:enhancement
+dns: the URL parameter `tag-match` for listing DNS records is now supported as the field `TagMatch` in `ListDNSRecordsParams`
+```
+
+```release-note:breaking-change
+dns: the fields `CreatedOn` and `ModifiedOn` are removed from `ListDNSRecordsParams`
+```
--- a/dns.go
+++ b/dns.go
@@ -47,18 +47,17 @@ const (
 )
 
 type ListDNSRecordsParams struct {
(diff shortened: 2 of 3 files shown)


jacobbednarz:
there are two different uses for tags; with and without predicates.

the first, without, is the simple one where we can do things like `/dns_records?tag=foo:bar&tag=team:DNS` which searches for an exact match or `/dns_records?tag=foo&tag=dns` which only searches the presence of those keys without considering the value. this should work as today,  example: https://go.dev/play/p/uLmPhZ3SWoA

the second and more difficult one is for predicates (like `/dns_records?tag.present=important&tag.exact=team:DNS`). i can't see a sensible way of achieving this with `go-querystring` since `map[string]string` ends up pretty muddled when it is encoded.

favonia:
@jacobbednarz Got it. Thanks. Then I think this PR correctly implements the basic mode (without predicates).

jacobbednarz:
👍 that's totally fine. the non-predicates are shorthand so we can extend the more advance operations later in if needed.

PR #1176, opened 2023-01-11
Add queue binding type for workers scripts

jbw1991:
## Description

Adds the Queue binding type. Workers scripts use this binding to write messages to a Queue.

* Added new unit test
* Manually tested alongside terraform provider changes here: https://github.com/cloudflare/terraform-provider-cloudflare/pull/2134

## Types of changes

## Checklist:

### DIFF
--- a/workers_bindings.go
+++ b/workers_bindings.go
@@ -39,6 +39,8 @@ const (
 	WorkerR2BucketBindingType WorkerBindingType = "r2_bucket"
 	// WorkerAnalyticsEngineBindingType is the type for Analytics Engine dataset bindings.
 	WorkerAnalyticsEngineBindingType WorkerBindingType = "analytics_engine"
+	// WorkerQueueBindingType is the type for queue bindings.
+	WorkerQueueBindingType WorkerBindingType = "queue"
 )
 
 type ListWorkerBindingsParams struct {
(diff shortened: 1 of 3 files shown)


jbw1991:
Changelog entry added!

As far as the API docs go, the workers script upload endpoint is documented (https://api.cloudflare.com/#worker-script-upload-worker) but it looks like it does not specifically cover each individual binding type.

I'm not sure if this repo contains documentation for the various workers binding types, but LMK if so and I can add a description.

workers_bindings.go
```diff
@@ -309,6 +311,34 @@ func (b WorkerAnalyticsEngineBinding) serialize(bindingName string) (workerBindi
 	}, nil, nil
 }
 
+// WorkerQueueBinding is a binding to a Workers Queue.
+//
+// https://developers.cloudflare.com/workers/platform/bindings/#queue-bindings
+type WorkerQueueBinding struct {
```
mtlemilio, line 317: How are we differentiating between a `producer` and `consumer` binding? https://developers.cloudflare.com/queues/configuration/

Should we add an optional sub-resource to allow folks to configure the consumer? Something like https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloudfunctions_function#event_trigger

jbw1991: This PR is only for the `producer` side of Queues config. I'll tackle all the consumer stuff separately, but yeah we are going to need to define a new type of sub-resource for consumers. Basically I was thinking that a script could have two sub resources: `queue_bindings` and `queue_consumers` (names could be changed...) that would handle each type.

ruimarinho: @jbw1991 are you planning on tackling queue consumers anytime soon?

PR #1185, opened 2023-01-25
Make Customer Endpoint optional in IPSEC tunnel config

jkunder:
Verified tunnel setup without specifying Customer endpoint IP in local setup

## Description
In Magic Transit IPSEC Tunnel config API, the customer endpoint IP parameter is optional.

Tested this with a private image. And verified that the tunnel setup is successful without specifying an IP address

## Types of changes

-  Enhancement : which provides flexibility during tunnel creation.


jacobbednarz:
thanks @jkunder 🥇

PR #1188, opened 2023-01-27
UpdateQueue function now matches API docs and allows a Queue's name t…

jbw1991:
## Description

The existing UpdateQueue implementation was not quite correct. The API endpoint only allows for updating a Queue's name. The existing name is passed in the URL, while the new name is passed as a parameter in the body. This PR updates the function to match the API expectations.

* Updated unit test
* Tested manually with local terraform provider changes to update a Queue's name

## Types of changes

## Checklist:

### DIFF
--- a/queue.go
+++ b/queue.go
@@ -78,7 +78,8 @@ type QueueConsumerResponse struct {
 }
 
 type UpdateQueueParams struct {
-	Name string `json:"queue_name,omitempty"`
+	Name        string `json:"-"`
+	UpdatedName string `json:"queue_name,omitempty"`
 }
 
 type ListQueueConsumersParams struct {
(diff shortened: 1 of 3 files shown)


queue.go
```diff
@@ -228,7 +221,7 @@ func (api *API) GetQueue(ctx context.Context, rc *ResourceContainer, queueName s
 // UpdateQueue updates a queue.
 //
 // API reference: https://api.cloudflare.com/#queue-update-queue
-func (api *API) UpdateQueue(ctx context.Context, rc *ResourceContainer, params UpdateQueueParams) (Queue, error) {
+func (api *API) UpdateQueue(ctx context.Context, rc *ResourceContainer, queueName string, params UpdateQueueParams) (Queue, error) {
```
jacobbednarz: the three parameter method signatures were intentional for the next stage of the SDKs. you can read more about it at https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md

jbw1991: I see. So how should I pass both parameters (existing queue name, and new queue name) into the function?
I could change `UpdateQueueParams` to something like:
```
type UpdateQueueParams struct {
        ExistingName string `json:"-"`
	UpdatedName string `json:"queue_name,omitempty"`
}
```

Then to make the API call:
```
	uri := fmt.Sprintf("/accounts/%s/workers/queues/%s", rc.Identifier, params.ExistingName)
	res, err := api.makeRequestContext(ctx, http.MethodPut, uri, params)
```

Would that be okay?

jacobbednarz: yep, that seems fine. I would go with `Name` and `UpdatedName` personally but either way. 

we generally avoid this sort of scenario because we don't make the URL resource identifier modifiable. it remains the same for the life of the resource. if we want to find by an attribute like name, it is a query parameter on a list endpoint.

jbw1991: 👍 Thanks, switched over to use the three parameter pattern.

jacobbednarz:
thanks!

PR #1196, opened 2023-02-01
dns: always send tags to allow clearing

jacobbednarz:


favonia:
@jacobbednarz Hi, I'm concerned about this PR. Does this mean there's no way to keep the existing tags (without re-specifying them) using the current Go API? Could we have `*[]string` or something like that to make it possible to keep existing tags?

favonia:
@jacobbednarz I hope it's still possible to keep existing tags with a zero value, and I am concerned that this PR will break that. Unfortunately, my plan does not allow me to test it. I wonder if I can use a test account to verify the current behavior of the API server and report a bug if the behavior is undesirable?

PR #1232, opened 2023-03-09
addressing: add support for Address Map resource

hunts:
## Description

This pull request adds support for the public `Address Map` apis.

About `Address Map`: https://developers.cloudflare.com/byoip/proxy-ips/
The API docs: https://api.cloudflare.com/#ip-address-management-address-maps-properties

## Types of changes

## Checklist:


addressing_address_map.go
```diff
@@ -0,0 +1,202 @@
...
+	"fmt"
+	"net/http"
+	"time"
+)
+
+// AddressMap contains information about an address map.
+type AddressMap struct {
+	ID           string                 `json:"id"`
+	Description  string                 `json:"description"`
+	DefaultSNI   string                 `json:"default_sni"`
```
aeriff:
```suggestion
	DefaultSNI   *string                 `json:"default_sni"`
```

Same here, this field can be null.

addressing_address_map.go
```diff
@@ -0,0 +1,202 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"time"
+)
+
+// AddressMap contains information about an address map.
+type AddressMap struct {
+	ID           string                 `json:"id"`
+	Description  string                 `json:"description"`
```
aeriff:
```suggestion
	Description  *string                 `json:"description"`
```
Sorry that the docs are not explicit about this (we need to fix it), but this field can be null.

addressing_address_map.go
```diff
@@ -0,0 +1,202 @@
...
+	DefaultSNI   string                 `json:"default_sni"`
+	Enabled      bool                   `json:"enabled"`
+	CanDelete    bool                   `json:"can_delete"`
+	CanModifyIPs bool                   `json:"can_modify_ips"`
+	Memberships  []AddressMapMembership `json:"memberships"`
+	IPs          []AddressMapIP         `json:"ips"`
+	CreatedAt    *time.Time             `json:"created_at"`
```
aeriff: The timestamp fields on all Address Maps endpoints are always non-null, but I am not sure if this is convention in cloudflare-go so I leave it up to you.

hunts: There are mixed usages of pointers and values for time fields in this repo. I'll make the change if the api does not return null.

jacobbednarz: `time.Time` usage is definitely varied (sorry!). i'm working on documenting type conventions better as people are commonly tripping up on uninitialised values and zero values with this library and i think i can do a better job of explaining where the traps are.

we have [builtin in helpers for type conversion](https://github.com/cloudflare/cloudflare-go/blob/master/convert_types.go) so this should be easy for most to adopt.

jacobbednarz: similarly, i recommend all booleans are boolean pointers so avoid some issues with Terraform.

addressing_address_map.go
```diff
@@ -0,0 +1,202 @@
...
+type AddressMapPatch struct {
+	Description *string `json:"description,omitempty"`
+	Enabled     *bool   `json:"enabled,omitempty"`
+	DefaultSNI  *string `json:"default_sni,omitempty"`
+}
+
+// ListAddressMaps lists all address maps owned by the account.
+//
+// API reference: https://developers.cloudflare.com/api/operations/ip-address-management-address-maps-list-address-maps
+func (api *API) ListAddressMaps(ctx context.Context, accountID string) ([]AddressMap, error) {
```
jacobbednarz: i've outlined some conventions for new functionality at https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md. in essence:

- all methods have max 3 parameters
- the third parameter is always a struct (or a single ID for `Get` methods)
- the second parameter is a `ResourceContainer`
- where possible, we stick to `Get`, `List`, `Update`, `Create` and `Delete` method prefixes for consistency
- `List` operations automatically paginate all results

you can check out https://github.com/cloudflare/cloudflare-go/blob/master/tunnel.go for prior art if you're wondering how it all hangs together.

hunts: I've made some changes according to these experimental items. The list api itself in this case does not support paging. Do I need to do anything with it?

addressing_address_map.go
hunts: Are you suggesting make all time and boolean fields pointers? And can you please elaborate what would be the issues Terraform run into with non-null fields? 

> `i recommend all booleans are boolean pointers so avoid some issues with Terraform`

hunts:
The `Coverage` action seems to be broken. @jacobbednarz . Is it a known issue?

jacobbednarz:
I'll have to have a look into it as to where the comparison is happening but that's fine to ignore it for now.

addressing_address_map.go
jacobbednarz: that would be my suggestion, yes. on the terraform side, there needs to be a way to determine if a field is uninitialised, false or true and the only reliable way of doing that is with pointer fields.

addressing_address_map.go
jacobbednarz: as long as the signature matches, that's fine. we have plenty of APIs that don't paginate but need to look the same.

addressing_address_map.go
jacobbednarz: i've gone ahead and pushed up some changes to address this

jacobbednarz:
thanks for this one @hunts 🍰

hunts:
Awesome! Thanks for the help @jacobbednarz on releasing this feature.

PR #1240, opened 2023-03-22
fix(json): fix invalid field name json mapping for workers subdomain

infosec-au:
## Description

We were unable to use the function inside this Golang API to retrieve the accounts worker subdomain. This was because the API response is different to what the code is expecting. The JSON field name from the API response is `subdomain` where as this code was expecting `name`. Changing the expected field name allows for us to successfully obtain the cloudflare subdomain name via this golang api.

Yes it has been tested.

API response:

```
{
  "result": {
    "subdomain": "fifty-lashes-blah"
  },
  "success": true,
  "errors": [],
  "messages": []
}
```

With my code change:

```
workers subdomain {Name:fifty-lashes-blah}
```

Before my code change:

```
workers subdomain {Name:}
```

Code:

```
	api, err := cloudflare.NewWithAPIToken(token.ApiToken)
	if err != nil {
		return nil, errors.Wrap(err, "Failed to create cloudflare API session")
	}

	accs, _, err := api.Accounts(ctx, cloudflare.AccountsListParams{})
	if err != nil {
		return nil, errors.Wrap(err, "Failed to get cloudflare accounts")
	}

	var accountIds []string

	for _, acc := range accs {
		accountIds = append(accountIds, acc.ID)
	}

	for _, accIdentifier := range accountIds {
		workersSubdomain, err := api.WorkersGetSubdomain(ctx, cloudflare.AccountIdentifier(accIdentifier))
		if err != nil {
			return nil, errors.Wrap(err, "Failed to list cloudflare worker domains")
		}
		fmt.Printf("workers subdomain %+v", workersSubdomain)
	}
```

## Types of changes

## Checklist:

- [] My change requires a change to the documentation.

### DIFF
--- a/workers_subdomain.go
+++ b/workers_subdomain.go
@@ -8,7 +8,7 @@ import (
 )
 
 type WorkersSubdomain struct {
-	Name string `json:"name,omitempty"`
+	Name string `json:"subdomain,omitempty"`
 }
 
 type WorkersSubdomainResponse struct {


jacobbednarz:
`name` is the publicly defined value in the API documentation so we'll need the team to update this accordingly.

see https://developers.cloudflare.com/api/operations/worker-subdomain-get-subdomain

infosec-au:
Hi @jacobbednarz -

Will the API response change so that it is `name` in the response as the API docs suggest? or will the docs change so that they reference `subdomain` instead?

Thank you for responding

jacobbednarz:
that's up to the team but this is technically a breaking change if the key is `subdomain` and not `name`.

jacobbednarz:
@infosec-au this has been updated in the API documentation. if you add a changelog entry to this PR, we can get it merged.

infosec-au:
Hi @jacobbednarz -

Can you please do this and get this merged, if you need it merged soon? I fixed this bug almost a year ago, and I don't have time right now to add a change log entry.

jacobbednarz:
afraid not. you've made changes to your fork's `master` which has branch protection enabled and is not editable by maintainers of the source repository. 

if you don't have time to add the file, that's fine we can close it out and some near future work can address it.

jacobbednarz:
closing PR due to age and lack of activity. if someone wants to revive, please do otherwise this will be covered in v2 automatically.

PR #1241, opened 2023-03-23
add access policies to access application

fuskiid:
An api call to https://api.cloudflare.com/client/v4/zones/{{zone}}/access/apps; in addition to what is documented at https://developers.cloudflare.com/api/operations/access-applications-list-access-applications will also return the application policies associated with each application.

This field is missing from both the cloudflare-go struct as well as the documentation as far as I can understand.

This can be demonstrated by making a call to the api and viewing the result return.

If this is an expected return value; could this be added to the developer api docs, and then incorporated into the the struct so the return may be usable?

Yes, it works as expected; added test from access_policy into access_application

## Types of changes

## Checklist:


jacobbednarz:
thanks for the PR however, Access policies are intentionally separate and available at https://github.com/cloudflare/cloudflare-go/blob/master/access_policy.go for management. we don't use the Access application for managing/reading the attached policies.

PR #1249, opened 2023-03-27
Enumerate service modes with explicit values

imdawon:
## Description

Our current type for the `Mode` field on the `ServiceModeV2` struct is a string. This allows for arbitrary strings, which are not supported service modes.

The `Mode` field now uses a new `ServiceMode` string type that has constant values that can be assigned. These values are all of the existing service modes.

The underlying `Mode` type has not changed, so we don't need to augment the current tests.
## Types of changes

## Checklist:

### DIFF
--- a/.changelog/1249.txt
+++ b/.changelog/1249.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+service_mode_v2: update `Mode` field to use new `ServiceMode` string type with explicit const service mode values
+```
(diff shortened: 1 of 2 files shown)


.changelog/1249.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+service_mode_v2: update `Mode` field to use new `ServiceMode` string type with explicit const service mode values
```
jacobbednarz:
```suggestion
devices_policy: update `Mode` field to use new `ServiceMode` string type with explicit const service mode values
```

PR #1264, opened 2023-04-07
Add support for pages deployment auto pagination

Cyb3r-Jak3:
## Description

Adds support for auto paginate for pages deployments. Support forcing deletion of a pages deployment.
Also removed an unused type and var.

Updated existing tests and they pass.

## Types of changes

## Checklist:


pages_deployment.go
```diff
@@ -217,7 +230,7 @@ func (api *API) GetPagesDeploymentLogs(ctx context.Context, rc *ResourceContaine
 // DeletePagesDeployment deletes a Pages deployment.
 //
 // API reference: https://api.cloudflare.com/#pages-deployment-delete-deployment
-func (api *API) DeletePagesDeployment(ctx context.Context, rc *ResourceContainer, projectName, deploymentID string) error {
+func (api *API) DeletePagesDeployment(ctx context.Context, rc *ResourceContainer, projectName, deploymentID string, params DeletePagesDeploymentParams) error {
```
jacobbednarz: this should have a maximum of three params - the last three should be condensed into one.

pages_project_test.go
```diff
@@ -583,9 +613,12 @@ func TestUpdatePagesProject(t *testing.T) {
 
 	mux.HandleFunc("/accounts/"+testAccountID+"/pages/projects/Test Pages Project", handler)
 
-	_, err := client.UpdatePagesProject(context.Background(), testAccountID, "Test Pages Project", *updateAttributes)
+	_, err := client.UpdatePagesProject(context.Background(), AccountIdentifier(""), *updateAttributes)
```
jacobbednarz, line 632: this test will be broken at the moment while i chat with @WalshyDev about what we are doing with friendly URL IDs since these shouldn't be able to change but i think we currently allow them to.

pagination.go
```diff
@@ -17,3 +17,10 @@ func (p ResultInfo) Next() ResultInfo {
 func (p ResultInfo) HasMorePages() bool {
 	return p.Page > 1 && p.Page < p.TotalPages
 }
+
+// DoneCount returns true for the last page and false otherwise.
+// Unlike Done(), this function uses the Total and PerPage fields to determine if
+// there are more pages to fetch.
+func (p ResultInfo) DoneCount() bool {
```
jacobbednarz: we shouldn't need this; if this is needed, we aren't sending the correct v4 response envelope and we should have the team address it.

Cyb3r-Jak3: I agree it shouldn't be needed. Unfortunately, there are other APIs besides deployments that don't return the full result_info information as well.

jacobbednarz: can you shoot me a list please? this is breaking an API contract if it is on a listing endpoint.

Cyb3r-Jak3: I've updated this to use the correct way. Once pages backend has support, then this will be all set.

Cyb3r-Jak3:
I don't know what the plan is for the project name test, I know that the API will reject it if you try. Also, `total_pages` is now in the API response for listing deployments.

PR #1270, opened 2023-04-18
PA-2043 - Add support for regional hostnames

aberglund-cf:
## Description

Adds methods which covers the Regional Hostname API calls documented here: https://developers.cloudflare.com/data-localization/regional-services/get-started/

## Has your change been tested?

- [x] Unit tests
- [x] Manual testing in staging environment 

## Types of changes

What sort of change does your code introduce/modify?

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] This change is using publicly documented (api.cloudflare.com or developers.cloudflare.com) and stable APIs.

[1]: https://help.github.com/articles/closing-issues-using-keywords/

PR #1276, opened 2023-05-04
WR-1279 add waiting room zone-level settings

piperswe:
## Description

This adds support for the new Waiting Room zone-level settings, which are used to enable search engine crawler bypass.

Tests were added, mirroring the other tests

## Types of changes

## Checklist:

### DIFF
--- a/waiting_room.go
+++ b/waiting_room.go
@@ -597,7 +597,7 @@ type UpdateWaitingRoomSettingsParams struct {
 }
 
 // UpdateWaitingRoomSettings lets you replace all zone-level Waiting Room settings. This is in contrast to
-// ChangeWaitingRoomSettings which lets you change individual settings.
+// PatchWaitingRoomSettings which lets you change individual settings.
 //
 // API reference: https://api.cloudflare.com/#waiting-room-update-zone-settings
 func (api *API) UpdateWaitingRoomSettings(ctx context.Context, rc *ResourceContainer, params UpdateWaitingRoomSettingsParams) (WaitingRoomSettings, error) {
--- a/waiting_room.go
+++ b/waiting_room.go
@@ -76,6 +76,12 @@ type WaitingRoomRule struct {
 	Enabled     *bool      `json:"enabled"`
 }
 
+// WaitingRoomSettings describes zone-level waiting room settings.
+type WaitingRoomSettings struct {
+	// Whether to allow verified search engine crawlers to bypass all waiting rooms on this zone
+	SearchEngineCrawlerBypass bool `json:"search_engine_crawler_bypass"`
+}
+
 // WaitingRoomPagePreviewURL describes a WaitingRoomPagePreviewURL object.
 type WaitingRoomPagePreviewURL struct {
 	PreviewURL string `json:"preview_url"`
@@ -96,6 +102,12 @@ type WaitingRoomDetailResponse struct {
 type WaitingRoomsResponse struct {
 	Response
 	Result []WaitingRoom `json:"result"`
+}
+
+// WaitingRoomSettingsResponse is the API response, containing zone-level Waiting Room settings.
+type WaitingRoomSettingsResponse struct {
+	Response
+	Result WaitingRoomSettings `json:"result"`
 }
 
 // WaitingRoomStatusResponse is the API response, containing the status of a waiting room.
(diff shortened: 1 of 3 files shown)


jacobbednarz:
thanks for the PR @piperswe 🙇 the changes themselves look fine (minor nitpicks inline) however, for the method signatures, i'd recommend swapping to the method signature conventions outlined in the [experimental documentation](https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md) which allows more flexibility but stronger type guarantees for the future. if you're looking for some prior art, you can checkout #1270 which was a new resource added within the last couple of weeks that follows these conventions. i've also commented on one of the methods to help you get started.

waiting_room.go
```diff
@@ -532,3 +550,56 @@ func (api *API) DeleteWaitingRoomRule(ctx context.Context, rc *ResourceContainer
 
 	return r.Result, nil
 }
+
+// WaitingRoomSettings fetches the Waiting Room zone-level settings for a zone.
+//
+// API reference: https://api.cloudflare.com/#waiting-room-get-zone-settings
+func (api *API) WaitingRoomSettings(ctx context.Context, zoneID string) (WaitingRoomSettings, error) {
```
jacobbednarz: here is the example from my comment in the PR review.

```suggestion
func (api *API) GetWaitingRoomSettings(ctx context.Context, *rc ResourceContainer, params GetWaitingRoomSettingsParams) (WaitingRoomSettings, error) {
```

even if you don't need the `GetWaitingRoomSettingsParams` right now, we'll add it to match the conventions and make it easy to extend in the future.

waiting_room.go
```diff
@@ -76,6 +76,18 @@ type WaitingRoomRule struct {
 	Enabled     *bool      `json:"enabled"`
 }
 
+// WaitingRoomSettings describes zone-level waiting room settings.
+type WaitingRoomSettings struct {
+	// Whether to allow verified search engine crawlers to bypass all waiting rooms on this zone
+	SearchEngineCrawlerBypass bool `json:"search_engine_crawler_bypass"`
```
jacobbednarz, line 82: at this point, i recommend all booleans be defined as pointers due to the number of value distinction issues we've hit in the past.

waiting_room.go
```diff
@@ -532,3 +550,56 @@ func (api *API) DeleteWaitingRoomRule(ctx context.Context, rc *ResourceContainer
...
+		return WaitingRoomSettings{}, fmt.Errorf("%s: %w", errUnmarshalError, err)
+	}
+	return r.Result, nil
+}
+
+// ChangeWaitingRoomSettings lets you change individual zone-level Waiting Room settings. This is
+// in contrast to UpdateWaitingRoomSettings which replaces all settings.
+//
+// API reference: https://api.cloudflare.com/#waiting-room-patch-zone-settings
+func (api *API) ChangeWaitingRoomSettings(ctx context.Context, zoneID string, settings WaitingRoomSettingsPatch) (WaitingRoomSettings, error) {
```
jacobbednarz: this is an interesting convention. in the past, we've just supplied one method that is capable of individually updating in place for 1 or more values. i don't have a better practice to recommend at the moment so if you want to keep both, i'm 👍

waiting_room_test.go
```diff
@@ -160,6 +168,14 @@ var waitingRoomRule = WaitingRoomRule{
 	LastUpdated: &testTimestampWaitingRoom,
 }
 
+var waitingRoomSettings = WaitingRoomSettings{
+	SearchEngineCrawlerBypass: true,
+}
+
+var waitingRoomSettingsPatch = WaitingRoomSettingsPatch{
+	SearchEngineCrawlerBypass: &t,
```
jacobbednarz: to save yourself doing this, the library does have a `BoolPtr` helper you can lean on for easier type conversion.

waiting_room.go
piperswe: 👍🏻

waiting_room.go
piperswe: `search_engine_crawler_bypass` will always be present in the response, and will always be either `true` or `false`. I think it would be misleading to make it be a tristate. I made it a boolean in the request params though, if the user wants to fallback to the default.

jacobbednarz: if it's always present, 👍

waiting_room.go
```diff
@@ -532,3 +544,76 @@ func (api *API) DeleteWaitingRoomRule(ctx context.Context, rc *ResourceContainer
...
+	}
+	return r.Result, nil
+}
+
+type UpdateWaitingRoomSettingsParams struct {
+	SearchEngineCrawlerBypass *bool `json:"search_engine_crawler_bypass,omitempty"`
+}
+
+// UpdateWaitingRoomSettings lets you replace all zone-level Waiting Room settings. This is in contrast to
+// ChangeWaitingRoomSettings which lets you change individual settings.
```
jacobbednarz:
```suggestion
// PatchWaitingRoomSettings which lets you change individual settings.
```

PR #1265, opened 2023-04-13
Add GetR2Bucket

Cyb3r-Jak3:
## Description

- Adds `GetR2Bucket` for getting info for individual buckets
- Changes CreationTime to *time.Time

## Has your change been tested?

All tests pass

## Types of changes

What sort of change does your code introduce/modify?

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.
- [x] This change is using publicly documented (api.cloudflare.com or developers.cloudflare.com) and stable APIs. (soon™️)

PR #1278, opened 2023-05-08
add support for cache reserve

### LINKED ISSUE #1277: Add support for Cache Reserve

v0.66.0

### Description

Cache Reserve appear to have come out of beta recently (see https://github.com/cloudflare/cloudflare-docs/pull/8561)

We use terraform and the cloudflare-terraform provider to manage all/most of our Cloudflare services and settings. Cache Reserve is currently not available in either. When we asked our Cloudflare reps we were told it would eventually become available once Cache Reserve is out of beta

Handy links:
* https://www.cloudflare.com/en-ca/products/cache-reserve/
* https://developers.cloudflare.com/cache/about/cache-reserve/
* https://developers.cloudflare.com/api/operations/zone-cache-settings-get-cache-reserve-setting

### Use cases

The primary goal for us is to be able to get the ability to configure Cache Reserve via terraform. For that to be possible, cloudflare-go needs to support this capability so that it can be added to https://github.com/cloudflare/terraform-provider-cloudflare (will create a feature request issue there as well)

At present, we manage everything through Terraform, except for Cache Reserve which we enable manually in the dashboard. This is less than ideal for companies relying on IaC and gitops processes.

```go
Implementation would probably be fairly standard as it is a simple on/off value
```

### References

_No response_

jfchevrette:
## Description

Add support for retrieving and updating the `Cache Reserve` setting on a zone

This is a rather naive implementation which follows the same pattern as the argo.go methods

Closes #1277 

Unit tests included

I also tested locally on a real account and I was able to retrieve and set the cache_reserve setting

## Types of changes

## Checklist:


zone_cache_reserve.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	return cacheReserveDetailsResponse.Result, nil
+}
+
+// UpdateZoneCacheReserve updates the cache reserve setting for a zone
+//
+// API reference: https://developers.cloudflare.com/api/operations/zone-cache-settings-change-cache-reserve-setting
+func (api *API) UpdateZoneCacheReserve(ctx context.Context, zoneID string, value string) (ZoneCacheReserve, error) {
+	if !contains(validOnOffSettingValues, value) {
+		return ZoneCacheReserve{}, fmt.Errorf("invalid setting value '%s'. must be 'on' or 'off'", value)
```
kwilczynski: @jfchevrette, this might be validated server-side.

I am unsure what the SDK/client policy is here (@jacobbednarz might offer some guidance), but I would leave it to the backend to return an error that the client then bubbles up.

This validation would make much more sense in the Terraform provider, where avoiding this round trip would make sense.

jfchevrette: I have no insight into the backend so I can't comment on that. I got the validation inspiration from https://github.com/cloudflare/cloudflare-go/blob/9388783618dc500a0a8a942c407104c81de77d9a/argo.go#L53

Additional validation can of course be done in the terraform provider but this is beyond this PR

jacobbednarz: yep, this validation can go as it's performed server side. we don't want validation in three places (lib, terraform provider and service) that all have different sources.

zone_cache_reserve.go
```diff
@@ -0,0 +1,79 @@
...
+
+type zoneCacheReserveSingleResponse struct {
+	Response
+	Result ZoneCacheReserve `json:"result"`
+}
+
+// GetZoneCacheReserve returns information about the current cache reserve settings
+//
+// API reference: https://developers.cloudflare.com/api/operations/zone-cache-settings-get-cache-reserve-setting
+func (api *API) GetZoneCacheReserve(ctx context.Context, zoneID string) (ZoneCacheReserve, error) {
```
jacobbednarz: have a look at https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md for how we are managing method signatures going forward for a consistent experience. #1265 has some prior art for these method signatures if you'd like something to lean on.

zone_cache_reserve.go
jfchevrette: I have removed the validation. Thank you

zone_cache_reserve.go
jfchevrette: I fail to see how I can map this to the service pattern described in experimental as cache reserve is not exposed as a service/entity on its own but more like a setting tied to a zone

Would it make sense to implement it (mostly move the code from the PR) as methods in the zone.go file? Similarly to the Zone SSL Settings methods here https://github.com/cloudflare/cloudflare-go/blob/10cc0fdc82ef5308226a1273f1f769f0de92c7eb/zone.go#L826-L858

jacobbednarz: the entity/service/relationships portion of this doesn't really matter as every endpoint is essentially it's own entity and should be treated as such. the intent is to get the definition and accessing the entity the same across the board (which is what the experimental client is about).

i've gone ahead and updated this PR with the experimental client signatures. in future, if we decide to also expose this as a zone level setting, we can explore that interface at another time.

PR #1303, opened 2023-06-08
tunnel: introduce custom `Duration` for time fields

jacobbednarz:
All Tunnel timeout values are based on seconds however, there isn't a great way to do this in a flexible way with Go due to `time.Duration` [not having marshal/unmarshal support in Go 1](https://github.com/golang/go/issues/10275). Instead, we introduce a custom `TunnelDuration` value here and ensure it always converts durations into seconds without impacting other users of `time.Duration`.

This is similarly handled in [cloudflared too](https://github.com/cloudflare/cloudflared/blob/a3bcf25faeb02c8d8e0ef2d758cd8ed6aaba852a/config/configuration.go#L426).

### DIFF
--- a/tunnel_test.go
+++ b/tunnel_test.go
@@ -189,6 +189,7 @@ func TestUpdateTunnelConfiguration(t *testing.T) {
 		fmt.Fprint(w, loadFixture("tunnel", "configuration"))
 	}
 
+	timeout, _ := time.ParseDuration("10s")
 	mux.HandleFunc(fmt.Sprintf("/accounts/%s/cfd_tunnel/%s/configurations", testAccountID, testTunnelID), handler)
 	want := TunnelConfigurationResult{
 		TunnelID: testTunnelID,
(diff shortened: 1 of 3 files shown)


jacobbednarz:
cc @obezuk

tunnel_test.go
```diff
@@ -8,6 +8,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/davecgh/go-spew/spew"
```
Cyb3r-Jak3: Don't think this is needed outside of debugging

tunnel_test.go
```diff
@@ -275,12 +278,13 @@ func TestGetTunnelConfiguration(t *testing.T) {
...
 			},
 			OriginRequest: OriginRequestConfig{
-				ConnectTimeout: DurationPtr(10),
+				ConnectTimeout: &TunnelDuration{timeout},
 			},
 		}}
 
 	actual, err := client.GetTunnelConfiguration(context.Background(), AccountIdentifier(testAccountID), testTunnelID)
 
+	spew.Dump(actual)
```
Cyb3r-Jak3: Looks like used for debugging

PR #1308, opened 2023-06-14
adds method to create image v2 direct upload url

### LINKED ISSUE #1248: Image direct upload V2

v0.63.0

### Description

https://developers.cloudflare.com/api/operations/cloudflare-images-create-authenticated-direct-upload-url-v-2

Cloudflare has a version 2 of image direct upload which supports additional parameters like signed urls, metadata...
If approved I can make the code changes and submit a PR.

### Use cases

This will implement the API

```go
Make a new function CreateImageDirectUploadURLV2 to keep backward compatibility
```

### References

_No response_

bbbshah:
This PR implements the missing feature to support `images/v2/direct_upload`
endpoint that lets client create direct upload url. The `v2` version supports
`metadata` and `requireSignedURLs` as additional request parameters.

## Description

Since the `v2` request supports `metadata` and `requireSignedURLs`,
a new struct `ImageDirectUploadURLV2Request` is added to images.go.
This struct is then passed as a param to `CreateImageDirectUploadURLV2`

Closes #1248 

Added a unit test for the newly added function. Also tested against a golang
client to create direct upload url with `metatadata` and `requireSignedURLs`,
and verified the uploaded image was private and retained all the `metadata`.

## Types of changes

## Checklist:

      and relies on stable APIs.


images.go
```diff
@@ -197,6 +203,57 @@ func (api *API) CreateImageDirectUploadURL(ctx context.Context, accountID string
 	return imageDirectUploadURLResponse.Result, nil
 }
 
+var imagesMultipartBoundary = "----CloudflareImagesGoClientBoundary"
+
+// CreateImageDirectUploadURLV2 creates an authenticated v2 direct upload url.
+//
+// API Reference: https://developers.cloudflare.com/api/operations/cloudflare-images-create-authenticated-direct-upload-url-v-2
+func (api *API) CreateImageDirectUploadURLV2(ctx context.Context, accountID string, params ImageDirectUploadURLV2Request) (ImageDirectUploadURL, error) {
```
jacobbednarz: can we update this to align with the library conventions ([link](https://github.com/cloudflare/cloudflare-go/blob/master/docs/conventions.md))?

i don't love this method name but due to the version existing in the URL, we don't really have much else to offer here. i'll chat with the team internally and see what we can do here.

jacobbednarz: ...maybe a `Version` field on the params struct that toggles the URL and we roll this into the existing create method?

bbbshah: The `CreateImageDirectUploadURLV2` method is significantly different from the existing method and may result into a series of if/else blocks and extracted private methods specific to each version. The request payload is also different between the two versions. The current proposed changes seems like the most clean way to do this. I'm open to other suggestions?

I've updated `ImageDirectUploadURLV2Request` with library conventions to use pointers for `bool` and `time.Time`. Was there anything else you wanted to change to meet the conventions?

jacobbednarz: we don't want V1/V2/V3 as part of the method call since that should be abstracted from end users and they should just be able to configure/control the method as needed instead. i'm fine with private methods that we invoke from inside the shared method with a specific comment documenting it.

as for the library conventions, it looks like the only one missing is the second parameter should be `ResourceContainer`, not `string` accountID.

bbbshah: Thank you for the input, i've made necessary changes. I ended up keeping all the implementation in the same function to keep it simple for now. Also updated the second parameter to by `ResourceContainer` instead of string. Let me know if there are other changes needed.

jacobbednarz:
i moved this to #1322 because your fork didn't allow maintainers to make edits.

PR #1330, opened 2023-07-07
dispatch namespace features

aaronlisman:
## Description

Adds support for several Workers for Platforms features:
- dispatch namespace bindings
- uploading workers 'to' a dispatch namespace

Also includes a few internal-ish items:
- 'unsafe' bindings, similar to wrangler's: https://github.com/cloudflare/workers-sdk/pull/411
- 'pipeline_hash', which is used internally and in a few experimental features. This is also present in wrangler: https://github.com/cloudflare/workers-sdk/blob/1ce32968b990fef59953b8cd61172b98fb2386e5/packages/wrangler/src/deploy/deploy.ts#L635

Yes, besides `pipeline_hash`, but they've only been tested against the mock handlers in the included tests.

## Types of changes

- [❌ ] Bug fix (non-breaking change which fixes an issue)
- [✅ ] New feature (non-breaking change which adds functionality)
- [❌ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [✅] My code follows the code style of this project.
- [❓] My change requires a change to the documentation.
  - Possibly. The public features are documented on ` developers.cloudflare.com `, but I don't believe the library changes are.
- [❌] I have updated the documentation accordingly.
- [✅] I have added tests to cover my changes.
- [✅] All new and existing tests passed.
- [❓] This change is using publicly documented in [cloudflare/api-schemas](https://github.com/cloudflare/api-schemas) 
      and relies on stable APIs.
  - The WFP APIs should be stable.
  - The 'unsafe' binding API is specifically to avoid adding support for some unstable features.
  - The `pipeline_hash` field is internal and unstable, but has been requested by some users.

### DIFF
--- a/workers_test.go
+++ b/workers_test.go
@@ -1209,7 +1209,7 @@ func TestUploadWorker_ToDispatchNamespace(t *testing.T) {
 		assert.Equal(t, http.MethodPut, r.Method, "Expected method 'PUT', got %s", r.Method)
 
 		mpUpload, err := parseMultipartUpload(r)
-		assert.NoError(t, err)
+		require.NoError(t, err)
 
 		assert.Equal(t, workerScript, mpUpload.Script)
 
@@ -1242,7 +1242,7 @@ func TestUploadWorker_UnsafeBinding(t *testing.T) {
 		assert.Equal(t, http.MethodPut, r.Method, "Expected method 'PUT', got %s", r.Method)
 
 		mpUpload, err := parseMultipartUpload(r)
-		assert.NoError(t, err)
+		require.NoError(t, err)
 
 		assert.Equal(t, workerScript, mpUpload.Script)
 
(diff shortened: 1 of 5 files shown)


workers_test.go
```diff
@@ -1140,3 +1198,74 @@ func TestUploadWorker_WithTailConsumers(t *testing.T) {
...
+	setup()
+	defer teardown()
+
+	namespaceName := "n1"
+
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, http.MethodPut, r.Method, "Expected method 'PUT', got %s", r.Method)
+
+		mpUpload, err := parseMultipartUpload(r)
+		assert.NoError(t, err)
```
ndisidore: super nit: this should probably be a `require.NoError`, as it doesn't make sense for the test to continue if this errors out

ndisidore (approved):

PR #1335, opened 2023-07-15
Add image upload by url

LewisYearsley:
## Description

The Cloudflare image API provides the ability to upload an image via a url, rather than as bytes. e.g. you might want to copy the image from some bucket to Cloudflare without having to manage the data yourself. This PR adds that functionality, heavily based on the existing file upload methodology.

Added test in the same fashion as the file based upload method.

## Types of changes

## Checklist:

      and relies on stable APIs.


jacobbednarz:
thanks for the PR here, appreciate it! for me though, i would rather see `UploadImage` with the ability to upload all images regardless of the method used. to get this over the line, let's update the methods here to conditionally build the payload (even if you need to use multiple internal methods) instead of the multiple methods which i believe is a better experience for anyone attempting to use this.

LewisYearsley:
No problem and for sure - I was actually in two minds about whether to have a combined signature! I’ll rework it today - thanks for looking.

dojolew:
@jacobbednarz handled your feedback - mind taking another look when you get time? 🙏

jacobbednarz:
thanks for this one! appreciate it 🙇

LewisYearsley:
Thanks @jacobbednarz! Any idea when this might be released?

jacobbednarz:
you can view the policy at https://github.com/cloudflare/cloudflare-go/blob/master/docs/changelog-process.md

PR #1344, opened 2023-08-01
AUTH-5315 adds new conditional_access_enabled attr to access IDP

ghost:
## Description

* adds new conditional_access_enabled attr to access IDP
* list/put IDP auth context endpoints
* auth context rule type

* Added unit tests for policy creation w/new rule and auth context endpoints

## Types of changes

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/access_group.go
+++ b/access_group.go
@@ -153,6 +153,15 @@ type AccessGroupSAML struct {
 	} `json:"saml"`
 }
 
+// AccessGroupAzureAuthContext is used to configure access based on Azure auth contexts.
+type AccessGroupAzureAuthContext struct {
+	AuthContext struct {
+		ID                 string `json:"id"`
+		IdentityProviderID string `json:"identity_provider_id"`
+		ACID               string `json:"ac_id"`
+	} `json:"auth_context"`
+}
+
 // AccessGroupAuthMethod is used for managing access by the "amr"
 // (Authentication Methods References) identifier. For example, an
 // application may want to require that users authenticate using a hardware
--- a/access_identity_provider.go
+++ b/access_identity_provider.go
@@ -90,7 +90,7 @@ type UpdateAccessIdentityProviderParams struct {
 	ScimConfig AccessIdentityProviderScimConfiguration `json:"scim_config"`
 }
 
-// AccessAuthContext represents an Access Azure Identity Provider Auth Context
+// AccessAuthContext represents an Access Azure Identity Provider Auth Context.
 type AccessAuthContext struct {
 	ID          string `json:"id"`
 	UID         string `json:"uid"`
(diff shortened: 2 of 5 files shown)


access_group.go
```diff
@@ -153,6 +153,15 @@ type AccessGroupSAML struct {
 	} `json:"saml"`
 }
 
+// AccessGroupAzureAuthContext is used to configure access based on Azure auth contexts.
+type AccessGroupAzureAuthContext struct {
+	AuthContext struct {
```
jacobbednarz, line 158: can we use external structs here? inline structs are hard to use inline.

access_identity_provider.go
```diff
@@ -235,3 +252,51 @@ func (api *API) DeleteAccessIdentityProvider(ctx context.Context, rc *ResourceCo
...
+// Zone API Reference: https://developers.cloudflare.com/api/operations/zone-level-access-identity-providers-update-an-access-identity-provider
+func (api *API) UpdateAccessIdentityProviderAuthContexts(ctx context.Context, rc *ResourceContainer, identityProviderID string) (AccessIdentityProvider, error) {
+	uri := fmt.Sprintf(
+		"/%s/%s/access/identity_providers/%s/auth_context",
+		rc.Level,
+		rc.Identifier,
+		identityProviderID,
+	)
+
+	res, err := api.makeRequestContext(ctx, http.MethodPut, uri, nil)
```
jacobbednarz, line 290: no payload?

access_group.go
ghost: For code consistency I'd say not in this PR if possible. We currently define all our groups this way

access_identity_provider.go
ghost: Yes this is similar to our refreshSCIM endpoint (although now that I'm looking maybe it's not in this repo yet). Essentially a PUT to /identity_providers/<ID>/auth_contexts will pull the updates from their IDP, they don't need to be provided in the request.

access_group.go
jacobbednarz: ah you're right. we can look at that another time.

access_identity_provider.go
```diff
@@ -89,6 +90,22 @@ type UpdateAccessIdentityProviderParams struct {
 	ScimConfig AccessIdentityProviderScimConfiguration `json:"scim_config"`
 }
 
+// AccessAuthContext represents an Access Azure Identity Provider Auth Context
```
jacobbednarz:
```suggestion
// AccessAuthContext represents an Access Azure Identity Provider Auth Context.
```

jacobbednarz (commented):
lgtm once we have public docs 🚀

PR #1346, opened 2023-08-01
Remove some internal details for Rulesets API

zakcutner:
## Description

* Removes some internal-only details that are not usable in the public API.
* Updates the API reference links for Rulesets API.

Using the existing Go unit tests, no new functionality has been added.

## Types of changes

This is technically a breaking change in the Go API due to some fields being removed, although it would not have worked to use these fields with the Rulesets API in any case.

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/rulesets.go
+++ b/rulesets.go
@@ -716,13 +716,11 @@ type UpdateRulesetResponse struct {
 type ListRulesetsParams struct{}
 
 type CreateRulesetParams struct {
-	Name                     string        `json:"name,omitempty"`
-	Description              string        `json:"description,omitempty"`
-	Kind                     string        `json:"kind,omitempty"`
-	Version                  *string       `json:"version,omitempty"`
-	Phase                    string        `json:"phase,omitempty"`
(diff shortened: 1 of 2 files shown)


rulesets.go
```diff
@@ -734,19 +730,16 @@ type UpdateRulesetParams struct {
 }
 
 type UpdateEntrypointRulesetParams struct {
-	Name        string        `json:"name,omitempty"`
```
zakcutner, line 737: This technically can be useful to specify but has some slightly inconsistent behaviour from the API if it is. If you specify it and the entry point ruleset didn't exist before, then this name will be used. If the entry point ruleset already exists though, this name will be ignored (and the ruleset will not be renamed).

Omitting the name always works consistently and will cause the API to set it to "default" if an entry point ruleset does not already exist. This is why I think it's probably best to not allow users to set the name explicitly to avoid confusion, but we can keep it here if that would be too much of a breaking change. What do you think @jacobbednarz?

jacobbednarz: works for me!

PR #1354, opened 2023-08-02
Add Workers tagging API methods

slester:
Today we must use a Raw request to add tags to Workers scripts; this implements it as a native Go API call.

## Description
Supports both namespace and unnamespaced Workers.

## Has your change been tested?
Yes, as we implemented it similarly ourselves

## Types of changes

What sort of change does your code introduce/modify?

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [ ] This change is using publicly documented in [cloudflare/api-schemas](https://github.com/cloudflare/api-schemas) 
      and relies on stable APIs.

PR #1355, opened 2023-08-02
Add RawResponse method which returns the standard Response with Result

slester:
We are manually tagging workers using an API that is not yet a part of the Go library (see #1354), but we don't have access to the response's Error, Messages, etc. when using the `Raw()` method, since it just returns the `"result"` field. This also returns the Response if applicable (`Success`, `Errors`, etc.)

## Description
Simply added a new method that returned `r` instead of `r.Result`

## Types of changes

## Checklist:

      and relies on stable APIs.


jacobbednarz:
given `Raw` and `RawResponse` are basically the same thing with a different part of the response returned, is there a reason we don't modify `Raw` to return the entire payload instead?

slester:
> given `Raw` and `RawResponse` are basically the same thing with a different part of the response returned, is there a reason we don't modify `Raw` to return the entire payload instead?

I'm totally up for that! I just didn't want to break existing APIs.

jacobbednarz:
for a low level method like this, I think it is fine. just be sure to include a changelog entry and 🚀

slester:
@jacobbednarz - thanks for the input, I've made those changes.

PR #1356, opened 2023-08-03
add per-hostname tls settings api and tests

vasilzhigilei:
## Description

Per-Hostname TLS Settings is a new API where customers can set TLS settings on a per-hostname basis while not being coupled to any specific certificate. SSL/TLS team.

Added unit tests.

## Types of changes

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/per_hostname_tls_settings.go
+++ b/per_hostname_tls_settings.go
@@ -1,0 +1,117 @@
+package cloudflare
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"time"
+)
+
+// HostnameTLSSetting represents the metadata for a user-created tls setting.
+type HostnameTLSSetting struct {
+	Hostname  string      `json:"hostname"`
+	Value     interface{} `json:"value"`
+	Status    string      `json:"status"`
+	CreatedAt time.Time   `json:"created_at"`
+	UpdatedAt time.Time   `json:"updated_at"`
+}
+
+// HostnameTLSSettingResponse represents the response from the PUT and DELETE endpoints for per-hostname tls settings.
+type HostnameTLSSettingResponse struct {
+	Response
+	Result HostnameTLSSetting `json:"result"`
+}
+
+// HostnameTLSSettingsResponse represents the response from the retrieval endpoint for per-hostname tls settings.
+type HostnameTLSSettingsResponse struct {
+	Response
+	Result     []HostnameTLSSetting `json:"result"`
+	ResultInfo `json:"result_info"`
+}
+
+// EditHostnameTLSSettingParams represents the data related to the per-hostname tls setting being edited.
+type EditHostnameTLSSettingParams struct {
+	Value interface{} `json:"value"`
+}
+
+type ListHostnameTLSSettingsParams struct {
+	PaginationOptions
+	Limit    int      `url:"limit,omitempty"`
+	Offset   int      `url:"offset,omitempty"`
+	Hostname []string `url:"hostname,omitempty"`
+}
+
+var (
+	ErrMissingHostnameTLSSettingName = errors.New("tls setting name required but missing")
+)
+
+// ListHostnameTLSSettings returns a list of all user-created tls setting values for the specified setting and hostnames.
+//
+// API reference: https://developers.cloudflare.com/api/operations/per-hostname-tls-settings-list
+func (api *API) ListHostnameTLSSettings(ctx context.Context, rc *ResourceContainer, setting string, params ListHostnameTLSSettingsParams) ([]HostnameTLSSetting, ResultInfo, error) {
+	if rc.Identifier == "" {
+		return []HostnameTLSSetting{}, ResultInfo{}, ErrMissingZoneID
+	}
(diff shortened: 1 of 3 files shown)


per_hostname_tls_settings.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+var (
+	ErrMissingHostnameTLSSettingName = errors.New("tls setting name required but missing")
+)
+
+// ListHostnameTLSSettings returns a list of all user-created tls setting values for the specified setting and hostnames.
+//
+// API reference: https://developers.cloudflare.com/api/operations/per-hostname-tls-settings-list
+func (api *API) ListHostnameTLSSettings(ctx context.Context, rc *ResourceContainer, setting string, params ListHostnameTLSSettingsParams) ([]HostnameTLSSetting, ResultInfo, error) {
```
jacobbednarz: the `setting` attribute should be a part of `params`. checkout https://github.com/cloudflare/cloudflare-go/blob/master/docs/conventions.md#methods for the latest docs on this.

per_hostname_tls_settings.go
```diff
@@ -0,0 +1,117 @@
...
+	if err := json.Unmarshal(res, &r); err != nil {
+		return []HostnameTLSSetting{}, ResultInfo{}, fmt.Errorf("%s: %w", errUnmarshalError, err)
+	}
+	return r.Result, r.ResultInfo, err
+}
+
+// EditHostnameTLSSetting will update the per-hostname tls setting for the specified hostname.
+//
+// API reference: https://developers.cloudflare.com/api/operations/per-hostname-tls-settings-put
+func (api *API) EditHostnameTLSSetting(ctx context.Context, rc *ResourceContainer, setting, hostname string, params EditHostnameTLSSettingParams) (HostnameTLSSetting, error) {
```
jacobbednarz: prefer `Update` as the prefix over "edit" or "change".

similar to https://github.com/cloudflare/cloudflare-go/pull/1356/files#r1283940593, we should update the method signature here.

per_hostname_tls_settings.go
```diff
@@ -0,0 +1,117 @@
...
+	if err := json.Unmarshal(res, &r); err != nil {
+		return HostnameTLSSetting{}, fmt.Errorf("%s: %w", errUnmarshalError, err)
+	}
+	return r.Result, nil
+}
+
+// DeleteHostnameTLSSetting will delete the specified per-hostname tls setting.
+//
+// API reference: https://developers.cloudflare.com/api/operations/per-hostname-tls-settings-delete
+func (api *API) DeleteHostnameTLSSetting(ctx context.Context, rc *ResourceContainer, setting, hostname string) (HostnameTLSSetting, error) {
```
jacobbednarz: again, just updating the method signature here https://github.com/cloudflare/cloudflare-go/pull/1356/files#r1283940593

per_hostname_tls_settings.go
```diff
@@ -0,0 +1,117 @@
...
+	"errors"
+	"fmt"
+	"net/http"
+	"time"
+)
+
+// HostnameTLSSetting represents the metadata for a user-created tls setting.
+type HostnameTLSSetting struct {
+	Hostname  string      `json:"hostname"`
+	Value     interface{} `json:"value"`
```
jacobbednarz: this is a bit of a smell for me (not knowing what the underlying type is) and will cause you issues in other tooling like terraform. do we need to accept multiple types here? can we settle on one?

per_hostname_tls_settings.go
```diff
@@ -0,0 +1,117 @@
...
+	Status    string      `json:"status"`
+	CreatedAt time.Time   `json:"created_at"`
```
jacobbednarz: https://github.com/cloudflare/cloudflare-go/blob/master/docs/conventions.md#timetime

per_hostname_tls_settings.go
vasilzhigilei: API input can be float, string, or list of strings (and the response will be string or list of strings for tls setting value)
The float part doesn't matter because the float input can be a string too.
But the list of strings cannot be sent in as a string to the API. It must be an actual list.
http2 setting can be "on" or "off"
min_tls_version setting can be float or string representations of 1.0, 1.1, 1.2, 1.3
ciphers setting must be a list of strings such as ["AES128-GCM-SHA256","ECDHE-RSA-AES128-GCM-SHA256"]

It's the ciphers input that makes this tricky. I see no other way other than to make this be an interface - at least if we want it to follow how the API works. For the Go library, I guess we can just always make it be a string. And I can make the func then convert that string into a list of ciphers. Would this be better?

Customer would then always input a string and the library converts it to what it needs to be (in the case of ciphers), and converts the API response ciphers list into a string.

EDIT: going to implement custom marshal/unmarshal and make the value always be a string to end-user

per_hostname_tls_settings.go
vasilzhigilei: Alright, will do. Wasn't sure since most implementations send params straight into the HTTP request, so I'll add some logic to split the params input into "going into the URL structure" vs "actual HTTP params"

PR #1365, opened 2023-08-09
Create Zone Hold functionality

### LINKED ISSUE #1364: Cloudflare Zone Hold Support

v0.74.0

### Description

Add sdk support for creating/deleting holds:

https://developers.cloudflare.com/api/operations/zones-0-hold-post
https://developers.cloudflare.com/api/operations/zones-0-hold-get

### Use cases

Add or remove zone holds via sdk

Interact with the API via:
https://github.com/cloudflare/terraform-provider-cloudflare/pull/2671

### References

_No response_

danquack:
## Description

Adds the ability to create a zone hold based on the hold API.

I called the apis manually. I am happy to write tests, but I'm not sure where to get started. I could stub the responses, but since this is basically just calling the raw api, im not sure what I would be testing (no additional logic).

## Types of changes

## Checklist:

      and relies on stable APIs.

closes #1364


hold.go
```diff
@@ -0,0 +1,112 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/goccy/go-json"
+)
+
+// ZoneHold represents a Zone Hold rule. A rule only permits access to
```
patryk: The comment doesn't match the expected behaviour. You may use copy found on https://developers.cloudflare.com/api/operations/zones-0-hold-post.

hold.go
```diff
@@ -0,0 +1,112 @@
...
+// endpoint.
+type ZoneHoldDeleteResponse struct {
+	Result ZoneHold `json:"result"`
+}
+
+// ZoneHoldCreateParams contains required and optional params
+// for creating a zone Hold.
+type ZoneHoldCreateParams struct {
+	ID                string `json:"id"`
+	Hold              bool   `json:"hold"`
```
patryk: These two fields are not defined for POST request.

hold.go
```diff
@@ -0,0 +1,112 @@
...
+	IncludeSubdomains *bool  `json:"include_subdomains,omitempty"`
+}
+
+// ZoneHoldDeleteParams contains required and optional params
+// for updating a zone Hold.
+type ZoneHoldDeleteParams struct {
+	ID        string  `json:"id"`
+	Hold      bool    `json:"hold"`
```
patryk: Same.

hold.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// ZoneHoldResponse represents a response from the Zone Hold endpoint.
+type ZoneHoldResponse struct {
+	Result ZoneHold `json:"result"`
+	Response
+	ResultInfo `json:"result_info"`
+}
+
+// ZoneHoldDeleteResponse represents a response from the List Zone Hold
```
tpaulus: Comment describes a List Zone Hold endpoint, which does not exist.

hold.go
```diff
@@ -0,0 +1,112 @@
...
+// endpoint.
+type ZoneHoldDeleteResponse struct {
+	Result ZoneHold `json:"result"`
+}
+
+// ZoneHoldCreateParams contains required and optional params
```
tpaulus: There is only 1 parameter (`include_subdomains`) for Create Zone Hold Request, and it is optional.

hold.go
danquack: A lot of these were copy pasta from a different file. I updated not, can you check if that works for you.

hold.go
```diff
@@ -0,0 +1,112 @@
...
+// for creating a zone Hold.
+type ZoneHoldCreateParams struct {
+	ID                string `json:"id"`
+	Hold              bool   `json:"hold"`
+	IncludeSubdomains *bool  `json:"include_subdomains,omitempty"`
+}
+
+// ZoneHoldDeleteParams contains required and optional params
```
tpaulus: There is only 1 parameter (`hold_after`) for a Delete Zone Hold request and it is optional.

hold.go
```diff
@@ -0,0 +1,112 @@
...
+// for updating a zone Hold.
+type ZoneHoldDeleteParams struct {
+	ID        string  `json:"id"`
+	Hold      bool    `json:"hold"`
+	HoldAfter *string `json:"hold_after,omitempty"`
```
tpaulus: This must be an RFC3339-formatted timestamp, should the type be changed to `time` and this conversion done as part of the Delete Operation to ensure operation performs as expected?

hold.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// CreateZoneHold creates a Zone ZoneHold rule for the given zone ID.
+//
+// API reference: https://developers.cloudflare.com/api/operations/zones-0-hold-post
+func (api *API) CreateZoneHold(ctx context.Context, zoneID string, params ZoneHoldCreateParams) (ZoneHold, error) {
+	uri := fmt.Sprintf("/zones/%s/hold", zoneID)
+	if params.IncludeSubdomains != nil {
+		uri = fmt.Sprintf("%s?include_subdomains=%t", uri, *params.IncludeSubdomains)
```
tpaulus: Thoughts on using `url` package's `Values` type for this encoding?

https://stackoverflow.com/a/36000522/8156448

hold.go
```diff
@@ -0,0 +1,112 @@
...
+	return response.Result, nil
+}
+
+// DeleteZoneHold deletes a Zone ZoneHold rule (based on the ID) for the given zone ID.
+//
+// API reference:https://developers.cloudflare.com/api/operations/zones-0-hold-delete
+func (api *API) DeleteZoneHold(ctx context.Context, zoneID string, params ZoneHoldDeleteParams) (ZoneHold, error) {
+	uri := fmt.Sprintf("/zones/%s/hold", zoneID)
+	if params.HoldAfter != nil {
+		uri = fmt.Sprintf("%s?hold_after=%s", uri, *params.HoldAfter)
```
tpaulus: Thoughts on using `url` package's `Values` type for this encoding?

https://stackoverflow.com/a/36000522/8156448

tpaulus (changes requested):

hold.go
danquack: converted.

hold.go
danquack: Id be fine either way, my thought to use sprintf was just because it was a total of one param

tpaulus (approved):

hold.go
jacobbednarz: we have [`buildURI`](https://github.com/cloudflare/cloudflare-go/blob/80b7949c7f0b8f051b7b5e79f6272f30c552c00c/utils.go#L13-L16) for this purpose. it allows adding query parameter tags to the struct without worrying about manually building path + query parameters.

example: https://github.com/cloudflare/cloudflare-go/blob/80b7949c7f0b8f051b7b5e79f6272f30c552c00c/tunnel_routes.go#L85

hold.go
```diff
@@ -0,0 +1,109 @@
...
+// ZoneHoldDeleteParams represents params for the Delete Zone Hold
+// endpoint.
+type ZoneHoldDeleteParams struct {
+	HoldAfter *time.Time `json:"hold_after,omitempty"`
+}
+
+// Enforce a zone hold on the zone, blocking the creation and activation of zones with this zone's hostname.
+//
+// API reference: https://developers.cloudflare.com/api/operations/zones-0-hold-post
+func (api *API) CreateZoneHold(ctx context.Context, zoneID string, params ZoneHoldCreateParams) (ZoneHold, error) {
```
jacobbednarz: we need to update the method signature here to match the library conventions - https://github.com/cloudflare/cloudflare-go/blob/master/docs/conventions.md#methods

hold.go
```diff
@@ -0,0 +1,109 @@
...
+		return ZoneHold{}, fmt.Errorf("%s: %w", errUnmarshalError, err)
+	}
+
+	return response.Result, nil
+}
+
+// Retrieve whether the zone is subject to a zone hold, and metadata about the hold.
+//
+// API reference: https://developers.cloudflare.com/api/operations/zones-0-hold-get
+func (api *API) ZoneHold(ctx context.Context, zoneID string) (ZoneHold, error) {
```
jacobbednarz: updating this to follow the library  conventions for CRUD naming

```suggestion
func (api *API) GetZoneHold(ctx context.Context, zoneID string) (ZoneHold, error) {
```

we'll also need to update the signature to match https://github.com/cloudflare/cloudflare-go/blob/master/docs/conventions.md#methods

jacobbednarz (changes requested):
thanks for the PR! 

i've left some feedback inline about the library conventions we need to update here but we'll also need to add test coverage. you can check out another file like https://github.com/cloudflare/cloudflare-go/blob/master/accounts_test.go to get an idea on the minimums we expect here.

hold.go
danquack: Questions on the ResourceContainer. I copied my functions from here? Are they outside convention?

https://github.com/cloudflare/cloudflare-go/blob/64904badcb5202c3ad02d4a1c90bbfa7432f9636/firewall.go#L117

if I switch to the container, where does zone ID come from? Does it come from the container or do I have to add it as a param?

jacobbednarz: yes, that is an older method that has not yet been converted to the patterns outlined in the conventions and [experimental client doc](https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md). you're best off checking out a more modern file like https://github.com/cloudflare/cloudflare-go/blob/master/rulesets.go#L758 to see how it works.

> if I switch to the container, where does zone ID come from? Does it come from the container or do I have to add it as a param?

`*ResourceContainer` is the basic way of describing a resource and its identifier (https://github.com/cloudflare/cloudflare-go/blob/master/resource.go). the method accepts this as a parameter but end usage should use the helper method like [`ZoneIdentifier`](https://github.com/cloudflare/cloudflare-go/blob/a0f7a6b7de3d9023ef32fa9f518b94e4d68e3317/resource.go#L99-L105) which handles creating the struct and the associated relationships for you.

PR #1372, opened 2023-08-24
Update pagination to compute current page based off total_count

jroyal:
Also fixes pagination for a few Access Resources

## Description

In terraform, if you try to use the [Access Application Resource](https://registry.terraform.io/providers/cloudflare/cloudflare/latest/docs/resources/access_application.html) in a datasource it will fail to find the app because the List function will only check the first page. 

Unit tests and a local reproduction of the original issue that is now resolved.

## Types of changes

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/.changelog/1372.txt
+++ b/.changelog/1372.txt
@@ -1,0 +1,8 @@
+```release-note:bug
+pagination: Will look at `total_count` and `per_page` to calculate `total_pages` if `total_pages` is zero
+access_application: Use autopaginate flag as expected
+access_ca_certificate: Use autopaginate flag as expected
+access_group: Use autopaginate flag as expected
+access_mutual_tls_certifcate: Use autopaginate flag as expected
+access_policy: Use autopaginate flag as expected
+```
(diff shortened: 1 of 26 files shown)


jroyal:
@jacobbednarz tbh I'm a little hesitant on the new commit about changing `total_count` in all the tests. That was why the tests were failing.

I think its the correct fix, since `count` should always equal `total_count` if there are less than 1 page worth of results. Since most tests only have 1 item this should make the tests match real behavior. I couldn't decide if should try to handle that differently though in pagination.go?

jacobbednarz:
iirc, the test data came from the old API docs which didn't have real validation so it may have been incorrect the entire time. I think fixing the tests here is the right move if we've uncovered issues.

jacobbednarz:
it's worth noting, we also don't have thorough impossible logic checks for pagination which we _could_ firm that up but I suspect there are some services that have a misunderstanding of all the result info fields so it may be better to fix to be more clear instead. 

(the second part is already something we are speaking about internally too)

jacobbednarz:
thanks for this!

Issue #1378, opened 2023-08-27
ImageDetails API responds with `meta` and not `metadata`


itsdax:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the library and it is still present.

### cloudflare-go version

v0.75.0

### Go environment

v1.20

### Expected output

GetImage calls to respond with Metadata on images that contain metadata.

### Actual output

No metadata in response.

### Code demonstrating the issue

```
// Upload image with metadata and grab ImageId


img, err := cloudflareApi.GetImage(ctx, cloudflare.AccountIdentifier(b.Store.cloudflareAccountId), imageId)
if err != nil {
  return nil, err
}

fmt.Println(image.Metadata) # This is empty

```

### Steps to reproduce

1. Upload an image and add metadata to it
2. Call `GetImage`
3.`Metadata` is always empty from the response

### References

This is likely because [the API](https://developers.cloudflare.com/api/operations/cloudflare-images-image-details) doesn't use `metadata` as expected from [the code here](https://github.com/cloudflare/cloudflare-go/blob/master/images.go#L32C3-L32C3), but instead uses `meta`. Thus the library does not parse the response value.

I can debug the network requests and confirm that the image does in fact has metadata. The library simply omits it when parsing:
```
{
  "result": {
    "id": "c2132442-7029-47fa-c62f-1c7d54733501",
    "filename": "<omitted>",
    "meta": { <------- this
      "userId": "<omitted>"
    },
    "uploaded": "2023-08-27T04:20:43.315Z",
    "requireSignedURLs": true,
    "variants": [
      "https://imagedelivery.net/<omitted>/c2132442-7029-47fa-c62f-1c7d54733501/1080",
      "https://imagedelivery.net/<omitted>/c2132442-7029-47fa-c62f-1c7d54733501/public"
    ]
  },
  "success": true,
  "errors": [],
  "messages": []
}

```

PR #1379, opened 2023-08-27
Fix issue parsing Image Details from API due to incorrect struct json field

itsdax:
## Description

Issue is described here: https://github.com/cloudflare/cloudflare-go/issues/1378
API returns the metadata object under the field `meta`. However the SDK expects the field `metadata` thus it always drops this data from the response.

Yes - after changing the struct field for Metadata to `meta`, `GetImage` calls return images with Metadata field populated.

https://developers.cloudflare.com/api/operations/cloudflare-images-image-details

Refer to the response example of Image details. Specifically the `meta` field. The current SDK expects `metadata` and thus it always drops the metadata from the response.

## Types of changes

## Checklist:

      and relies on stable APIs.


images.go
```diff
@@ -29,7 +29,7 @@ const (
 type Image struct {
 	ID                string                 `json:"id"`
 	Filename          string                 `json:"filename"`
-	Metadata          map[string]interface{} `json:"metadata,omitempty"`
+	Metadata          map[string]interface{} `json:"meta,omitempty"`
```
jacobbednarz: let's rename this field to `Meta` to match the API response

jacobbednarz (commented):
thanks for the PR! added some notes inline.

images.go
itsdax: Should I categorize this as a breaking change if this would break builds of applications referencing the old field?

jacobbednarz: yeah, that's fine to update the changelog entry.

PR #1382, opened 2023-08-29
semgrep: updated IPv4 regex pattern

hiteshbedre:
## Description

Current IPv4 validation accepting non-IPv4 addresses such as 000000.010101.256.101, 192.168.0.300, 256.0.0.0 and bunch of other cases I mentioned in regex101 link. Newly defined pattern will take care all of such cases.
You can play with newly defined regex pattern here: https://regex101.com/r/j4lkR3/1

Changes has been tested on regex playground

## Types of changes

## Checklist:

      and relies on stable APIs.


jacobbednarz:
let me have a think about this one as the regex is intentionally broad to catch anything that looks like it may be an IP, not just valid ones. this ensures that even if someone provides an invalid IP, it recommends the documented ranges we expect here.

hiteshbedre:
If that is the case we need to better keep existing regex as is. Closing the pull request.

jacobbednarz:
to be clear here, I'm not against refining the regex. I just want to have a look over the previous tests cases that this helped and make sure we would have still caught them.

hiteshbedre:
Oh, I had the impression that the current broader-matching regex was acceptable. My apologies for the misinterpretation. Kindly inform me once you've had the opportunity to look into previous test cases. Reopening🚀

hiteshbedre:
@jacobbednarz Have you had a chance to review the previous test cases to ensure that this PR's changes would still catch them? Your input is valuable.

PR #1393, opened 2023-09-10
fix(dns): keep comments intact with zero values of UpdateDNSRecordParams

### LINKED ISSUE #1371: Zero values of UpdateDNSRecordParams should keep the comments
### Confirmation

### cloudflare-go version

0.75

### Go environment

N/A

### Expected output

Previously, the Go interface will keep the original comments because of `omitempty`. It's not ideal because there's no easy way to clear comments, where an empty string will keep the original comment because `omitempty` will omit empty strings. The API has since then changed to clear comments when an empty string is used, and the Go library was updated, but then one cannot easily keep the current comments. I believe the expected behavior of `Update` with zero values is to keep everything, and using strings cannot distinguish the two important cases: (1) keep the current comment and (2) clear the current comment.

Suggested solution: use `*string` for the `Comment` field and add back `omitempty`.

### Actual output

See above

```go
// This will clear the comments
c.UpdateDNSRecord(ctx, ZoneIdentifier(zone), UpdateDNSRecordParams{
		ID:      id,
		Content: ip,
	}
)
```

### Steps to reproduce

See above

### References

- https://github.com/cloudflare/cloudflare-go/pull/1166
- https://github.com/cloudflare/cloudflare-go/pull/1194

favonia:
## Description

Calling `UpdateDNSRecord` with an (almost) zero value of `UpdateDNSRecordParams` will cause the existing comments to be erased. The solution is straightforward---change the type of the field `Comment` from `string` to `&string`, just like how the field `Proxy` has the type `&bool` instead of `bool`.

Closes #1371.

Yes, but the test must be done against the real server. It is difficult to include a real test in the Go library.

## Types of changes

## Checklist:

- [x] I have added tests to cover my changes. **CAVEAT: A real test can only be done against the real API server, not against the mock server within the Go library.**
      and relies on stable APIs.

### DIFF
--- a/dns_test.go
+++ b/dns_test.go
@@ -616,7 +616,63 @@ func TestUpdateDNSRecord_ClearComment(t *testing.T) {
 
 	_, err := client.UpdateDNSRecord(context.Background(), ZoneIdentifier(testZoneID), UpdateDNSRecordParams{
 		ID:      dnsRecordID,
-		Comment: "",
+		Comment: StringPtr(""),
+	})
+	require.NoError(t, err)
+}
+
+func TestUpdateDNSRecord_KeepComment(t *testing.T) {
+	setup()
+	defer teardown()
+
+	input := DNSRecord{
+		ID: "372e67954025e0ba6aaa6d586b9e0b59",
+	}
+
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, http.MethodPatch, r.Method, "Expected method 'PATCH', got %s", r.Method)
+
+		var v DNSRecord
+		err := json.NewDecoder(r.Body).Decode(&v)
+		require.NoError(t, err)
+		v.ID = "372e67954025e0ba6aaa6d586b9e0b59"
+		assert.Equal(t, input, v)
+
+		w.Header().Set("content-type", "application/json")
+		fmt.Fprint(w, `{
+			"success": true,
+			"errors": [],
+			"messages": [],
+			"result": {
+				"id": "372e67954025e0ba6aaa6d586b9e0b59",
+				"type": "A",
+				"name": "example.com",
+				"content": "198.51.100.4",
+				"proxiable": true,
+				"proxied": false,
+				"ttl": 120,
+				"locked": false,
+				"zone_id": "d56084adb405e0b7e32c52321bf07be6",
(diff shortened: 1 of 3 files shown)


dns_test.go
```diff
@@ -476,7 +477,7 @@ func TestGetDNSRecord(t *testing.T) {
 			"auto_added": true,
 			"source":     "primary",
 		},
-		Comment: "This is a comment",
+		Comment: &comment,
```
jacobbednarz: we can use [`StringPtr`](https://github.com/cloudflare/cloudflare-go/blob/master/convert_types.go#L571) to simplify this here

dns_test.go
```diff
@@ -567,9 +568,10 @@ func TestUpdateDNSRecord_ClearComment(t *testing.T) {
 	setup()
 	defer teardown()
 
+	emptyComment := ""
 	input := DNSRecord{
 		ID:      "372e67954025e0ba6aaa6d586b9e0b59",
-		Comment: "",
+		Comment: &emptyComment,
```
jacobbednarz: similar to https://github.com/cloudflare/cloudflare-go/pull/1393/files#r1320954470

jacobbednarz:
did you manually confirm these API calls are working for your use case on the PATCH endpoint? this lgtm

favonia:
@jacobbednarz Thanks for the reminder. I did the tests about `GetDNSRecord`, `UpdateDNSRecord` and `ListDNSRecords` (but not `CreateDNSRecord`), and added some comments. (I actually made some mistake but the manual testing got it.) I decided to keep the type of the `Comment` field in `DNSRecord`, `ListDNSRecordsParams` and `CreateDNSRecordParams` as `string` (reverting my earlier commit in this PR), only changing that of `UpdateDNSRecordParams` to `&string`. The reason is that the server would omit the empty comment field when returning the results. I hope such inconsistency is okay.

jacobbednarz:
> I hope such inconsistency is okay.

it is! and it only that, this is a shining example of where the CRUD operation specific params for a method allow us to make isolated changes without breaking all the method usages at once.

favonia:
@jacobbednarz Personally I think this is ready to go!

jacobbednarz:
thank you!

PR #1395, opened 2023-09-11
refactor: replace type conversion functions with generic ones

favonia:
## Description

Use generic functions (made possible by Go 1.18) to replace all existing type conversion helper functions.

N/A

## Types of changes

Originally, I thought these functions are only for internal use and made them private, but then I realized other Cloudflare repositories are actively using these.

I changed `Ptr` to `Ref` because otherwise I don't know how to name the generic dereferencing.

## Checklist:


dlp_profile_test.go
```diff
@@ -125,7 +125,8 @@ func TestDLPProfiles(t *testing.T) {
 			},
 			CreatedAt: &createdAt,
 			UpdatedAt: &updatedAt,
-		}}
+		},
+	}
```
favonia, line 129: My linter also changed code like these lines.

waiting_room_test.go
```diff
@@ -5,21 +5,21 @@ import (
...
+var (
+	waitingRoomID                         = "699d98642c564d2e855e9661899b7252"
+	waitingRoomEventID                    = "25756b2dfe6e378a06b033b670413757"
+	waitingRoomRuleID                     = "25756b2dfe6e378a06b033b670413757"
+	testTimestampWaitingRoom              = time.Now().UTC()
+	testTimestampWaitingRoomEvent         = time.Now().UTC()
+	testTimestampWaitingRoomEventPrequeue = time.Now().UTC()
+	testTimestampWaitingRoomEventStart    = testTimestampWaitingRoomEventPrequeue.Add(5 * time.Minute)
+	testTimestampWaitingRoomEventEnd      = testTimestampWaitingRoomEventStart.Add(1 * time.Minute)
+	waitingRoomJSON                       = fmt.Sprintf(`
```
favonia, line 22: My lint also changed lines like these.

favonia:
I guess we could deprecate the current functions instead of deleting them, but on the other hand, it can be done by a simple search-and-replace.

jacobbednarz:
thanks for the PR but I think this is a little premature for us. we have a bunch of libraries (both internally and externally) that are using these types and not all are 1.18+. it is also difficult for some of them to make this migration since the my are forks and pulling in partial changes. 

for now, I'm going to close this out but once we revisit how we will be doing 1.18+ only changes, we can look at this swap too.

favonia:
@jacobbednarz That's a very reasonable decision. I only meant to bring up this possibility. :-)

PR #1401, opened 2023-09-13
add observatory support

broswen:
## Description
Add support for Cloudflare Observatory endpoints.

Unit test have been created matching the API endpoints spec.

## Types of changes

## Checklist:

      and relies on stable APIs.
  - public API documentation coming soon.

### DIFF
--- a/observatory.go
+++ b/observatory.go
@@ -1,0 +1,406 @@
+package cloudflare
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+	"strconv"
+	"strings"
+	"time"
+)
+
+var (
+	ErrMissingObservatoryUrl    = errors.New("missing required page url")
+	ErrMissingObservatoryTestID = errors.New("missing required test id")
+)
+
+// ObservatoryPage describes all the tests for a web page.
+type ObservatoryPage struct {
+	URL               string                `json:"url"`
+	Region            labeledRegion         `json:"region"`
+	ScheduleFrequency string                `json:"scheduleFrequency"`
+	Tests             []ObservatoryPageTest `json:"tests"`
+}
+
+// ObservatoryPageTest describes a single test for a web page.
+type ObservatoryPageTest struct {
+	ID                string                      `json:"id"`
+	Date              time.Time                   `json:"date"`
+	URL               string                      `json:"url"`
+	Region            labeledRegion               `json:"region"`
+	ScheduleFrequency *string                     `json:"scheduleFrequency"`
+	MobileReport      ObservatoryLighthouseReport `json:"mobileReport"`
+	DesktopReport     ObservatoryLighthouseReport `json:"desktopReport"`
+}
+
+// labeledRegion describes a region the test was run in.
+type labeledRegion struct {
+	Value string `json:"value"`
+	Label string `json:"label"`
+}
+
+// ObservatorySchedule describe a test schedule.
+type ObservatorySchedule struct {
+	URL       string `json:"url"`
+	Region    string `json:"region"`
+	Frequency string `json:"frequency"`
+}
+
+// ObservatoryLighthouseReport describes the web vital metrics result.
+type ObservatoryLighthouseReport struct {
+	PerformanceScore int    `json:"performanceScore"`
+	State            string `json:"state"`
+	DeviceType       string `json:"deviceType"`
+	// TTFB is time to first byte
+	TTFB int `json:"ttfb"`
+	// FCP is first contentful paint
+	FCP int `json:"fcp"`
+	// LCP is largest contentful pain
+	LCP int `json:"lcp"`
+	// TTI is time to interactive
+	TTI int `json:"tti"`
+	// TBT is total blocking time
+	TBT int `json:"tbt"`
+	// SI is speed index
+	SI int `json:"si"`
+	// CLS is cumulative layout shift
+	CLS   float64          `json:"cls"`
+	Error *lighthouseError `json:"error,omitempty"`
+}
+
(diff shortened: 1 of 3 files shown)


observatory.go
```diff
@@ -0,0 +1,403 @@
...
+	URL               string                `json:"url"`
+	Region            labeledRegion         `json:"region"`
+	ScheduleFrequency string                `json:"scheduleFrequency"`
+	Tests             []ObservatoryPageTest `json:"tests"`
+}
+
+// ObservatoryPageTest describes a single test for a web page.
+type ObservatoryPageTest struct {
+	ID                string                      `json:"id"`
+	Date              time.Time                   `json:"date"`
```
jacobbednarz: https://github.com/cloudflare/cloudflare-go/blob/master/docs/conventions.md#timetime

observatory.go
```diff
@@ -0,0 +1,403 @@
...
+		return nil, fmt.Errorf("%s: %w", errUnmarshalError, err)
+	}
+	return r.Result, nil
+}
+
+type GetObservatoryPageTrendParams struct {
+	URL        string
+	Region     string     `url:"region"`
+	DeviceType string     `url:"deviceType"`
+	Start      time.Time  `url:"start"`
```
jacobbednarz: https://github.com/cloudflare/cloudflare-go/blob/master/docs/conventions.md#timetime

observatory.go
```diff
@@ -0,0 +1,403 @@
...
+}
+
+// GetObservatoryPageTrend returns a the trend of web vital metrics for a page in a specific region.
+//
+// API reference: https://api.cloudflare.com/#speed-list-page-trend
+func (api *API) GetObservatoryPageTrend(ctx context.Context, rc *ResourceContainer, params GetObservatoryPageTrendParams) (*ObservatoryPageTrend, error) {
+	if params.URL == "" {
+		return nil, ErrMissingObservatoryUrl
+	}
+	v := url.Values{}
```
jacobbednarz: i think majority of this can be removed if you use the [`buildURI`](https://github.com/cloudflare/cloudflare-go/blob/f968533e2861e3d84c52e6e5f194c31038604580/utils.go#L13) helper and `url` struct tags (example: https://github.com/cloudflare/cloudflare-go/blob/f968533e2861e3d84c52e6e5f194c31038604580/zone_hold.go#L30)

observatory.go
```diff
@@ -0,0 +1,403 @@
...
+
+// CreateObservatoryPageTest starts a test for a page in a specific region.
+//
+// API reference: https://api.cloudflare.com/#speed-create-test
+func (api *API) CreateObservatoryPageTest(ctx context.Context, rc *ResourceContainer, params CreateObservatoryPageTestParams) (*ObservatoryPageTest, error) {
+	if params.URL == "" {
+		return nil, ErrMissingObservatoryUrl
+	}
+	uri := fmt.Sprintf("/zones/%s/speed_api/pages/%s/tests", rc.Identifier, url.PathEscape(params.URL))
+	res, err := api.makeRequestContext(ctx, http.MethodPost, uri, struct {
```
jacobbednarz: i'd advise using a dedicated field on the struct for this (inlining make things more verbose and harder to extract).

observatory.go
```diff
@@ -0,0 +1,403 @@
...
+	var r ObservatoryPageTestResponse
+	err = json.Unmarshal(res, &r)
+	if err != nil {
+		return nil, fmt.Errorf("%s: %w", errUnmarshalError, err)
+	}
+	return &r.Result, nil
+}
+
+type CreateObservatoryScheduledPageTestParams struct {
+	URL       string
```
jacobbednarz: i think you'll want this to prevent it from ending up in the URL

```suggestion
	URL       string `url:"-"`
```

observatory.go
```diff
@@ -0,0 +1,403 @@
+package cloudflare
+
+import (
+	"context"
+	"encoding/json"
```
jacobbednarz: we've swapped to `github.com/goccy/go-json` for json parsing/marshaling over the stdlib for a few reasons (lint coming soon!)

jacobbednarz:
awesome start. just a few style/convention nitpicks but looking very close!

observatory.go
broswen: I'm not sure I understand what you are talking about.
Do you mean making a `Request` struct held by the `CreateObservatoryPageTestParams` struct that is passed as the request body?

jacobbednarz: i would expect to be able to pass `params` to `makeRequestContext` as is and have it marshal the correct payload. i think you're nearly there though; we just need the `json` struct tags to be added to  `CreateObservatoryPageTestParams`.

```go

type CreateObservatoryPageTestParams struct {
	URL    string `json:"-"`
	Region string `json:"region"`
}
```

then this line becomes 

```
res, err := api.makeRequestContext(ctx, http.MethodPost, uri, params)
```

jacobbednarz: updated in [dfce46a](https://github.com/cloudflare/cloudflare-go/pull/1401/commits/dfce46a404fca4193f4694e7b1b3bce5d628c67c)

observatory.go
```diff
@@ -0,0 +1,388 @@
...
+}
+
+// GetObservatoryPageTest returns a specific test for a page.
+//
+// API reference: https://api.cloudflare.com/#speed-get-test
+func (api *API) GetObservatoryPageTest(ctx context.Context, rc *ResourceContainer, params GetObservatoryPageTestParams) (*ObservatoryPageTest, error) {
+	if params.URL == "" {
+		return nil, ErrMissingObservatoryUrl
+	}
+	if params.URL == "" {
```
jacobbednarz:
```suggestion
	if params.TestID == "" {
```

PR #1403, opened 2023-09-14
AUTH-5428 added tags to access namespace

russellkerns:
## Description

Tags are a new resource to tie to an application.  This change allows for that CRUD on access tags as well as associating them with applications

Unit tests

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x] My code follows the code style of this project.
- [x ] My change requires a change to the documentation.
- [x ] I have updated the documentation accordingly.
- [x ] All new and existing tests passed.
- [ x] This change is using publicly documented in [cloudflare/api-schemas](https://github.com/cloudflare/api-schemas) 
      and relies on stable APIs.

### DIFF
--- a/access_tag.go
+++ b/access_tag.go
@@ -1,0 +1,85 @@
+package cloudflare
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/goccy/go-json"
+)
+
+type AccessTag struct {
+	Name     string `json:"name,omitempty"`
+	AppCount int    `json:"app_count,omitempty"`
+}
+
+type AccessTagListResponse struct {
+	Response
+	Result     []AccessTag `json:"result"`
+	ResultInfo `json:"result_info"`
+}
+
+type AccessTagResponse struct {
+	Response
+	Result AccessTag `json:"result"`
+}
+
+type ListAccessTagsParams struct{}
+
+type CreateAccessTagParams struct {
+	Name string `json:"name,omitempty"`
+}
+
+func (api *API) ListAccessTags(ctx context.Context, rc *ResourceContainer, params ListAccessTagsParams) ([]AccessTag, error) {
+	uri := buildURI(fmt.Sprintf("/%s/%s/access/tags", rc.Level, rc.Identifier), params)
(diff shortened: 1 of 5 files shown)


access_tag.go
```diff
@@ -0,0 +1,108 @@
...
+func (api *API) DeleteAccessTag(ctx context.Context, rc *ResourceContainer, tagName string) error {
+	uri := fmt.Sprintf("/%s/%s/access/tags/%s", rc.Level, rc.Identifier, tagName)
+	_, err := api.makeRequestContext(ctx, http.MethodDelete, uri, nil)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (api *API) UpdateAccessTag(ctx context.Context, rc *ResourceContainer, params UpdateAccessTagParams) (AccessTag, error) {
```
jacobbednarz: this does something however, i don't quite see what it is changing. the `params.Name` passed in is used for the old and new value. should there be a difference available to the values?

jacobbednarz:
a minor question about updating as i don't think it will do anything useful as is but otherwise 👍 .

access_tag.go
russellkerns: we're gonna have to do something hacky to make this work because I'm realizing with TF you're not gonna be specifying the url you're making the request to so we need to differentiate between the name that is the current name of the tag and what the updated name will be since the name of the tag is also its unique identifier in the api url

russellkerns: In retrospect we're gonna get rid of the update functionality on tags because you would also have to update all the tags in the apps that have said tags anyway (the api does this inherently but that doesn't fix tf state)

jacobbednarz: it won't just be terraform that is impacted here, it will also be the SDKs. reason being is that we are using a modifiable value in the URL and the value to update. i don't recommend this particular API design but since we have it, the `Update` method can have `TagName` and `UpdatedTagName` to differentiate the two (which makes both values addressable without conflicts).

russellkerns: updated to make the update function a noop

PR #1410, opened 2023-09-26
fix: incorrect data type in UpdateCustomNameserverZoneMetadataParams

rwhelan:
## Description

Incorrect data type in UpdateCustomNameserverZoneMetadataParams struct.  Using as-is would return a 500.

the `ns_set` field is an `int` but the struct is currently setup with a `string`

Found out the API would not accept the request without being an int.

## Types of changes

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/.changelog/1410.txt
+++ b/.changelog/1410.txt
@@ -1,3 +1,3 @@
 ```release-note:bug
-cloudflare: fix incorrect data type in UpdateCustomNameserverZoneMetadataParams struct
+custom_nameservers: change `NSSet` from string to int to match API response
 ```
--- a/.changelog/1410.txt
+++ b/.changelog/1410.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+cloudflare: fix incorrect data type in UpdateCustomNameserverZoneMetadataParams struct
+```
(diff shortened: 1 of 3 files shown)


custom_nameservers_test.go
```diff
@@ -196,7 +196,6 @@ func TestAccountCustomNameserver_GetAccountCustomNameserverZoneMetadata(t *testi
 		w.Header().Set("content-type", "application/json")
 		fmt.Fprintf(w, `{
 			"result": {
-				"type": "account",
 				"ns_set": "1",
```
jacobbednarz: based on the type change here, this should be an int

custom_nameservers_test.go
```diff
@@ -208,7 +207,6 @@ func TestAccountCustomNameserver_GetAccountCustomNameserverZoneMetadata(t *testi
 
 	mux.HandleFunc("/zones/"+testZoneID+"/custom_ns", handler)
 	want := CustomNameserverZoneMetadata{
-		Type:    "account",
 		NSSet:   "1",
```
jacobbednarz: based on the type change here, this should be an int

.changelog/1410.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+cloudflare: fix incorrect data type in UpdateCustomNameserverZoneMetadataParams struct
```
jacobbednarz:
```suggestion
custom_nameservers: change `NSSet` from string to int to match API response
```

jacobbednarz:
thanks for this @rwhelan ✌️

PR #1412, opened 2023-09-28
fix double encoding of url

broswen:
## Description
This fixes a bug where `buildURI` would double encode forward slashes in the the `URL` parameter.

Let me know if there is a better way to handle this.

I have tested with a personal zone with URLs that include multiple path segments.
I updated the unit tests to use a URL that has a longer path with multiple segments.

## Types of changes

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/.changelog/1412.txt
+++ b/.changelog/1412.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+observatory: fix double url encoding
+```
--- a/observatory.go
+++ b/observatory.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/goccy/go-json"
+	"github.com/google/go-querystring/query"
 )
 
 var (
@@ -137,7 +138,9 @@ func (api *API) GetObservatoryPageTrend(ctx context.Context, rc *ResourceContain
 	if params.URL == "" {
 		return nil, ErrMissingObservatoryUrl
 	}
-	uri := buildURI(fmt.Sprintf("/zones/%s/speed_api/pages/%s/trend", rc.Identifier, url.PathEscape(params.URL)), params)
+	// cannot use buildURI because params.URL contains "/" that should be encoded and buildURI will double encode %2F into %252F
+	v, _ := query.Values(params)
+	uri := fmt.Sprintf("/zones/%s/speed_api/pages/%s/trend?%s", rc.Identifier, url.PathEscape(params.URL), v.Encode())
 	res, err := api.makeRequestContext(ctx, http.MethodGet, uri, nil)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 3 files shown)


jacobbednarz:
ah, i see what is happening here; the `v.Encode()` is double encoding it. while your fix works, it does make the usage a little unconventional and if we need to fix core bugs in the future with URL parsing (just like this one) we won't have a single chokepoint to patch it. 

instead, i'd prefer we look at a way to address this in `buildURI` itself. here is a a side by side comparison of the outputs with my idea for a fix - https://go.dev/play/p/8CE_hPEifE3. to make this backwards compatible, i'd probably want to introduce a third parameter of configuration options to `buildURI`.

```
type buildURIConfiguration struct {
	UnescapeQueryParams bool
}

func buildURI(path string, options interface{}, config buildURIConfiguration) string {
	v, _ := query.Values(options)

	queryParams := v.Encode()
	if config.UnescapeQueryParams {
		queryParams, _ = url.QueryUnescape(queryParams)
	}

	return (&url.URL{Path: path, RawQuery: queryParams}).String()
}
```

WDYT? i'd love to make this the default but i think it may potentially break some usages of it (but i could be wrong!)

broswen:
That seems like it could also be a problem, but specifically the issue isn't with double encoding a query param but with a path parameter.

The endpoints take the url as a path parameter. For example if the url was `example.com/subpath` the endpoint would be.
`https://api.cloudflare.com/client/v4
/zones/{zone_identifier}/speed_api/schedule/example.com%2Fsubpath`

Since `/` is a valid delimiter it doesn't get path escaped by `url.URL`. But if you path escape the URL beforehand it contains a `%2F` and `url.URL` attempts to escape the `%` a second time leaving you with `%252F`.

jacobbednarz:
ah sorry. I must have misread the sprintf as I thought this was an issue in the query parameter. 

if this is in the path, any reason why we need to be manually escaping it? any reason we don't just pass it though as is and have the server side handle any security concerns?

broswen:
Because a request the the endpoint 
`https://api.cloudflare.com/client/v4 /zones/{zone_identifier}/speed_api/schedule/example.com%2Fsubpath`
Is different from
`https://api.cloudflare.com/client/v4 /zones/{zone_identifier}/speed_api/schedule/example.com/subpath`

The endpoint routing gets confused and doesn't send it to the proper upstream.
Plus there are other endpoints such as the following which will break if the url path parameter isn't escaped.
`https://api.cloudflare.com/client/v4
/zones/{zone_identifier}/speed_api/pages/{url}/tests`

 `/pages/example.com%2Fsubpath/tests` vs `/pages/example.com/subpath/tests`

This just seems to be an annoyance with accepting a URL as a path parameter instead of query param or JSON body.

jacobbednarz:
👍 for now, let's leave this as is and when i'm back we'll have a chat about this. in a generated SDK world, this will be making your maintenance/life alot harder than what it needs to be so we'll see what we can come up with to minimise that.

PR #1424, opened 2023-10-19
Add Browser Isolation Non-identity onramp support

pierreca:
## Description

Add a `non_identity_enabled` JSON field to the Browser Isolation settings structure in the Zero trust account.

The unit test was updated to reflect the new fields.

This is the corresponding dashboard setting:
image

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly. NOTE: Feature already documented
      and relies on stable APIs. NOTE: OpenAPI source updated, awaiting roll out.

### DIFF
--- a/teams_accounts.go
+++ b/teams_accounts.go
@@ -47,8 +47,8 @@ type TeamsAccountSettings struct {
 }
 
 type BrowserIsolation struct {
-	UrlBrowserIsolationEnabled bool `json:"url_browser_isolation_enabled"`
-	NonIdentityEnabled         bool `json:"non_identity_enabled"`
+	UrlBrowserIsolationEnabled *bool `json:"url_browser_isolation_enabled,omitempty"`
+	NonIdentityEnabled         *bool `json:"non_identity_enabled,omitempty"`
 }
 
 type TeamsAntivirus struct {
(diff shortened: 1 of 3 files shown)


teams_accounts.go
```diff
@@ -48,6 +48,7 @@ type TeamsAccountSettings struct {
 
 type BrowserIsolation struct {
 	UrlBrowserIsolationEnabled bool `json:"url_browser_isolation_enabled"`
+	NonIdentityEnabled         bool `json:"non_identity_enabled"`
```
jacobbednarz: i would recommend updating both of these fields to match the convention at https://github.com/cloudflare/cloudflare-go/blob/master/docs/conventions.md#booleans to avoid future you some issues (at least in terraform land)

pierreca: @jacobbednarz oh I remember running into this before..
isn't that a breaking change though?

jacobbednarz: it is however:

- it mitigates a bunch of zero values elsewhere; and
- we're (technically) pre 1.0 of the library so this sort of thing is fine within semver

pierreca: should I fix the changelog to reflect this?

jacobbednarz: already done :) if you refresh you should see it now.

PR #1428, opened 2023-10-26
ip_access_rules: Add ListIPAccessRules api call

dkoston:
Adds List IP Access Rules API Call [docs](https://developers.cloudflare.com/api/operations/ip-access-rules-for-a-zone-list-ip-access-rules)

## Description

Provides access to an api endpoint not provided by this library currently

Yes, tested basic call

## Types of changes

## Checklist:

- [no] My change requires a change to the documentation.
- [n/a] I have updated the documentation accordingly.
      and relies on stable APIs.


ip_access_rules.go
```diff
@@ -0,0 +1,97 @@
...
+	ConfigurationTarget string                       `json:"configuration.target,omitempty"`
+	ConfigurationValue  string                       `json:"configuration.value,omitempty"`
+	Match               ListIpAccessRulesMatchOption `json:"match,omitempty"`
+	Mode                RulesModeOption              `json:"mode,omitempty"`
+	Notes               string                       `json:"notes,omitempty"`
+}
+
+type ListIpAccessRulesParams struct {
+	Direction         string                       `url:"direction,omitempty"`
+	EGSPaginationJSON EGSPaginationJSON            `url:"egs-pagination.json,omitempty"`
```
jacobbednarz: the `egs-pagination.json` is an error in the schema.
```suggestion
```

ip_access_rules.go
```diff
@@ -0,0 +1,97 @@
...
+	Filters           ListIpAccessRulesFilters     `url:"filters,omitempty"`
+	Order             ListIpAccessRulesOrderOption `url:"order,omitempty"`
+	Page              int                          `url:"page,omitempty"`
+	PerPage           int                          `url:"per_page,omitempty"`
```
jacobbednarz:
```suggestion
	PaginationOptions
```

ip_access_rules.go
```diff
@@ -0,0 +1,97 @@
...
+	RuleModeJsChallenge      RulesModeOption              = "js_challenge"
+	RuleModeManagedChallenge RulesModeOption              = "managed_challenge"
+	RuleModeWhitelist        RulesModeOption              = "whitelist"
+)
+
+type EGSPaginationJSON struct {
+	Page    int `url:"page,omitempty"`
+	PerPage int `url:"per_page,omitempty"`
+}
+
```
jacobbednarz:
```suggestion
```

ip_access_rules.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/goccy/go-json"
+)
+
+type ListIpAccessRulesOrderOption string
+type ListIpAccessRulesMatchOption string
+type RulesModeOption string
+
+const (
+	ConfigurationTarget      ListIpAccessRulesOrderOption = "configuration.target"
+	ConfigurationValue       ListIpAccessRulesOrderOption = "configuration.value"
```
jacobbednarz: these aren't strings; it is nested object notation.

ip_access_rules.go
```diff
@@ -0,0 +1,97 @@
...
+	"context"
+	"fmt"
+	"net/http"
+
+	"github.com/goccy/go-json"
+)
+
+type ListIpAccessRulesOrderOption string
+type ListIpAccessRulesMatchOption string
+type RulesModeOption string
```
jacobbednarz: we need to prefix this to prevent it polluting the global namespace

```suggestion
type IPAccessRulesModeOption string
```

ip_access_rules.go
```diff
@@ -0,0 +1,97 @@
...
+	RuleModeWhitelist        RulesModeOption              = "whitelist"
+)
+
+type EGSPaginationJSON struct {
+	Page    int `url:"page,omitempty"`
+	PerPage int `url:"per_page,omitempty"`
+}
+
+type ListIpAccessRulesFilters struct {
+	ConfigurationTarget string                       `json:"configuration.target,omitempty"`
```
jacobbednarz: same as https://github.com/cloudflare/cloudflare-go/pull/1428/files#r1373906139

ip_access_rules.go
```diff
@@ -0,0 +1,97 @@
...
+
+// ListIpAccessRules
+//
+// Fetches IP Access rules of a zone. You can filter the results using several optional parameters.
+//
+// API reference: https://developers.cloudflare.com/api/operations/ip-access-rules-for-a-zone-list-ip-access-rules
+func (api *API) ListIpAccessRules(ctx context.Context, rc *ResourceContainer, params ListIpAccessRulesParams) ([]IpAccessRule, error) {
+
+	if rc.Identifier == "" {
+		return []IpAccessRule{}, ErrMissingZoneID
```
jacobbednarz: `ErrMissingZoneID` is incorrect here as the IP access API is accessible on both the zone and account level. you'll need a more generic error like `ErrMissingResourceIdentifier`

ip_access_rules.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	uri := buildURI(fmt.Sprintf("/zones/%s/firewall/access_rules/rules", rc.Identifier), params)
```
jacobbednarz:
```suggestion
	uri := buildURI(fmt.Sprintf("/%s/%s/firewall/access_rules/rules", rc.Level, rc.Identifier), params)
```

ip_access_rules.go
dkoston: Hmmm. That's not what the API docs say. Maybe they are wrong:
Screenshot 2023-10-27 at 09 08 08




https://developers.cloudflare.com/api/operations/ip-access-rules-for-a-zone-list-ip-access-rules

ip_access_rules.go
dkoston: Yea, I'm surprised there isn't a package for each resource to avoid conflicts and make things easier to find.

Kind of tough when there are 25+ `CreateX()` functions in the main package, no?

Maybe for v2 it could be `cloudflare.zones.Operation()`, `cloudflare.firewallRules.Operation()` like the Python library?

ip_access_rules.go
dkoston: looks like also available on user level. Adjusted thanks!

ip_access_rules.go
```diff
@@ -0,0 +1,93 @@
...
+// ListIpAccessRules
+//
+// Fetches IP Access rules of a zone/user/account. You can filter the results using several optional parameters.
+//
+// API references:
+//   - https://developers.cloudflare.com/api/operations/ip-access-rules-for-a-user-list-ip-access-rules
+//   - https://developers.cloudflare.com/api/operations/ip-access-rules-for-a-zone-list-ip-access-rules
+//   - https://developers.cloudflare.com/api/operations/ip-access-rules-for-an-account-list-ip-access-rules
+func (api *API) ListIpAccessRules(ctx context.Context, rc *ResourceContainer,
+	params ListIpAccessRulesParams) ([]IpAccessRule, *ResultInfo, error) {
```
dkoston: adjusted to also return *ResultInfo since there can be many pages

ip_access_rules.go
jacobbednarz: the documentation is using incorrect notation. if you look at the examples (and the live API calls), it is a nested JSON block.

ip_access_rules.go
jacobbednarz: this is already in the works; see https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md for details.

ip_access_rules.go
dkoston: @jacobbednarz sorry for slow response, didn't see email from github. I really don't see how that's possible. This is a `string` in which you are specifying how to order the results. Nested JSON is part of the IP access rule itself but not the `?order=X` param, unless I'm really missing something here.

ip_access_rules.go
dkoston: You may consider using typed functional options in addition to defining specific request structs in the new version so that if more options are created or needed across the library, you don't have to make breaking changes.

here's an example where I added the missing `per_page` option for custom hostnames in our version of this library (breaking change):

https://github.com/onrocketdotcom/cloudflare-go/pull/3/files#diff-432b285f81cb984bfd1ce00fa760be757c96784135a1b9ecd49d3b6c9a26522bR508

For example, this example in experimental.md would change from:

```
dParams := &cloudflare.DNSRecordListParams{
  AccountID: "d8e8fca2dc0f896fd7cb4cb0031ba249"
}
r, _, _ := c.DNSRecord.List(context.TODO(), dParams)
```

to

```
dParams := &cloudflare.DNSRecordListParams{
  AccountID: "d8e8fca2dc0f896fd7cb4cb0031ba249"
}
r, _, _ := c.DNSRecord.List(context.TODO(), dParams, WithPaginationOptions(), WithFilterOptions())
```

I don't see function definitions listed in expiremental.md so perhaps that's already what you're thinking.

ip_access_rules.go
jacobbednarz: ah yes, we're looking at two different things here.

dkoston:
fixed the lint issue

jacobbednarz:
your fork doesn't allow maintainers to send changes so here is a patch instead. `IP` should be uppercase in it's usage. here is a patch for it - https://gist.github.com/jacobbednarz/57245ff4c77be94127294a761fd9d00f

dkoston:
Thanks @jacobbednarz! Patch applied

PR #1454, opened 2023-12-01
Add support for Opt-In Split Tunnel Overlapping IPs

da-cf:
## Description
Add two new fields to the device settings policy. This is to add API support for the Opt-In Split Tunnel Overlapping IPs feature:
1. `lan_allow_minutes` (uint) used to set the amount of time a user can access a LAN subnet.
2. `lan_allow_subnet_size` (uint) the allowed subnet size for the lan 

Added to existing device settings policy tests.

## Types of changes

## Checklist:

      and relies on stable APIs.


devices_policy.go
```diff
@@ -57,6 +57,8 @@ type DeviceSettingsPolicy struct {
 	Default             bool              `json:"default"`
 	ExcludeOfficeIps    *bool             `json:"exclude_office_ips"`
 	Description         *string           `json:"description"`
+	LanAllowMinutes     *uint             `json:"lan_allow_minutes"`
```
jacobbednarz: given LAN is an acronym, we should uppercase the usage in all these cases much like we would for HTTP or TCP (`Lan_` => `LAN_`)

da-cf: Hey Jacob, thank you for the prompt review.

While I agree that this should have been `LAN_` from the start, most, if not all, of the teams working on this have already standardized with the `lan_allow_minutes` json field name.
I don't know if implementing this change is worth having a handful of people update their field name, but I'm interested to hear your thoughts.

Thanks again :)

jacobbednarz: 👋 my point is only referring to the struct name here which is what we are adding support for here; the json key is perfect.

da-cf: Oh that's perfect, then. Will implement :)

PR #1463, opened 2023-12-14
feat: added gateway rule notification settings

rexscaria:
New notification settings are added to the gateway rules as optional settings. This PR adds support for them in the library.

API DOC change at https://developers.cloudflare.com/api/operations/zero-trust-gateway-rules-create-zero-trust-gateway-rule

## Types of changes

- [ X] New feature (non-breaking change which adds functionality)

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/teams_rules.go
+++ b/teams_rules.go
@@ -63,6 +63,8 @@ type TeamsRuleSettings struct {
 
 	// Resolver policy settings.
 	DnsResolverSettings *TeamsDnsResolverSettings `json:"dns_resolvers,omitempty"`
+
+	NotificationSettings *TeamsNotificationSettings `json:"notification_settings"`
 }
 
 type TeamsGatewayUntrustedCertAction string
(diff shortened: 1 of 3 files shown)


jacobbednarz:
once the lint failures and struct namespacing is here, lgtm 🚢

teams_rules.go
```diff
@@ -77,6 +79,12 @@ type UntrustedCertSettings struct {
 	Action TeamsGatewayUntrustedCertAction `json:"action"`
 }
 
+type NotificationSettings struct {
```
jacobbednarz: until we have per product namespacing, we'll need to prefix this to avoid polluting the global scope.

```suggestion
type TeamsNotificationSettings struct {
```

teams_rules.go
```diff
@@ -77,6 +79,12 @@ type UntrustedCertSettings struct {
...
+	Enabled    bool   `json:"enabled"`
```
jacobbednarz: https://github.com/cloudflare/cloudflare-go/blob/master/docs/conventions.md#booleans

rexscaria: fixed. Although we don't have an unset state for this variable, I have made this into a pointer.

PR #1467, opened 2023-12-19
Add Zaraz

omarmosid:
This PR adds support for the Zaraz API

## Description

The primary goal is to be able to terraform Zaraz (RM-17254)

The following endpoints need to be covered.

1. `/config` (GET and PUT)
2. `/workflow` (GET and PUT)
3. `/publish` (POST)
4. `/history` (GET)
5. `/export` (GET)
6. `/default` (GET)

Please note that these aren't publicly documented as of today, but the plan is to add that too.

Yes. Added a unit test for every function. The API itself is stable and is used regularly via the dash.

## Types of changes

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## 0.85.0 (Unreleased)
 
+DEPENDENCIES:
+
+* deps: bumps github.com/go-git/go-git/v5 from 5.4.2 to 5.11.0 ([#1470](https://github.com/cloudflare/cloudflare-go/issues/1470))
+* deps: bumps github.com/urfave/cli/v2 from 2.26.0 to 2.27.0 ([#1471](https://github.com/cloudflare/cloudflare-go/issues/1471))
+* deps: bumps github.com/urfave/cli/v2 from 2.27.0 to 2.27.1 ([#1472](https://github.com/cloudflare/cloudflare-go/issues/1472))
+
 ## 0.84.0 (December 20th, 2023)
 
 ENHANCEMENTS:


omarmosid:
cc: @jonnyparris

Refaerds (approved):

zaraz.go
```diff
@@ -0,0 +1,234 @@
...
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/goccy/go-json"
+)
+
+type ZarazConfig = map[string]interface{}
```
jacobbednarz: is there a reason why we don't opt for a struct here?

zaraz.go
```diff
@@ -0,0 +1,234 @@
...
+	Result string `json:"result"`
+	Response
+}
+
+type ZarazPublishResponse struct {
+	Result string `json:"result"`
+	Response
+}
+
+type UpdateZarazConfigParams = map[string]interface{}
```
jacobbednarz: this should be a struct to prevent any unknown keys from being pushed into the parameters.

zaraz.go
```diff
@@ -0,0 +1,234 @@
...
+	var recordResp ZarazConfigResponse
+	err = json.Unmarshal(res, &recordResp)
+	if err != nil {
+		return ZarazConfigResponse{}, fmt.Errorf("%s: %w", errUnmarshalError, err)
+	}
+
+	return recordResp, nil
+}
+
+func (api *API) UpdateZarazConfig(ctx context.Context, rc *ResourceContainer, params UpdateZarazConfigParams) (ZarazConfigResponse, error) {
```
jacobbednarz: you definitely want `UpdateZarazConfigParams` to be typed here otherwise i can put anything in here without type constraints which may lead to API errors.

zaraz.go
```diff
@@ -0,0 +1,234 @@
...
+	var response ZarazWorkflowResponse
+	err = json.Unmarshal(res, &response)
+	if err != nil {
+		return ZarazWorkflowResponse{}, fmt.Errorf("%s: %w", errUnmarshalError, err)
+	}
+
+	return response, nil
+}
+
+func (api *API) UpdateZarazWorkflow(ctx context.Context, rc *ResourceContainer, workflowToUpdate string) (ZarazWorkflowResponse, error) {
```
jacobbednarz: the third parameter of all methods should be the `params` - see https://github.com/cloudflare/cloudflare-go/blob/master/docs/conventions.md#methods for examples.

zaraz.go
```diff
@@ -0,0 +1,234 @@
...
+	var response ZarazWorkflowResponse
+	err = json.Unmarshal(res, &response)
+	if err != nil {
+		return ZarazWorkflowResponse{}, fmt.Errorf("%s: %w", errUnmarshalError, err)
+	}
+
+	return response, nil
+}
+
+func (api *API) PublishZarazConfig(ctx context.Context, rc *ResourceContainer, description string) (ZarazPublishResponse, error) {
```
jacobbednarz: the third parameter of all methods should be the params - see https://github.com/cloudflare/cloudflare-go/blob/master/docs/conventions.md#methods for examples.

zaraz.go
```diff
@@ -0,0 +1,234 @@
...
+type ZarazPublishResponse struct {
+	Result string `json:"result"`
+	Response
+}
+
+type UpdateZarazConfigParams = map[string]interface{}
+
+type ZarazConfigRow struct {
+	ID          int64      `json:"id,omitempty"`
+	UserId      string     `json:"usedId,omitempty"`
```
jacobbednarz: `ID` is usually one of the words we uppercase (`UserId` => `UserID`)

zaraz.go
```diff
@@ -0,0 +1,234 @@
...
+	var response ZarazPublishResponse
+	err = json.Unmarshal(res, &response)
+	if err != nil {
+		return ZarazPublishResponse{}, fmt.Errorf("%s: %w", errUnmarshalError, err)
+	}
+
+	return response, nil
+}
+
+func (api *API) GetZarazConfigHistory(ctx context.Context, rc *ResourceContainer, params GetZarazConfigHistoryParams) ([]ZarazConfigRow, *ResultInfo, error) {
```
jacobbednarz: generally a "fetch all" type of method is prefixed with `List` to denote it is capable of returning multiple values - https://github.com/cloudflare/cloudflare-go/blob/master/docs/experimental.md#consistent-crud-method-signatures

PR #1474, opened 2024-01-02
Add Zaraz

omarmosid:
## Description

The primary goal is to be able to terraform Zaraz (RM-17254)

The following endpoints need to be covered.

/config (GET and PUT)
/workflow (GET and PUT)
/publish (POST)
/history (GET)
/export (GET)
/default (GET)

Yes. Added a unit test for every function. The API itself is stable and is used regularly via the dash.

## Types of changes

## Checklist:

      and relies on stable APIs.


zaraz.go
```diff
@@ -0,0 +1,417 @@
...
+	ZarazVersion  int64                    `json:"zarazVersion"`
+	Consent       ZarazConsent             `json:"consent,omitempty"`
+	DataLayer     *bool                    `json:"dataLayer,omitempty"`
+	Dlp           []any                    `json:"dlp,omitempty"`
+	HistoryChange *bool                    `json:"historyChange,omitempty"`
+	Settings      ZarazConfigSettings      `json:"settings,omitempty"`
+	Variables     map[string]ZarazVariable `json:"variables,omitempty"`
+}
+
+type UpdateZarazWorkflowParams struct{}
```
jacobbednarz: this needs to be updated with a field for usage (instead of a `string`)

zaraz.go
```diff
@@ -0,0 +1,417 @@
...
+
+type PublishZarazConfigParams struct{}
```
jacobbednarz: this needs to be updated with a field for usage (instead of a `string`)

PR #1480, opened 2024-01-17
Fixes update seats with wrong body: expected array, sent single object

### LINKED ISSUE #1479: UpdateAccessUsersSeats does not work because API expects an Array, and method sends a single object
### Confirmation

### cloudflare-go version

v0.86.0

### Go environment

GO111MODULE=''
GOARCH='amd64'
GOBIN=''
GOCACHE='/home/auyer/.cache/go-build'
GOENV='/home/auyer/.config/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='amd64'
GOHOSTOS='linux'
GOINSECURE=''
GOMODCACHE='/home/auyer/go/pkg/mod'
GONOPROXY=''
GONOSUMDB=''
GOOS='linux'
GOPATH='/home/auyer/go'
GOPRIVATE=''
GOPROXY='https://proxy.golang.org,direct'
GOROOT='/usr/lib/go'
GOSUMDB='sum.golang.org'
GOTMPDIR=''
GOTOOLCHAIN='auto'
GOTOOLDIR='/usr/lib/go/pkg/tool/linux_amd64'
GOVCS=''
GOVERSION='go1.21.6'
GCCGO='gccgo'
GOAMD64='v1'
AR='ar'
CC='gcc'
CXX='g++'
CGO_ENABLED='1'
GOMOD='/home/auyer/code/cloudflare-go/go.mod'
GOWORK=''
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
PKG_CONFIG='pkg-config'
GOGCCFLAGS='-fPIC -m64 -pthread -Wl,--no-gc-sections -fmessage-length=0 -ffile-prefix-map=/tmp/go-build3957979889=/tmp/go-build -gno-record-gcc-switches'

### Expected output

```
{
	"errors": [],
	"messages": [],
	"result": [
			{
		"access_seat": false,
		"created_at": "2014-01-01T05:20:00.12345Z",
		"gateway_seat": false,
		"seat_uid": "seat_id",
		"updated_at": "2014-01-01T05:20:00.12345Z"
			}
	],
	"success": true,
	"result_info": {
			"count": 1,
			"page": 1,
			"per_page": 20,
			"total_count": 2000
	}
}
```

### Actual output

```
{
  “result”: null,
  “success”: false,
  “errors”: [
  {
    “code”: 12074,
    “message”: “access.api.error.invalid_user_seat”
    }
  ],
  “messages”:...
}
```

```
userSeatRequest := cloudflare.UpdateAccessUserSeatParams{
  SeatUID: user.SeatUID, 
  AccessSeat: cloudflare.BoolPtr(false), 
  userSeatRequest.GatewaySeat: cloudflare.BoolPtr(false)
}
removeSeatResponse, err := api.UpdateAccessUserSeat(ctx, account, userSeatRequest)
fmt.Println(removeSeatResponse, err)
```

### Steps to reproduce

1 Get a valid Seat ID
2. send the request 

### References

https://developers.cloudflare.com/api/operations/zero-trust-seats-update-a-user-seat#request-body

auyer:
## Description

The expected body is an array. Thus I fixes the body in the existing method, and added a method for multiple UserSeeats.
I can also remove the new method, and make the existing one take a list.
83c77693e654a05fbeb7338cdff8da474fe38fc9 fixes #1479

Yes. I made requests using this, and wrote tests to cover it.

## Types of changes

## Checklist:

- [?] I have updated the documentation accordingly. (not sure if I should add docs elsewhere) 
      and relies on stable APIs.


access_seats.go
```diff
@@ -53,6 +53,42 @@ func (api *API) UpdateAccessUserSeat(ctx context.Context, rc *ResourceContainer,
...
+		return []AccessUpdateAccessUserSeatResult{}, fmt.Errorf("%s: %w", errUnmarshalError, err)
+	}
+
+	return updateAccessUserSeatResponse.Result, nil
+}
+
+// UpdateAccessUsersSeats updates many Access User Seats.
+//
+// API documentation: https://developers.cloudflare.com/api/operations/zero-trust-seats-update-a-user-seat
+func (api *API) UpdateAccessUsersSeats(ctx context.Context, rc *ResourceContainer, params []UpdateAccessUserSeatParams) ([]AccessUpdateAccessUserSeatResult, error) {
```
jacobbednarz: the `params` here should match the [conventions](https://github.com/cloudflare/cloudflare-go/blob/master/docs/conventions.md) so `UpdateAccessUsersSeatsParams` and be a dedicated struct.

auyer: Ok!
Could it be a type alias like `type UpdateAccessUsersSeatsParams []UpdateAccessUserSeatParams` referencing the struct from the other request ? Or does it need to be a completely new struct ?

Its worth to mention that they are both calling the same API method, and the only reason to not change the original method is to not break compatibility.

But now that I think about it, a variadic input would also work without breaking compatibility and without needing to add a new function. 
It would allow the users to call it with only one `UpdateAccessUserSeatParams` entries (as they do now), with multiple entries, or even with an array.

Ex: 
```go
func (api *API) UpdateAccessUserSeat(ctx context.Context, rc *ResourceContainer, params ...UpdateAccessUserSeatParams) ([]AccessUpdateAccessUserSeatResult, error) {
```
It could be used like this:
```go
// single entry
result, err := client.UpdateAccessUserSeat(
    ctx, AccountIdentifier(), UpdateAccessUserSeatParams{}
    )

// AND
// multiple entries
Result, err := client.UpdateAccessUserSeat(
    ctx, AccountIdentifier(), UpdateAccessUserSeatParams{}, UpdateAccessUserSeatParams{}
    )

// AND
// array into variadic 
Result, err := client.UpdateAccessUserSeat(
    ctx, AccountIdentifier(), []UpdateAccessUserSeatParams{}{}...
    )
```

Tell me what do you think! Thanks

jacobbednarz: separate struct please. we have it in place to allow adding fields without needing to have as many breaking changes and it allows the two methods to change independently.

we have conventions for (at most) 3 parameters in the method signatures to prevent really long methods.

auyer: I believe its up the the conventions now!
Thanks !
Let me know if you need anything else.

PR #1484, opened 2024-01-18
flarectl: fix duplicate alias in zone subcommand

### LINKED ISSUE #1465: Zone command group doesn't work in flarectl amd64 build of v0.83.0 github release
### Confirmation

### cloudflare-go version

v.0.83.0

### Go environment

N/A

### Expected output

```
$ ./flarectl zone list
                 ID                |    NAME     |     PLAN     | STATUS  
-----------------------------------+-------------+--------------+---------
  [REDACTED] | [REDACTED] | Free Website | active  
$ echo $?
0
```

### Actual output

```
$ ./flarectl zone list
$ echo $?
1
```

N/A

### Steps to reproduce

Download amd64 build from https://github.com/cloudflare/cloudflare-go/releases/tag/v0.83.0, untar and run any zone subcommand

### References

_No response_

thomasqueirozb:
Fixes: #1465 

## Description
check and certs were both aliased to "c". Changed certs alias from "c" to "ct".

This would've been way more easier to debug if a log were introduced before os.Exit(1). I'd be happy to add it here but I'm unsure if not logging is the intended behavior.

https://github.com/cloudflare/cloudflare-go/blob/c3ac69e2e73752b34b57aeba74e6a3fbb9cbc600/cmd/flarectl/flarectl.go#L715-L718

Just checked v0.82.0 where the zone command worked and `zone c` use the create command instead of the certs command. It looks like the `c` alias was changed from `certs` to `create` since the create command was added (d0f7110d28f7c9f322c104b64a432923074216c5 )

Yes, `flarectl zone` now works

Not appropriate

## Types of changes

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/.changelog/1484.txt
+++ b/.changelog/1484.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+flarectl: alias zone certs to "ct" instead of duplicating the "c" alias
+```
--- a/cmd/flarectl/flarectl.go
+++ b/cmd/flarectl/flarectl.go
@@ -225,7 +225,7 @@ func main() {
 				},
 				{
 					Name:    "certs",
-					Aliases: []string{"c"},
+					Aliases: []string{"ct"},
 					Action:  zoneCerts,
 					Usage:   "Custom SSL certificates for a zone",
 				},


jacobbednarz:
> This would've been way more easier to debug if a log were introduced before os.Exit(1). I'd be happy to add it here but I'm unsure if not logging is the intended behavior.

this is an unfortunate side effect of the design of the CLI tool here. all errors end up caught here and duplicated in the output since the commands will return an error.

thanks for fixing this one!

PR #1447, opened 2023-11-23
Multiple access application pages result in bad data returned

theFong:
## Description

When there are multiple pages of access applications, I observed incorrect create times, update times, and other objects being assigned for each access application object in the list returned. I didn't look into the root cause of why this was occurring because it seems like the code should work (and does for values), but the following fix addresses the problem by not reassigning the entire object on each page.

## Has your change been tested?

I have tested the fix in our own code base. I did not add a new test to cover this because it wasn't trivial to add a test for multiple pages. Happy to add if someone points me in the right direction.

## Types of changes

What sort of change does your code introduce/modify?

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] This change is using publicly documented in [cloudflare/api-schemas](https://github.com/cloudflare/api-schemas) 
      and relies on stable APIs.

[1]: https://help.github.com/articles/closing-issues-using-keywords/

PR #1493, opened 2024-02-03
Fixes more Paginations loops with value not being reset, and wrong value cloud be returned

auyer:
## Description

All this endpoints are susceptible to returning wrong data due to the Struct not getting reset for every iteration of the loop.
I fixed it by moving the variable declaration to inside the loop.

This is related to my other PR #1482 that fixed this for a single api endpoint.
I also found this PR that accomplished the same result in a different way #1016 . If you prefer this method, let me know. I can change them all to be in the same way.
And there is also this PR #1447 that is still missing the change-log. I cloud remove the 3 cases that PR fixes from mine, in case you want to merge that one too. 

Yes, `go test ./... --race`, and also some API calls. 
But as I mentioned in #1482, I was not able to create tests that fail without this change, but it happens in runtime.
Most of the tests in this repo do not cover pagination. 

## Types of changes

## Checklist:

      and relies on stable APIs.


auyer:
Hi @jacobbednarz , could you review this please?
Its the same change repeated in all paginations  I found that needed it.

Thanks!

jacobbednarz:
i haven't managed to set aside anytime for this lately but my concern is that there is no test coverage for this condition so it's hard to identify if this is actually addressing a problem. this hasn't been observed anywhere that we are using (in hundred of paginations) so i'm hesitant to blindly accept these ones. before we merge this, we need to either 1) find cases where this is a problem to confirm the fix or 2) identify a way to isolate this within a test case.

jacobbednarz:
v2 provides the option to handle the pagination automatically so this is no longer needed.

PR #1506, opened 2024-02-17
fix: domain list endpoint only accepts get requests

RazaGR:
https://developers.cloudflare.com/api/operations/registrar-domains-list-domains

### DIFF
--- a/.changelog/1506.txt
+++ b/.changelog/1506.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+fix: Fix request method to call domain list endpint from POST to GET
+```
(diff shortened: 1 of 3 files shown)


RazaGR:
@jacobbednarz any idea if registering a domain would be possible through api?  for non-enterprise users

.changelog/1506.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+fix: Fix request method to call domain list endpint from POST to GET
```
jacobbednarz:
```suggestion
registrar: Fix request method to call domain list endpoint from POST to GET
```

jacobbednarz (approved):

jacobbednarz:
> @jacobbednarz any idea if registering a domain would be possible through api? for non-enterprise users

it is not possible today; only update (https://developers.cloudflare.com/api/operations/registrar-domains-update-domain)

PR #1744, opened 2024-04-10
Add support for create/update individual ruleset rules

### LINKED ISSUE #1211: Add support for updating individual rules within Rulesets

v0.61.0

### Description

Started encountering an issue with one of our integrations, where we use the API to update transform rules in Cloudflare. Upon review of the API docs, I found that there are a pair of methods that can be used for updating individual transform rules, rather than the entire ruleset, and these methods do no appear to represented in the SDK:

https://developers.cloudflare.com/api/operations/account-rulesets-update-an-account-ruleset-rule
https://developers.cloudflare.com/api/operations/zone-rulesets-update-a-zone-ruleset-rule

Unrelated, but potentially relevant -  the method update the entire ruleset no longer works on one of my zones, but this PATCH method does. I'm not entirely sure why.

### Use cases

Updating a user-friendly URL to redirect to an updated versioned URL, without having to resubmit the entire ruleset for the phase.

```go
// Append to the bottom of rulesets.go

// UpdateZoneRulesetRule updates an individual rule within a ruleset for a zone.
//
// API reference: https://developers.cloudflare.com/api/operations/zone-rulesets-update-a-zone-ruleset-rule
func (api *API) UpdateZoneRulesetRule(ctx context.Context, zoneID, rulesetIdentifier, ruleIdentifier string, rule RulesetRule) (Ruleset, error) {
	return api.updateRulesetRule(ctx, ZoneRouteRoot, zoneID, rulesetIdentifier, ruleIdentifier, rule)
}

// UpdateAccountRulesetRule updates an individual rule within a ruleset for an account.
//
// API reference:  https://developers.cloudflare.com/api/operations/account-rulesets-update-an-account-ruleset-rule
func (api *API) UpdateAccountRulesetRule(ctx context.Context, accountID, rulesetIdentifier, ruleIdentifier string, rule RulesetRule) (Ruleset, error) {
	return api.updateRulesetRule(ctx, AccountRouteRoot, accountID, rulesetIdentifier, ruleIdentifier, rule)
}

// updateRulesetRule updates a ruleset based on the zone or account, the
// identifier and the rules.
func (api *API) updateRulesetRule(ctx context.Context, identifierType RouteRoot, identifier, rulesetIdentifier, ruleIdentifier string, rule RulesetRule) (Ruleset, error) {
	uri := fmt.Sprintf("/%s/%s/rulesets/%s/rules/%s", identifierType, identifier, rulesetIdentifier, ruleIdentifier)
	res, err := api.makeRequestContext(ctx, http.MethodPatch, uri, rule)
	if err != nil {
		return Ruleset{}, err
	}

	result := GetRulesetResponse{}
	if err := json.Unmarshal(res, &result); err != nil {
		return Ruleset{}, fmt.Errorf("%s: %w", errUnmarshalError, err)
	}

	return result.Result, nil
}
```

### References

_No response_

tcinbis:
## Description

This PR extends the `ruleset.go` and adds two new methods `CreateRulesetRule` and `UpdateRulesetRule`. The code was only slightly modified from the corresponding ruleset methods. This when merged closes #1211.

The changes have been tested both using automated tests run locally (everything passing) and against a live CF Zone WAF API (**not** account API). Further, the new methods are also covered by two new tests.

## Types of changes

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/.changelog/1744.txt
+++ b/.changelog/1744.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+rulesets: Add support for create/update calls for individual ruleset rules
+```
--- a/rulesets.go
+++ b/rulesets.go
@@ -733,6 +733,10 @@ type UpdateRulesetParams struct {
 	Rules       []RulesetRule `json:"rules"`
 }
 
+type UpdateRulesetRuleParams struct {
+	RulesetRule
+}
+
 type UpdateEntrypointRulesetParams struct {
 	Phase       string        `json:"-"`
 	Description string        `json:"description,omitempty"`
@@ -830,6 +834,25 @@ func (api *API) UpdateRuleset(ctx context.Context, rc *ResourceContainer, params
 
 	uri := fmt.Sprintf("/%s/%s/rulesets/%s", rc.Level, rc.Identifier, params.ID)
 	res, err := api.makeRequestContext(ctx, http.MethodPut, uri, params)
+	if err != nil {
+		return Ruleset{}, err
+	}
+
+	result := UpdateRulesetResponse{}
+	if err := json.Unmarshal(res, &result); err != nil {
+		return Ruleset{}, fmt.Errorf("%s: %w", errUnmarshalError, err)
+	}
+
+	return result.Result, nil
+}
+
+func (api *API) UpdateRulesetRule(ctx context.Context, rc *ResourceContainer, rulesetID string, params UpdateRulesetRuleParams) (Ruleset, error) {
+	if params.ID == "" {
+		return Ruleset{}, ErrMissingResourceIdentifier
+	}
+
+	uri := fmt.Sprintf("/%s/%s/rulesets/%s/rules/%s", rc.Level, rc.Identifier, rulesetID, params.ID)
+	res, err := api.makeRequestContext(ctx, http.MethodPatch, uri, params)
 	if err != nil {
 		return Ruleset{}, err
 	}
--- a/rulesets.go
+++ b/rulesets.go
@@ -733,6 +733,10 @@ type UpdateRulesetParams struct {
 	Rules       []RulesetRule `json:"rules"`
 }
 
+type CreateRulesetRuleParams struct {
+	RulesetRule
+}
+
 type UpdateRulesetRuleParams struct {
 	RulesetRule
 }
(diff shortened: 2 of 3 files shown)


jacobbednarz:
thanks for the PR! as v2 is just around the corner, instead of merging this, can you see if [ruleset.Rules](https://github.com/cloudflare/cloudflare-go/blob/v2/rulesets/rule.go) covers what you're after here?

tcinbis:
thanks for the hint I will definitely check out v2 now. When you say "around the corner" what are we talking about in terms of time? I know estimates are a tricky but just to get a rough understanding.

jacobbednarz:
you can use it today, see https://github.com/cloudflare/cloudflare-go/discussions/1538

it's just waiting for us to clean up a couple of UX improvements and slap a GA label on it.

tcinbis:
Okay so I looked into v2 and while it is great, I am afraid that the functionality this PR would introduce is either missing or then  I can't find it.

### What is the functionality I am looking for?
 
We need to be able to CRUD individual rules in a WAF ruleset without updating the entire ruleset.
Otherwise, to my understanding we could be running into race conditions in which two individual clients are performing an ruleset update (please correct me if that is wrong).

jacobbednarz:
the link i provided above is for managing individual ruleset rules - the same as the PR you've proposed. you can confirm this by looking at the HTTP method and the path called. 

as for your issue or concurrency, it exists in a ruleset and rule and neither will mitigate it because the backing service doesn't offer concurrency flags or a locking mechanism. you can still have two HTTP calls to the same resource which will result in a race condition; that is something you'll need to mitigate within your own coordination.

tcinbis:
Maybe you could point me to the methods that can be used to updating the zone WAF rules, since I only manage to find the methods that use the accounts endpoint, but not he zone endpoints (searching for `zones/%s/rulesets` does not return any results for me).

Thank you very much!

jacobbednarz:
see [ruleset.Rules](https://github.com/cloudflare/cloudflare-go/blob/v2/rulesets/rule.go) - if you provide an account identifier, it uses that, otherwise if you provide a zone identifier, it will use that instead.

jacobbednarz:
closing as v2 has this functionality today and you can use them alongside one another if needed.

PR #1939, opened 2024-05-03
add Position to RulesetRule

### LINKED ISSUE #1938: RulesetRule missing position arguments
### Confirmation

### cloudflare-go version

v0.x, v2.x

### Expected output

The ability to update a RulesetRule position by PATCH:

https://developers.cloudflare.com/ruleset-engine/rulesets-api/update-rule/#change-the-order-of-a-rule-in-a-ruleset

Looks to be missing from API docs too.

RulesetRule struct is missing Position field. https://github.com/cloudflare/cloudflare-go/blob/1096abb50b128c8c55e952a8d7cd1f894181da61/rulesets.go#L648-L663

vroy:
This will allow to update position of a specific rule in PATCH:

* fixes: https://github.com/cloudflare/cloudflare-go/issues/1938
* https://developers.cloudflare.com/ruleset-engine/rulesets-api/update-rule/#change-the-order-of-a-rule-in-a-ruleset

## Types of changes

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/.changelog/1939.txt
+++ b/.changelog/1939.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+rulesets: add position to ruleset rule
+```
--- a/rulesets.go
+++ b/rulesets.go
@@ -660,6 +660,7 @@ type RulesetRule struct {
 	RateLimit              *RulesetRuleRateLimit              `json:"ratelimit,omitempty"`
 	ExposedCredentialCheck *RulesetRuleExposedCredentialCheck `json:"exposed_credential_check,omitempty"`
 	Logging                *RulesetRuleLogging                `json:"logging,omitempty"`
+	Position               *RulesetRulePosition               `json:"position,omitempty"`
 }
 
 // RulesetRuleRateLimit contains the structure of a HTTP rate limit Ruleset Rule.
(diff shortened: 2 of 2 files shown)


jacobbednarz:
thanks for the PR however, the API schema documentation for `position` is missing so we cannot add it until that is resolved by the service team.

once that is added, it will automatically appear in 2.x.

jacobbednarz:
apologies - i was looking at the wrong branch. this does already exist in the API schemas and the 2.x SDK. see https://github.com/cloudflare/cloudflare-go/blob/v2/rulesets/rule.go#L5341

as it exists in 2.x and isn't used elsewhere in the v0 library, i think we're fine to close this one out and recommend using 2.x instead of backporting this one.

vroy:
Thanks, somehow missed when searching the v2 branch.

We'll certainly be looking at adopting 2.x SDK but until it becomes the main branch we're hesitant. Can this really not be backported?

Screenshot 2024-05-06 at 9 47 39 AM

jacobbednarz:
as that release note mentions, it's stable for majority of projects. we are leaving that warning in place for now as we are still cleaning up type names, etc. which are technically breaking changes and folks should be aware of it.

we have no timeline on when we switch the primary branch (nor should it matter given that is a GitHub-ism; you can use any release with confidence that it will largely work). 

we have other tooling that is in the process of porting over to the new SDKs and until those are complete, the legacy SDK will remain as is today but we are not extending the current footprint _except_ where those tools depend on the legacy SDK (see examples of recent zero trust changes for Terraform). `rule.position` values are not used in downstream tooling today, so we don't have plans to backport and support it.

if you really want this but also want to stick onto v0, you can run these libraries side by side. check out cloudflare/terraform-provider-cloudflare#3262 if you'd like some prior art to use.

PR #1944, opened 2024-05-05
fix: use correct name when running dns create-or-update

OhMyMndy:
When running `flarectl dns create-or-update` the update fails because it uses `name + '.' + zone` instead of just `name`

## Description

Change the name which is used to fetch the DNS record when it needs to be updated

Yes, tested this with my previously failing command `flarectl dns create-or-update --zone mydomain.com --name '*.mydomain.com' --content 127.0.0.1 --type A`

## Types of changes

## Checklist:

      and relies on stable APIs.


cmd/flarectl/dns.go
```diff
@@ -83,7 +83,7 @@ func dnsCreateOrUpdate(c *cli.Context) error {
 		return err
 	}
 
-	records, _, err := api.ListDNSRecords(context.Background(), cloudflare.ZoneIdentifier(zoneID), cloudflare.ListDNSRecordsParams{Name: name + "." + zone})
+	records, _, err := api.ListDNSRecords(context.Background(), cloudflare.ZoneIdentifier(zoneID), cloudflare.ListDNSRecordsParams{Name: name})
```
OhMyMndy, line 86: On line 117, the record itself is create with just `Name: name`, not zone appended to it

jacobbednarz:
thanks for the PR, however, i'm hesitant to merge this as i think the intended use here is to omit the zone name from your `--name` value. merging this would break the (from memory) intentional use where the zone name is not included in the `--name` value.

the CLI is in maintenance only mode so i don't want to change the functionality here but perhaps a better in between is to strip the zone name if it is supplied as part of `--name`?

OhMyMndy:
Omitting zone name from the name value works perfectly indeed!

PR #2722, opened 2024-07-03
Feat/add gateway categories

brennoo:
This pull request adds support for the `ListGatewayCategories` endpoint.

## Description
Our main driver is that we are looking into have this data available as a data resource in the terraform provider, please let me know if this is the right direction.

besides the test case I created [this script](https://gist.github.com/brennoo/e3e04bf76cebe1ccdbfe5b550627204a) and the output seems correct

## Types of changes

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/.changelog/2722.txt
+++ b/.changelog/2722.txt
@@ -1,3 +1,3 @@
 ```release-note:enhancement
-gateway: add ListGatewayCategories which returns all gateway categories.
+gateway_categories: add ListGatewayCategories which returns all gateway categories.
 ```
--- a/.changelog/2722.txt
+++ b/.changelog/2722.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+gateway: add ListGatewayCategories which returns all gateway categories.
+```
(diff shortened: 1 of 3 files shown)


.changelog/2722.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+gateway: add ListGatewayCategories which returns all gateway categories.
```
jacobbednarz:
```suggestion
gateway_categories: add ListGatewayCategories which returns all gateway categories.
```

jacobbednarz:
this functionality already exists in `v2` - is there a reason why we cannot use that instead?

brennoo:
> this functionality already exists in `v2` - is there a reason why we cannot use that instead?

no specific reason, I think when I was going thru the provider code I read somewhere that v2 should not be used

jacobbednarz:
you may have misunderstood; the Terraform SDKv2 shouldn't be used. the v2 Go library is where we are moving towards. 

we can merge this since it's already done but the Terraform provider should use v2 where possible (as that is what we're using going forward).

brennoo:
yep it is clear to me I mixed the V2s - if you think there is value in it feel free to merge - I will get back to the provider and when I start something I will use V2 Go library

thanks Jacob!

PR #2762, opened 2024-07-20
zones skips the first result if not paging

mabels:
The zones API call skips the initial result if used with no paging.

## Description

I miss the initial zones.

I'm too lazy to start a zones_test.go -- but a regression would be nice. 

## Types of changes

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/zones.go
+++ b/zones.go
@@ -92,6 +92,7 @@ func (s *ZonesService) List(ctx context.Context, params *ZoneListParams) ([]Zone
 		var zones []Zone
 		params.PerPage = defaultZonesPerPage
 		params.Page = 1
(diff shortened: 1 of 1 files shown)


jacobbednarz:
sorry for the confusion here but you're using part of the experimental client which has been superseded by v2.

if you're using 0.x, the correct call is [`ListZones`](https://github.com/cloudflare/cloudflare-go/blob/master/zone.go#L376).

if you're using 2.x, the correct call is [`client.Zones.List()`](https://github.com/cloudflare/cloudflare-go/blob/v2/zones/zone.go#L65).

i'll make a note to go back and clean up the old implementation now it is no longer being used.

PR #2793, opened 2024-07-29
feat: add flag to skip the app launcher landing pg

sha-e-256:
## Description

A new flag, `skip_app_launcher_login_page`, is now available which can be toggled to skip the App Launcher landing page a.k.a. the page that is shown to users when they visit their auth. domain i.e. `<team>.cloudflareaccess.com`.

This flag is documented for an app of the type `App Launcher App` at https://developers.cloudflare.com/api/operations/access-applications-add-an-application: 

> image

This flag is a part of Access and is an App Launcher customization property.

The unit test to test the creation of an App Launcher application has been updated to include the `skip_app_launcher_login_page` flag.

Screenshots N/A. 

## Types of changes

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/.changelog/2793.txt
+++ b/.changelog/2793.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+resource/access_application: add `skip_app_launcher_login_page` flag to skip the App Launcher landing page
+```
--- a/.changelog/2793.txt
+++ b/.changelog/2793.txt
@@ -1,3 +1,3 @@
 ```release-note:enhancement
-resource/access_application: add `skip_app_launcher_login_page` flag to skip the App Launcher landing page
+access_application: add `skip_app_launcher_login_page` flag to skip the App Launcher landing page
 ```
--- a/access_application.go
+++ b/access_application.go
@@ -250,11 +250,12 @@ type SaasApplication struct {
 }
 
 type AccessAppLauncherCustomization struct {
-	LandingPageDesign     AccessLandingPageDesign `json:"landing_page_design"`
-	LogoURL               string                  `json:"app_launcher_logo_url"`
-	HeaderBackgroundColor string                  `json:"header_bg_color"`
-	BackgroundColor       string                  `json:"bg_color"`
-	FooterLinks           []AccessFooterLink      `json:"footer_links"`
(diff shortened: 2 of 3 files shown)


access_application.go
```diff
@@ -250,11 +250,12 @@ type SaasApplication struct {
...
-	LogoURL               string                  `json:"app_launcher_logo_url"`
-	HeaderBackgroundColor string                  `json:"header_bg_color"`
-	BackgroundColor       string                  `json:"bg_color"`
-	FooterLinks           []AccessFooterLink      `json:"footer_links"`
+	LandingPageDesign        AccessLandingPageDesign `json:"landing_page_design"`
+	LogoURL                  string                  `json:"app_launcher_logo_url"`
+	HeaderBackgroundColor    string                  `json:"header_bg_color"`
+	BackgroundColor          string                  `json:"bg_color"`
+	FooterLinks              []AccessFooterLink      `json:"footer_links"`
+	SkipAppLauncherLoginPage bool                    `json:"skip_app_launcher_login_page"`
```
sha-e-256: I noticed that all the `bool` properties in this repo are bool pointers (`*bool`) with the `omitempty` struct tag; however, I believe the `skip_app_launcher_login_page` property should be `bool` and not a bool pointer since the API does not distinguish between the value being omitted and the value being set to false.
For example, the `auto_redirect_to_identity` property is also a bool pointer, but I believe it should just be a bool for the same reasons I mentioned above.
Please let me know how it is determined whether if the type should be `bool` or a bool pointer.

sha-e-256: I see that the linter enforces the usage of boolean pointers for properties of type `bool`: 

> use-pointers-for-booleans
> Booleans should always be represented as pointers in structs with an omitempty marshaling tag (most commonly as JSON). This ensures you can determine unset, false and truthy values.  

I'll go ahead and update my change to use bool pointers.

.changelog/2793.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:enhancement
+resource/access_application: add `skip_app_launcher_login_page` flag to skip the App Launcher landing page
```
jacobbednarz:
```suggestion
access_application: add `skip_app_launcher_login_page` flag to skip the App Launcher landing page
```

PR #2855, opened 2024-08-09
Add automatic ssl mode setting support

yj7o5:
## Description
This PR adds support for reading and modifying the new automatic ssl mode setting. 

Yes, there are tests in placed to verify both reading and writing the setting value. 

## Types of changes

## Checklist:

      and relies on stable APIs.


zone.go
```diff
@@ -865,6 +896,23 @@ func (api *API) UpdateZoneSSLSettings(ctx context.Context, zoneID string, sslVal
 	return r.Result, nil
 }
 
+// UpdateZoneAutomaticSSLSettingMode update information about Automatic SSL mode setting to the specified zone.
+//
+// API reference: https://api.cloudflare.com/[#-add-url-slug-here]
+func (api *API) UpdateZoneAutomaticSSLSettingMode(ctx context.Context, zoneID, mode string) (ZoneSSLSetting, error) {
```
jacobbednarz: we already have a method for fetching/updating zone settings (https://github.com/cloudflare/cloudflare-go/blob/master/zone.go#L958). is there a reason we cannot reuse that here?

yj7o5: Correct, however, there is a slight difference in the response we return for this specific setting. As it returns a field called `next_scheduled_scan`: https://github.com/cloudflare/cloudflare-go/pull/2855/files#diff-ceb3700215079abfa8dfc51b35af780ed6292f4747db0d4710c6c72de18a0a66R178

Also I think the return type here should reference the `ZoneAutomaticSSLModeSetting` struct vs `ZoneSSLSetting`. Will fix that typo.

jacobbednarz: you could always put it in a pointer with `omitempty` which will conditionally set it when it is provided.

jacobbednarz: in the generated version of the libraries, these endpoints are flattened into a single struct so that would make sense to reflect here too if possible.

yj7o5: updated to use the existing Get/UpdateZoneSettings construct and added the field as part of the ZoneSetting struct with omitempty markup

Saeedazimi69 (commented):
Review

PR #2857, opened 2024-08-11
implement the Unwrap method for custom error types

favonia:
## Description

Since the introduction of error wrapping in Go 1.13, more and more Go errors implement the `Unwrap` method to work with the new error checking paradigm using `errors.{Is,As}`. This PR is to implement such an `Unwrap` method for custom error types defined in this library that wrap around a private `Error`. By implementing `Unwrap` that returns the private `Error`, these custom errors will be automatically seen by the Go standard library as wrappers of the private `Error`. Here's some sample code enabled by this PR:

```go
var cloudflareErr cloudflare.Error
if errors.As(err, &cloudflareErr) {
	fmt.Printf("HTTP status code: %d", cloudflareErr.StatusCode)
}
```

This is extremely useful when trying to distinguish "errors from the Cloudflare API itself" and "errors due to other reasons (e.g., network errors)". Also, for reasons, only `cloudflare.Error` implements the `ErrorMessageContains` method. In sum, the `Unwrap` function gives users a clean interface to retrieve the private `Error` that works very well with the current Go standard library.

Not yet. I'm worried a test involving `errors.As` would not run on platforms using older Go compilers.

## Types of changes

## Checklist:

### DIFF
--- a/.changelog/2857.txt
+++ b/.changelog/2857.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+errors: implement the Unwrap method for custom error types to access the wrapped errors via errors.Is and errors.As
+```
--- a/errors.go
+++ b/errors.go
@@ -152,6 +152,10 @@ func (e RequestError) Type() ErrorType {
 	return e.cloudflareError.Type
 }
 
+func (e RequestError) Unwrap() error {
+	return e.cloudflareError
+}
+
 func NewRequestError(e *Error) RequestError {
 	return RequestError{
 		cloudflareError: e,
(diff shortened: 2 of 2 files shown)


favonia:
I now see that the 2.x library already has all the goodies. Perhaps this PR can be seen as an attempt to make 0.x more usable before completely abandoning it.

jacobbednarz:
when it comes to the Cloudflare errors, you really don't want to be using `Unwrap` as it is here since it may contain multiple errors and you need to handle that in something like `Error()` to get the nicer formatting/handling (which could always be ported over but isn't what we're doing here). i.e. https://github.com/cloudflare/terraform-provider-cloudflare/blob/b33577b3af8207bc41b5ab4ae3ffe66f53ee8ff7/internal/sdkv2provider/resource_cloudflare_access_keys_configuration.go#L57-L64

we can add this here, however, going forward, we limiting what we do in 0.x as it detracts effort from the automatically generated versions.

PR #2931, opened 2024-08-21
add waiting_room enabled_origin_commands support

broswen:
## Description

Add support for the new `enabled_origin_commands` parameter when creating, modifying, and reading a waiting room.

Updated unit tests, also tested the branch changes using the Cloudflare API.

## Types of changes

## Checklist:

      and relies on stable APIs.
  - https://developers.cloudflare.com/api/operations/waiting-room-create-waiting-room

### DIFF
--- a/waiting_room.go
+++ b/waiting_room.go
@@ -39,7 +39,7 @@ type WaitingRoom struct {
 	CookieSuffix               string              `json:"cookie_suffix"`
 	AdditionalRoutes           []*WaitingRoomRoute `json:"additional_routes,omitempty"`
 	QueueingStatusCode         int                 `json:"queueing_status_code"`
-	EnabledOriginCommands      []string            `json:"enabled_origin_commands"`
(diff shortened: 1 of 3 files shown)


waiting_room.go
```diff
@@ -39,6 +39,7 @@ type WaitingRoom struct {
 	CookieSuffix               string              `json:"cookie_suffix"`
 	AdditionalRoutes           []*WaitingRoomRoute `json:"additional_routes,omitempty"`
 	QueueingStatusCode         int                 `json:"queueing_status_code"`
+	EnabledOriginCommands      []string            `json:"enabled_origin_commands"`
```
jacobbednarz: are you happy to always have this value sent? even when empty?

broswen: It won't hurt, but we can omit it when it's empty if that makes more sense

jacobbednarz: up to you and what the service expects. we can do either here but it will mean that if it is set, the end user would need to always fetch it first in order to keep it as part of their payload otherwise it will be removed if forgotten.

PR #3027, opened 2024-09-05
Basic support for account load balancers

tpolich:
## Description

Basic support for account level load balancers.

Duplicated the existing zone tests and altered them for account based LB which are identical except they don't have a "name".

## Types of changes

## Checklist:

- [soon] This change is using publicly documented in [cloudflare/api-schemas](https://github.com/cloudflare/api-schemas) 
      and relies on stable APIs.

### DIFF
--- a/.changelog/3027.txt
+++ b/.changelog/3027.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+load_balancing: add account load balancer support
+```
--- a/load_balancing.go
+++ b/load_balancing.go
@@ -689,10 +689,8 @@ func (api *API) CreateLoadBalancer(ctx context.Context, rc *ResourceContainer, p
 
 	var uri string
 	switch rc.Level {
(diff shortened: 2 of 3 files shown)


jacobbednarz (changes requested):
as the CRUD method signatures already include the `*ResourceContainer` we can add a condition inside of those methods for the account level load balancers instead of using seperate new methods (we already have zone + user).

tpolich:
the changelog check blows up on a go version mismatch but I've updated the functions to switch on the resource container level

jacobbednarz:
updated the usage of the `ResourceContainer` to trim down one extra switch statement. let me know when this is ready with docs and we'll get it merged in.

i'll take a look at the changelog test failure but it looks correct for the usage here.

tpolich:
oh that's a very nice improvement I didn't realize the constants actually were the text strings at the end of the day.

PR #3221, opened 2024-09-20
Adding semgrep.yml for Secrets scanning and more

hrushikeshdeshpande:
## Description

Adding Semgrep.yml file as part of App&ProdSec team's initiative to onboard Semgrep for Secrets scanning and more.

Tested on multiple of our repos.

## Types of changes

New feature (non-breaking change which adds functionality)

## Checklist:

My code follows the code style of this project.

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -1,0 +1,24 @@
+on:
+  pull_request: {}
+  workflow_dispatch: {}
+  push: 
+    branches:
+      - main
+      - master
+  schedule:
+    - cron: '0 0 * * *'
+name: Semgrep config
+jobs:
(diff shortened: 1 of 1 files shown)


jacobbednarz:
@hrushikeshdeshpande are you able to have a look and integrate this with our existing semgrep setup? i don't think it makes sense to have duplicated builds for the same tool if we can avoid it.

hrushikeshdeshpande:
The existing semgrep files we have are for rules and ignoring patterns. I'm afraid I can't integrate CI workflows to any of those (to the best of my knowledge) :/

jacobbednarz:
let's merge this for now and we can have a look at a follow up to consolidate where possible.

PR #3233, opened 2024-09-23
Add support for Speed Brain

cf-ypark:
## Description
Add support for Speed Brain feature

Unit tests have been added that align with the documented API spec.

## Types of changes

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/speed_brain.go
+++ b/speed_brain.go
@@ -1,0 +1,54 @@
+package cloudflare
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/goccy/go-json"
+)
+
+// SpeedBrainSetting defines the structure for speed brain settings.
+type SpeedBrainSetting struct {
+	ID         string     `json:"id"`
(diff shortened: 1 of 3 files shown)


speed_brain.go
jeehoon-cf, line 1: Is this OK to be placed in root, not under /internal ?

cf-ypark: is there a reason why it should fall under /internal?

jeehoon-cf: I was looking at different repo.

jeehoon-cf (approved):

cf-ypark:
not sure why `changelog-check` is failing @jacobbednarz any help?

jacobbednarz:
i don't think we need this PR as it would already be supported using https://github.com/cloudflare/cloudflare-go/blob/master/zone.go#L928-L985. terraform we _should_ just be able to add the key to the array have this Just Work™

cf-ypark:
got it, @jacobbednarz I'll close this one then. thanks!

PR #3364, opened 2024-10-09
Adds snippet support

joshuamsager:
## Description

Adds snippet support to the API

I wasn't sure how far away v2/v3 will land and my team would love to have Snippets supported in Terraform. If this is accepted, I can work on getting the Snippet rules next.

Adds:
* GetSnippet
* ListSnippets
* DeleteSnippet
* PutSnippet

I still have some further testing to do to make sure that these all behave as expected.

## Types of changes

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/.changelog/3364.txt
+++ b/.changelog/3364.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+snippets: Add support to upload snippets
+```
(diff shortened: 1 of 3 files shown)


jacobbednarz:
v2/v3 already has this support so unless you need something specifically in the v0 range, i would recommend using it instead of adding this.

we already support the newer library in the terraform provider as well.

PR #3388, opened 2024-10-14
Add WaitingRoom.CookieAttributes property

jorygeerts:
## Description

Add the [`WaitingRoom.CookieAttributes`](https://developers.cloudflare.com/api/operations/waiting-room-create-waiting-room) property.

Note that this property is mentioned in the API docs and in some guidelines (eg [Waiting Room / Additional options / Embed in an iFrame](https://developers.cloudflare.com/waiting-room/additional-options/embed-waiting-room-in-iframe/), but is not available via dash.cloudflare.com (which is mentioned in the iFrame guide).

This property is also supported in the Python and Typescript SDKs.

I've tested this locally by setting this property do non-default values in the input for the `CreateWaitingRoom()` and `UpdateWaitingRoom()` calls, and then observing that `ListWaitingRooms()` returns the correct values.

I haven't added automated tests, because there is no logic change. All a unit test would prove is that the name of the property can be copy-pasted from the Go tag into a unittest.

## Types of changes

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/waiting_room.go
+++ b/waiting_room.go
@@ -104,7 +104,7 @@ type WaitingRoomRoute struct {
 	Path string `json:"path"`
 }
 
-// WaitingRoomCookieAttributes describes a WaitingRoomCookieAttributes object
+// WaitingRoomCookieAttributes describes a WaitingRoomCookieAttributes object.
 type WaitingRoomCookieAttributes struct {
 	Samesite string `json:"samesite"`
 	Secure   string `json:"secure"`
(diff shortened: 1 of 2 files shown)


waiting_room.go
```diff
@@ -103,6 +104,12 @@ type WaitingRoomRoute struct {
 	Path string `json:"path"`
 }
 
+// WaitingRoomCookieAttributes describes a WaitingRoomCookieAttributes object
```
jacobbednarz:
```suggestion
// WaitingRoomCookieAttributes describes a WaitingRoomCookieAttributes object.
```

PR #3529, opened 2024-10-28
FLPROD-796: Extend snippet interface with DELETE and GET a single sni…

WowVeryLogin:
Adding missing method to Get and Delete a single snippet

### DIFF
--- a/snippets.go
+++ b/snippets.go
@@ -3,6 +3,7 @@ package cloudflare
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"mime/multipart"
 	"net/http"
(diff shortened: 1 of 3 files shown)


snippets.go
```diff
@@ -33,6 +38,48 @@ type SnippetRequest struct {
...
+	if err != nil {
+		return err
+	}
+
+	result := SnippetResponse{}
+	if err := json.Unmarshal(res, &result); err != nil {
+		return fmt.Errorf("%s: %w", errUnmarshalError, err)
+	}
+
+	if result.Success != true {
```
jacobbednarz:
```suggestion
	if !result.Success {
```

snippets.go
```diff
@@ -33,6 +38,48 @@ type SnippetRequest struct {
...
+		return fmt.Errorf(result.Messages[0].Message)
```
jacobbednarz:
```suggestion
		return errors.New(result.Messages[0].Message)
```

snippets.go
WowVeryLogin: Fixed

snippets.go
WowVeryLogin: Fixed

PR #3577, opened 2024-11-07
Fix disabling redact PII on logging settings

dfialho:
## Description

Trying to disable redact PII on the logging
settings would have no effect because the logging
settings object would be marshaled without the
`redact_pii` field which result in the default
value being used, which is `true` in this case.

To fix this, the `redact_pii` is now included
even set to `false` when the logging settings
object is marshaled.

Besides the automated test, the change was
manually validated by using the lib to call
the API to make the change that was causing
problems.

## Types of changes

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/teams_accounts.go
+++ b/teams_accounts.go
@@ -132,7 +132,7 @@ type TeamsAccountLoggingConfiguration struct {
 
 type TeamsLoggingSettings struct {
 	LoggingSettingsByRuleType map[TeamsRuleType]TeamsAccountLoggingConfiguration `json:"settings_by_rule_type"`
-	RedactPii                 bool                                               `json:"redact_pii,omitempty"`
+	RedactPii                 *bool                                              `json:"redact_pii,omitempty"`
 }
 
 type TeamsDeviceSettings struct {
--- a/teams_accounts_test.go
+++ b/teams_accounts_test.go
@@ -4,11 +4,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/stretchr/testify/require"
 	"net/http"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestTeamsAccount(t *testing.T) {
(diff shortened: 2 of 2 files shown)


teams_accounts_test.go
```diff
@@ -245,19 +247,23 @@ func TestTeamsAccountUpdateLoggingConfiguration(t *testing.T) {
...
+		request := readJson(t, r)
+		require.False(t, request["redact_pii"].(bool))
+
 		w.Header().Set("content-type", "application/json")
 		fmt.Fprintf(w, `{
 			"success": true,
 			"errors": [],
 			"messages": [],
-			"result": {"settings_by_rule_type":{"dns":{"log_all":false,"log_blocks":true}, "http":{"log_all":true,"log_blocks":false}, "l4": {"log_all": false, "log_blocks": true}},"redact_pii":true}
+			"request": {"settings_by_rule_type":{"dns":{"log_all":false,"log_blocks":true}, "http":{"log_all":true,"log_blocks":false}, "l4": {"log_all": false, "log_blocks": true}},"redact_pii":true}
```
jacobbednarz: this doesn't seem right; this should be `result` per our v4 envelope.

dfialho: This was just a copy & paste from another test. I guess this is not the only test that is currently incorrect.

dfialho: Something went really wrong with the commits here... Let me fix it.

teams_accounts.go
```diff
@@ -132,7 +132,7 @@ type TeamsAccountLoggingConfiguration struct {
 
 type TeamsLoggingSettings struct {
 	LoggingSettingsByRuleType map[TeamsRuleType]TeamsAccountLoggingConfiguration `json:"settings_by_rule_type"`
-	RedactPii                 bool                                               `json:"redact_pii,omitempty"`
+	RedactPii                 bool                                               `json:"redact_pii"`
```
jacobbednarz: https://github.com/cloudflare/cloudflare-go/blob/master/docs/conventions.md#booleans

```suggestion
	RedactPii                 *bool                                               `json:"redact_pii"`
```

PR #3594, opened 2024-11-12
Fix turnstile's UpdateTurnstileWidgetParams

punkeel:
## Description

We do not allow changing the region, so remove that.

There is a meaningful difference between setting a field to nil vs false: in one case, the previous value is kept. In the other, the false is used. With omitempty, go did the wrong thing–and prevented setting values to false (as they would NOT be sent over the wire).

🤷 

## Types of changes

I guess it's technically a breaking change to fix this. The alternative is to either always set all fields, but then `UpdateTurnstileWidgetParams{domains: ...}` would override EVERYTHING else, and that is a much worse breaking change.

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/.changelog/3594.txt
+++ b/.changelog/3594.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+turnstile: fix UpdateTurnstileWidgetParams (support setting values to `false`)
+```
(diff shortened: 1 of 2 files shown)


punkeel:
changelog failed because:
```
go: downloading golang.org/x/exp/typeparams v0.0.0-20221212164502-fae10dda9338
go: golang.org/x/tools/cmd/goimports@latest (in golang.org/x/tools@v0.27.0): go.mod:3: invalid go version '1.22.0': must match format 1.23
internal/tools/tools.go:22: running "go": exit status 1
```

don't see how that's related to my PR

PR #3596, opened 2024-11-12
FLPROD-796: Fixing wrong response type for snippet update endpoint

WowVeryLogin:
Update (PUT) endpoint returns a single object and not a list of objects. Also fixed wrong endpoint name.


.changelog/3596.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+Wrong response type for Snippet update endpoint
```
jacobbednarz:
```suggestion
snippets: fix response type for `UpdateZoneSnippet`
```

PR #3763, opened 2024-12-20
Add support for security.txt API

Cyb3r-Jak3:
## Description

Adds support for security.txt API
[API Docs](https://developers.cloudflare.com/api/resources/security_txt/)

New tests added and all tests pass

## Types of changes

## Checklist:

      and relies on stable APIs.

### DIFF
--- a/.changelog/3763.txt
+++ b/.changelog/3763.txt
@@ -1,0 +1,3 @@
+```release-note:enhancement
+security_txt: adds support for security txt apis
+```
--- a/security_txt.go
+++ b/security_txt.go
@@ -1,0 +1,96 @@
+package cloudflare
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/goccy/go-json"
+	"net/http"
+	"time"
+)
+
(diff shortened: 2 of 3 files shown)


jacobbednarz:
any reason why you can't use v3 or newer of the library? it already includes this endpoint.

Cyb3r-Jak3:
Sorry, I added this because I didn't see it available in the latest v3 tag release.

jacobbednarz:
if you can, let's use that instead so that we can move off v0 of the library where possible. also means you don't need to keep these in sync with the schemas.

Cyb3r-Jak3:
It looks like it is only available via the next branch. Is the recommendation to use that branch?

jacobbednarz:
either `@next` or `@<sha>` to avoid it moving. there will be releases coming once a few more things are ironed out.

jacobbednarz:
Closing this as the v0.x branch is no longer receiving active development. Instead, you should look to upgrade to the latest version of the library (v4 at the time of writing this).

Issue #4155, opened 2025-04-25
Rate limit errors are not returned as ErrorTypeRateLimit


dixneuf19:
### Confirm this is a Go library issue and not an underlying Cloudflare API issue

- [x] This is an issue with the Go library

### Describe the bug

Hi,

I am using https://github.com/kubernetes-sigs/external-dns to manage our Cloudflare records and we have a bug which crashes the pod when we are rate limited by the Cloudflare API. 

The external-dns code tries to handle this kind of error by checking if it is such a rate limit error and failing softly, using the `ClientRateLimited` function. 

However if rate limited, this errors won't be returned and instead a generic error with the message `exceeded available rate limit retries` will be sent.

It seems for me that the culprit is this line 

https://github.com/cloudflare/cloudflare-go/blob/57714bfbdeea095ec27f4b9bafe65bda4f178d96/cloudflare.go#L286

which returns a generic error instead of continuing to return a parsed error, with the correct type.

This generic error is treated as a fatal error by external-dns, leading to the crash.

For more details you can check the issue on external-dns repo https://github.com/kubernetes-sigs/external-dns/issues/4876#issuecomment-2830636087.

### To Reproduce

Add this test to `cloudflare_test.go`

```go
func TestClient_RateLimitError(t *testing.T) {
	setup()
	defer teardown()

	handler := func(w http.ResponseWriter, r *http.Request) {
		assert.Equal(t, http.MethodGet, r.Method, "Expected method 'GET', got %s", r.Method)
		w.Header().Set("content-type", "application/json")
		w.WriteHeader(429) // HTTP 429 Too Many Requests
		fmt.Fprint(w, `{
			"success": false,
			"errors": [{
				"code": 10000,
				"message": "Rate limit exceeded"
			}],
			"messages": [],
			"result": null
		}`)
	}

	mux.HandleFunc("/zones", handler)

	_, err := client.ListZonesContext(context.Background())

	// Check if that the error is a rate limit error
	var apiErr *Error
	assert.ErrorAs(t, err, &apiErr, "Expected error to be of type cloudflare.Error")
	assert.Equal(t, apiErr.ClientRateLimited(), true, "Expected error to be a rate limit error")
}
```

Run it, it fails but we expect that it is a rate limit error...

### Code snippets

```Go
// The code used in external-dns

zonesResponse, err := p.Client.ListZonesContext(ctx)
	if err != nil {
		var apiErr *cloudflare.Error
[...]

PR #4156, opened 2025-04-25
fix: return rate limit error with correct type

dixneuf19:

## Additional context & links

See https://github.com/cloudflare/cloudflare-go/issues/4155

### DIFF
--- a/.changelog/4156.txt
+++ b/.changelog/4156.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+api: return rate limit amd 5xx errors with correct type
+```
--- a/cloudflare.go
+++ b/cloudflare.go
@@ -262,13 +262,7 @@ func (api *API) makeRequestWithAuthTypeAndHeadersComplete(ctx context.Context, m
 		// retry if the server is rate limiting us or if it failed
 		// assumes server operations are rolled back on failure
 		if respErr != nil || resp.StatusCode == http.StatusTooManyRequests || resp.StatusCode >= 500 {
-			if resp != nil && resp.StatusCode == http.StatusTooManyRequests {
-				respErr = errors.New("exceeded available rate limit retries")
-			}
-
-			if respErr == nil {
-				respErr = fmt.Errorf("received %s response (HTTP %d), please try again later", strings.ToLower(http.StatusText(resp.StatusCode)), resp.StatusCode)
-			}
+			// Retry the request
 			continue
 		} else {
 			respBody, err = io.ReadAll(resp.Body)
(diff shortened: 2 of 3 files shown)


coderkakarrot:
PR LGTM. I have face the issue of getting 429 from CF. CF support team asked for the RayID but I was unable to retrieve it since method returns without adding RayID.

Starttoaster:
I found this PR from the related external-dns Issue.

It seems like the original intent in the code was to drop out of the request retry for loop when a rate limit is found. This changes that behavior to continue retrying the request. What are your thoughts on keeping the existing behavior but just sending the correct error type?

Basically, I think this PR might be changing more than one thing, when only one thing needed to change.

edit: Additionally, see the bot message about adding a changelog. In the link it tells you how to do it.

dixneuf19:
> It seems like the original intent in the code was to drop out of the request retry for loop when a rate limit is found.

Actually no, if there is a rate limit or a 5xx, it just fills `respErr` but continues the main loop (using `continue`) until we reach `api.retryPolicy.MaxRetries`. And that's the case for all calls where we have a rate limit or a 5xx.

This part alone is not an issue.

However, after the loop, the logic is wrong: if we have anything in `respErr`, then throw a generic error.

```go
// still had an error after all retries
	if respErr != nil {
		return nil, respErr
	}
```

So if the last call in the loop was a rate limit or 5xx, we send a generic error instead of a specific one. Rate limit and 5xx are even handled below this code, but are never reached.

So the idea of this PR is:

1. To remove this global catch which breaks the typed error management
2. Also remove the part where we manually fill the `respErr`, since we won't use it anyway later in the function.

Does that make sense ? Thanks for your feedback @Starttoaster 

Also, there is definitely a minor breaking change in the format of returned errors, since it will use typed Errors instead of generic ones with custom messages. I don't think that is an intended behavior and that clients should rely on generic error messages. But that is the burden of maintainers to decide if we need to at least notice of this breaking change, or find another way to preserve the existing behavior...

I will also fill out the changelog part, maybe it will help it get noticed/merged faster.

Starttoaster:
I see what you mean, yeah I think this is right, actually. 

@jacobbednarz sorry for the ping but could this get a review? I understand that this project is pretty much bot-maintained at this point. But this error handling discrepancy is currently affecting the 9 k8s clusters I run external-dns on. And while the error could be handled differently on the external-dns side, I actually think it makes a lot of sense that they're looking to get the change in this library considering the existence of a rate limiting error type that was not being used correctly.

semo6ss:
s

dixneuf19:
I opened a support case on Cloudflare to try to move this forward. Unfortunately the Community Triage Meeting falls during the night for my timezone (Paris).

dixneuf19:
Well, this PR probably won't be accepted since the `v0` is not maintained anymore 😢 https://github.com/cloudflare/cloudflare-go/issues/4155#issuecomment-2978255382

PR #4162, opened 2025-05-16
feat: add range function support for auto-pagers

jamestelfer:

## Changes being requested

Adds a "Range()" method to V4PagePaginationArrayAutoPager[T] instances that enables the use of the Go 1.23 range functions feature.

This enables a paginator to be used with more modern syntax:

```go
for record := pager.Range() {
}

// instead of

for pager.Next() {
    record := pager.Current()
}
```

This addition uses Go build tags to ensure that it can co-exist happily with the generated code, and only become available when used in a compatible codebase. Since this is generated, I looked to see if there was an easy way to submit this idea upstream to Stainless, but there wasn't anything obvious.

I don't see this as a critical addition. It just adapts the already good auto pager into the relatively new `Seq[T]` idiom. There are additional functions being added that take advantage of this interface (see `slices`) that make this additionally interesting.

It can be used as-is (although I'd appreciate suggestions on adding some testing to this if so), but you may prefer suggesting this to the Stainless crew for upstream addition.

## Additional context & links

- [Go blog on range functions](https://go.dev/blog/range-functions)

Note also that the only thing that requires Go 1.23+ is the `iter.Seq[T]` type. The `Range` method could also be defined without reference to this:

```go
func (r *V4PagePaginationArrayAutoPager[T]) Range() func(yield func(T) bool) {
  // body elided
}
```

No version constraint is required here, but its utility in earlier versions of Go is questionable.

### DIFF
--- a/packages/pagination/range.go
+++ b/packages/pagination/range.go
@@ -1,0 +1,18 @@
+//go:build go1.23
+
+package pagination
+
+import "iter"
+
+// Range return an iterator suitable for use with the Go 1.23+ range-functions
+// feature. Iterating over an auto-pager then enables the simplified
+// `for n := pager.Range() {}` syntax.
+func (r *V4PagePaginationArrayAutoPager[T]) Range() iter.Seq[T] {
+	return func(yield func(T) bool) {
+		for r.Next() {
+			if !yield(r.Current()) {
+				return
+			}
+		}
+	}
+}


jacobbednarz:
thanks for the suggestion. this isn't something we'll take right now (as we already have `AutoPaging()` and need to be < 1.23 compatible) but we'll keep note of it for future updates.

jamestelfer:
No problems @jacobbednarz, thanks for the response!

Note that the change as presented is fully compatible with Go versions < 1.23. It will not be compiled for earlier versions. As shown in the PR description too, the return type can be defined without the `iter.Seq[T]` reference, and it would compile without issue on earlier Go versions (it just wouldn't be as useful).

I do understand that it still might not be right to include (among other things, it creates an API that would need to be supported) but I wanted to make sure that the compatibility was well understood.

Thanks again!

PR #4176, opened 2025-07-31
Fix DLP codegen

mgirouard:

## Changes being requested

We codegen'd an extra `body` param on two DLP methods. It appears that the second one was meant to replace the first.

## Additional context & links

n/a


mgirouard:
This PR seems to have picked up extra changes. Fixing.

mgirouard:
Declining this for now.

PR #4208, opened 2025-11-19
bug(ListItem): ListItemDeleteParamsItem struct missing ID field

### LINKED ISSUE #4207: ListItemDeleteParamsItem missing ID field
### Confirm this is a Go library issue and not an underlying Cloudflare API issue

- [x] This is an issue with the Go library

### Describe the bug

I suspect the "ID" field is missing from `ListItemDeleteParamsItem`.

Background: Cloudflare has a feature where it stores lists of IPs, hostnames, or ASNs to be used in formulas.  The web UI for this is in account -\> Manage Account -\> Settings -\> Lists  

I'm trying to write Go lang code that deletes a single item from one of these lists. The list item has an ID stored in the variable `listID`.

Here's my attempt at doing this:

```
	item, err = client.Rules.Lists.Items.Delete(
		context.TODO(),
		listID,
		rules.ListItemDeleteParams{
			AccountID: cloudflare.F(accountID),
			Items: cloudflare.F(
				[]rules.ListItemDeleteParamsItem{{ID: cloudflare.F(itemID)}},
			),
			Items: cloudflare.F(
				[]rules.ListItemDeleteParamsItem{
					{ID: cloudflare.F(itemID)},
				},
			),
		},
	)
```

Sadly this gets the following error:

```
filename.go:222:39: unknown field ID in struct literal of type rules.ListItemDeleteParamsItem
```

I suspect that this may be a bug in the Cloudflare SDK because the field `Items` is of type `param.Field[[]ListItemDeleteParamsItem]` which does not include a field "ID". Nor does the struct `ListItemDeleteParamsItem` include a field ID.

If I modify the SDK to add such a field, I'm able to compile my code and individual items.

```
type ListItemDeleteParamsItem struct {
	ID param.Field[string] `json:"id,required"`
}
```

Is the `ID` field intentionally left out of `ListItemDeleteParamsItem`? If not, would you accept a PR to add it?

### To Reproduce

See above

### Code snippets

```Go
See above.
```

### OS

macOS

### Go version

go1.25.4

### Library version

github.com/cloudflare/cloudflare-go/v6 v6.2.0

tlimoncelli:
- [X] I understand that this repository is auto-generated and my pull request may not be merged

## Changes being requested

Add missing `ID` field to `struct ListItemDeleteParamsItem`

## Additional context & links

Fixes https://github.com/cloudflare/cloudflare-go/issues/4207

I have a hunch that this code is autogenerated and will be overwritten. However I don't have access to the upstream data that is used to autogenerate the code, which would be required to make this change the right way.  (If I'm wrong about that, please point me in the right direction!)


tlimoncelli:
Fixed in 6.4.0

PR #4245, opened 2026-03-21
fix(secrets_store): StoreNewParams.MarshalJSON sends object not array

### LINKED ISSUE #4244: SDK Bug: StoreNewParams.MarshalJSON sends array instead of object for Secrets Store API
## Bug Description

The `StoreNewParams.MarshalJSON()` method in `secrets_store/store.go` incorrectly serializes the `Body` field as a JSON array instead of a JSON object.

## Expected Behavior

When creating a Secrets Store, the API expects a JSON object like:
```json
{"account_id": "...", "name": "my-store"}
```

## Actual Behavior

The SDK sends a JSON array:
```json
[{"account_id": "...", "name": "my-store"}]
```

This causes the API to return `400 Bad Request` with error `invalid_json_body`.

## Root Cause

In `secrets_store/store.go`, the `StoreNewParams` struct has:
```go
type StoreNewParams struct {
    AccountID param.Field[string]  `path:"account_id,required"`
    Body      []StoreNewParamsBody `json:"body,required"`
}
```

The `StoreNewParamsBody` struct has:
```go
type StoreNewParamsBody struct {
    Name param.Field[string] `json:"name,required"`
}
```

And `StoreNewParams.MarshalJSON()` calls:
```go
func (r StoreNewParams) MarshalJSON() (data []byte, err error) {
    return apijson.MarshalRoot(r.Body)  // This marshals the slice, not the object
}
```

The `apijson.MarshalRoot()` function marshals `r.Body` directly, which is a `[]StoreNewParamsBody`, resulting in a JSON array.

## API Endpoint

`POST /accounts/{account_id}/secrets_store/stores`

## Request/Response Example

**Request:**
```http
POST /accounts/xxx/secrets_store/stores
{"body":[{"name":"test-store"}]}
```

**Response:**
```json
{
  "result": null,
  "success": false,
  "errors": [{"code": 1001, "message": "invalid_json_body"}]
}
```

## Suggested Fix

The `StoreNewParams.MarshalJSON()` should marshal a single object, not the array. One approach:
```go
func (r StoreNewParams) MarshalJSON() (data []byte, err error) {
    if len(r.Body) > 0 {
        return apijson.MarshalRoot(r.Body[0])  // Marshal first element as object
    }
    return apijson.MarshalRoot(StoreNewParamsBody{})
}
```

## Impact

This breaks the Terraform provider's ability to create Secrets Store resources.

## Affected Version

`github.com/cloudflare/cloudflare-go/v6 v6.8.1-0.20260305173533-4278d1a35939`

## References

- Terraform Provider PR: https://github.com/cloudflare/terraform-provider-cloudflare/pull/6940

trafgals:
## Summary

Fixes `StoreNewParams.MarshalJSON()` which incorrectly serialized the `Body` field as a JSON array instead of a JSON object for the Secrets Store API.

## Problem

The API expects a JSON object like `{"name": "my-store"}` but the SDK was sending `[{"name": "my-store"}]` which caused `400 Bad Request` with error `invalid_json_body`.

## Solution

The fix marshals the first element of the Body slice instead of the entire slice, and handles the empty slice case gracefully.

## Changes

- **secrets_store/store.go**: Modified `StoreNewParams.MarshalJSON()` to marshal `r.Body[0]` instead of `r.Body`
- **secrets_store/store_new_params_test.go**: Added unit tests to verify the fix

## Tests

Added comprehensive unit tests that verify:
1. Single body element marshals as an object, not an array
2. Empty body marshals without error
3. First element is used when multiple elements are provided

Fixes #4244

### DIFF
--- a/secrets_store/store.go
+++ b/secrets_store/store.go
@@ -215,7 +215,10 @@ type StoreNewParams struct {
 }
 
 func (r StoreNewParams) MarshalJSON() (data []byte, err error) {
-	return apijson.MarshalRoot(r.Body)
+	if len(r.Body) > 0 {
+		return apijson.MarshalRoot(r.Body[0])
+	}
+	return apijson.MarshalRoot(StoreNewParamsBody{})
 }
 
 type StoreNewParamsBody struct {
--- a/secrets_store/store_new_params_test.go
+++ b/secrets_store/store_new_params_test.go
@@ -1,0 +1,133 @@
+// File generated from our OpenAPI spec by Stainless. See CONTRIBUTING.md for details.
+
+package secrets_store
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/cloudflare/cloudflare-go/v6/internal/param"
+)
+
+func TestStoreNewParamsMarshalJSON(t *testing.T) {
+	tests := []struct {
+		name           string
+		params         StoreNewParams
+		expectedHasKey bool
+	}{
+		{
+			name: "single body element marshals as object not array",
+			params: StoreNewParams{
+				AccountID: param.Field[string]{Present: true, Value: "account-123"},
+				Body: []StoreNewParamsBody{{
+					Name: param.Field[string]{Present: true, Value: "test-store"},
+				}},
+			},
+			expectedHasKey: true,
+		},
+		{
+			name: "empty body marshals empty object",
+			params: StoreNewParams{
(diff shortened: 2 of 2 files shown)


vaishakdinesh:
We'd like this to be fixed through codegen. Will get the issue on the service teams radar. Thank you!

